PR #137, opened 2015-01-08
add optional progressbar to double deletions

aebrahim:
uses the IProgress package to create progressbars for both
 command line and notebook interfaces.

notebook
cmd


hredestig:
this is getting stale, any objections to closing it (can always be re-opened later)?

cdiener:
Nope, I think in those days [tqdm](https://github.com/noamraph/tqdm) is the preferred option anyways.

PR #140, opened 2015-01-13
metabolite notes field should not always serialize as a string

jlerman44:
This edit allows metabolite notes fields to not always serialize as a plain string to keep consistent with how it can be used for reactions.

What if the user puts a dict in for notes? This is useful for some applications.


jlerman44:
There is also a "parse_formula function" for the metabolites. The only thing is, when reading a json file, the method just reads component formulas as a string and does not try to parse it?

aebrahim:
Thanks for pointing this out Josh. Hopefully e1cf81b fixed these issues.

PR #152, opened 2015-03-04
support for sbml level 3 v1 with fbc v2

aebrahim:


zakandrewking:
I'm getting this strange error in SBML3 export: SBRG/BIGG2#82

I can share the models that are causing it.

zakandrewking:
I also have a question about this:

https://github.com/aebrahim/cobrapy/blob/242949bc313b696e1eabf1609730cf20dc7983c0/cobra/io/sbml.py#L367
```
     sbml_model = sbml_doc.createModel(cobra_model.description.split('.')[0])
```

Does SBML export use model.description, not model.id, as the ID in the SBML model? Is there a reason for that?

Thanks!!

zakandrewking:
And another note: If there are spaces in the description, then the ID/description is not saved to the SBML file. My version of libSBML: python-libsbml==5.11.4

Sorry to spam your PR. We can move this to a separate issue if it really is a bug.

zakandrewking:
Another possible bug, that occurs in 0.3.2 and this PR:

```
import cobra
m = cobra.core.Model('my_model')
m.add_metabolites(cobra.core.Metabolite('formmfr(b)_c'))
cobra.io.write_sbml_model(m, '/Users/zaking/Downloads/test.xml')
```

The SBML file has this entry (no ID):

```
      <species name="formmfr(b)_c" substanceUnits="substance" hasOnlySubstanceUnits="false" boundaryCondition="false" constant="false">
        <notes>
          <html xmlns="http://www.w3.org/1999/xhtml"/>
        </notes>
      </species>
```

I've tracked it down to `sbml_species.setId(the_id)`, which cannot accept parentheses. Is this documented somewhere? Probably confuses users. We could strip them from the IDs, or through a warning when generating SBML that these characters are a problem. `sbml_species.setId(the_id)` responds with an error code (`-4`) when this is the case.

zakandrewking:
OK, I've tracked down the issues from SBRG/BIGG2#82.

Gene IDs cannot have `-` in them, or start with a digit. I solved the first one with a `SBML_DASH` expression (like `SBML_DOT`), but I don't know how to deal with the leading digits. Does SMBL require that gene IDs match `/[a-zA-Z][a-zA-Z_]+/` like reaction and metabolite IDs?

aebrahim:
Sorry, I just got back to these. In case you haven't solved it yet, cobrapy doesn't really do all the id escaping, and it's left up to the user.

I'm not sure if cobrapy should by default be escaping id's, but a utility function to do it would probably be a useful thing to implement.

And yes, gene id's need to start with a letter just like gene/metabolite id's.

zakandrewking:
Yeah, the only thing I would request is an informative warning. The exceptions that pop up when non-standard characters are in the IDs are pretty opaque.

PR #153, opened 2015-03-06
Support for COIN-OR through coin-or/CyLP

aebrahim:


jonls:
It seems that CyLP is licensed under the [Common Public License](https://github.com/coin-or/CyLP/blob/master/LICENSE). I could be wrong but it seems to me that this license is not compatible with the GPLv3. The Free Software Foundation [lists the CPL as incompatible with the GPL](https://www.gnu.org/licenses/license-list.en.html). Have you considered that it may be a violation of the license to use CyLP?

aebrahim:
We are not including cylp's source or any binaries compiled from it. Do you
still think this is a violation? If so, how is it?

jonls:
I think it depends on whether you think that importing Python modules creates a derived work. As far as I can tell, the Free Software Foundation believes that linking to a library, even dynamically linking or linking using Python or Java `import`, etc., creates a derived work. So you have to be able to distribute the code of CyLP under the same conditions as the GPL but the CPL has additional requirements that makes this impossible.

jonls:
By the way, I think the same problem occurs already since COBRApy uses Cplex and you are obviously not able to provide the source code of Cplex. It seems that the GPL makes it possible to explicitly [make an exception](https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs) for certain libraries. [Here is an example](http://cs.brown.edu/~melsner/correlation-readme.html) of a GPL software package that uses Cplex and explicitly allows this as a special exception. It may be possible to make an exception for CyLP too but I don't know if this would be allowed by the Common Public License.

aebrahim:
I really don't see how importing a module in python creates a derived work
as it only references an API, which any other codebase implementing the
same API would work with. However IANAL, and my understanding may
admittedly be a little bit weak.

Do you have any good references about why this would be the case? The
closest thing I found was this argument
<http://www.gnu.org/licenses/gpl-faq.html#InterpreterIncompat> but it is
strictly in reference to the JNI, which I think is pretty different from
"import" in python.

jonls:
I see your point about the API and I agree that someone could, in theory, reimplement the API with a compatible license. I don't think this is sufficient to argue that COBRApy is completely independent from the library. As far as I understand, one of the notable disagreements about the interpretation of the GPL was [CLISP using the GPL licensed readline library](https://en.wikipedia.org/wiki/GNU_Readline#Choice_of_the_GPL_as_GNU_Readline.27s_license). CLISP ended up changing to the GPL even though a BSD-licensed library had been created that reimplemented the readline API.

I think the GPL itself is quite clear. It requires you to provide the source code of the "Corresponding Source" under the terms of the GPL, and in the definition of "Corresponding Source" it says:

> Corresponding Source includes [...] the source code for shared libraries and dynamically linked subprograms that the work is specifically designed to require, such as by intimate data communication or control flow between those subprograms and other parts of the work.

aebrahim:
The example that you give is something that is a C library, which must be
linked against in order to run (just like that JNI clause in the previous
link I sent you). I'm still not convinced this extends to "import" in
python. Given that you can even run GPL code on a closed-source interpreter
like MATLAB, I'm not sure that examples like these are directly applicable
to this situation in python.

aebrahim:
This is kind of late to be resurrecting this thread...

@jonls FYI we have relicensed cobrapy so it can be used under either LGPL or GPL. Hopefully that alleviates your concerns.

PR #165, opened 2015-06-13
Improve Python3 compatibility

afif-elghraoui:
The changes here consist of using the print function as well
as catching exceptions using "as" rather than commas.

### DIFF
--- a/cobra/external/ppmap.py
+++ b/cobra/external/ppmap.py
@@ -30,6 +30,8 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+from __future__ import print_function
+
 """
 Very basic parallel processing support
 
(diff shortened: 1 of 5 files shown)


aebrahim:
Thanks for pull request and pointing these out.

The cobra.external module (which just has ppmap) is deprecated anyways, so I've just deleted it.

Is there a reason to make the *_java solvers python 3 compatible? To my knowledge, jython (whose support is deprecated anyways) doesn't support python 3.

afif-elghraoui:
> Is there a reason to make the *_java solvers python 3 compatible? To 
> my knowledge, jython (whose support is deprecated anyways) doesn't 
> support python 3.
I guess there isn't. I didn't consider jython at all when doing this.

Thanks for cleaning up the other parts, though.

aebrahim:
No problem. If I were you (unless you actively use jython), I wouldn't mess too much with the _java stuff. Still, I very much appreciate the bugfixes. Please keep them coming!

PR #166, opened 2015-06-30
docs: Remove duplicated word

parir:


aebrahim:
Thanks for pointing this out. I fixed the issue from a separate commit (the rst file is autogenerated from an IPython notebook).

PR #167, opened 2015-07-07
Correct bug in remove_from_model in Metabolite.py

mauriceccy:
Correct a small bug in remove_from_model in core/Metabolite.py when method == "destructive'.


aebrahim:
Thanks for pointing this out.

Issue #125, opened 2014-09-19
json-schema for json export


zakandrewking:
To ensure the long-term viability of exported JSON models, it would be good to define a schema for them. [json-schema](http://json-schema.org/) lets you define these easily, and validators are available in a number of languages, including Python and JavaScript.

[Escher](https://zakandrewking.github.io/escher/) will use json-schema for defining the Escher map file format. We can also consider *.escher* and *.cobra* filename extensions to clarify which file is which.

I can help out with this.

phantomas1234:
Ok, regarding #170, why would you have an annotation attribute for metabolites and not for reactions? SBML would also be ok but much harder to implement I guess. I generated a couple of universal models from metanetx and using cobrapy's reaction.annotation was pretty useful in that respect. Is there any specific reason to not include this in the JSON spec?

aebrahim:
There isn't really a good reason not to include it in the JSON spec, other than "laziness." If it makes your life easier, we can include this in the JSON format (from the cobrapy perspective, not sure about the escher perpsective).

I brought up SBML for storage of these efforts for a kind of different reason: in my quest to make cobrapy more "maintainable", I wanted to focus on specific use cases for each of the supported I/O formats to avoid duplication of effort and features:

SBML - general purpose storage of models
JSON - working with escher
mat - working with cobra toolbox (until it supports SBML3+fbc2)
pickle - specific cobrapy-only use cases

If we go with this, it would make sense for SBML to be the places where features like this get supported. While working with SBML files and all their various flavors is often a pain (and getting support for features we need will probably take way too much time and effort), the unfortunate fact is that it is the standard for modeling, and cobra should probably "play ball" and use it.

I'd like to know your general thoughts on this, @zakandrewking and @phantomas1234 

@phantomas1234 What are your use cases for reaction.annotation? Was it something like ```reaction.annotation["metanetx_id"] = 10343```? Adding support for something like that to SBML shouldn't be too hard (and something I was planning on implementing anyways), but for arbitrary expressions you are right that SBML support will be much harder.

phantomas1234:
Laziness doesn't really count though because I served you the PR on a golden platter :smiley: 

[...]

PR #170, opened 2015-07-17
Add annotation to default reaction attributes

phantomas1234:
to prevent to_json(model) discarding reaction annotations.

Would that be possible?


aebrahim:
@zakandrewking and I were talking about finalizing a JSON spec (related to #125) so we can discuss adding this there?

Alternately, would it be preferable to encode this information in SBML? cobrapy read and write SBML level 3 + fbc2 very quickly without needing libsbml (it also can do this seamlessly to compressed files, if file size is a concern).

aebrahim:
Implemented in #171

PR #171, opened 2015-07-27
create a JSON schema

### LINKED ISSUE #125: json-schema for json export

aebrahim:
The defined format changed a few things in a compatible way by marking
some attributes as optional.

This fixes #125 and replaces #170

### DIFF
--- a/cobra/io/json.py
+++ b/cobra/io/json.py
@@ -16,16 +16,38 @@ except ImportError:
     class bool_:
         pass
 
-_DEFAULT_REACTION_ATTRIBUTES = {
-    'id', 'name', 'subsystem', 'lower_bound', 'upper_bound',
-    'objective_coefficient', 'notes', 'gene_reaction_rule', 'variable_kind'}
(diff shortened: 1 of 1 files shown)


aebrahim:
@zakandrewking can you confirm that this works with escher?

A sample file is available here: http://drosophi.la/textbook.json

aebrahim:
Also worth mentioning: We had previously discussed a "header." But I really didn't want to break compatibility when  I realized I don't have to, so I just added an optional version field to the main object (and if the version is not specified, it is assumed to be 1).

zakandrewking:
Looks good to me :neckbeard:

PR #179, opened 2015-08-16
Pass test exit status to the system

afif-elghraoui:
This is necessary to inform the operating system whether or not the
tests succeeded.

### DIFF
--- a/cobra/test/__init__.py
+++ b/cobra/test/__init__.py
@@ -63,4 +63,7 @@ suite = create_test_suite()
 
 def test_all():
     """###running unit tests on cobra py###"""
-    _unittest.TextTestRunner(verbosity=2).run(create_test_suite())
+    status = not _unittest.TextTestRunner(verbosity=2).run(
+        create_test_suite()
+    ).wasSuccessful()
+    return status


aebrahim:
Hmm. I don't think a python function should be calling sys.exit like that. Someone running this code block would not expect the python interpeter to quit on them.

```python
>>> import cobra.test
>>> cobra.test.test_all()
```

Can you modify the PR so the function will only have return the status?

Right now doing ```python setup.py test``` will correctly inform the OS if the tests succeeded (this is how travis is able to determine if the run succeeded).

afif-elghraoui:
body p { margin-bottom: 0cm; margin-top: 0pt; } 
  
  
    Ok, I completely forgot about python setup.py test. I can
    change the patch, but there's really no point to it in this case
    since someone at the interpreter can just do success =
      test_all().wasSuccessful() or something.

afif-elghraoui:
Wait, sorry. My last comment was completely wrong.

The reason why I didn't use `python setup.py test` was because I needed to be able to automate testing of the _installed_ version of the package. For that, I used `python -c "from cobra.tests import test_all; test_all()`, but that wouldn't work because no status was returned.

aebrahim:
OK. Why don't you have the test_all function return the success status, and then use

```python -c "from sys import exit; from cobra.test import test_all; exit(test_all())"```

afif-elghraoui:
That works. I've already revised the patch.

aebrahim:
Thanks.

Would you be willing rebase it into a single commit? (If you're too busy, just lmk and I'll do it).

I'm trying to keep the commit history clean.

afif-elghraoui:
Done.

PR #182, opened 2015-09-16
split write_cobra_model_to_sbml_file() into two functions

jeicher:
To fully incorporate cobrapy into JWS Online (jjj.bio.vu.nl) we need to be able to access the libsbml document object generated by cobrapy (specifically the toSBML() function). Thus, we have split the write_cobra_model_to_sbml_file() function in io/sbml.py into 2 functions, one which simply returns the libsbml document, the other which actually writes to a file.


aebrahim:
Looks good! Please rebase this into a single commit, and I will merge it.

jeicher:
Done!

PR #183, opened 2015-09-27
add kegg_ids fetching in sbml 2 legacy

jerkos:
Just add attribute (a set) to metabolite to store KEGG_IDS. 

Could be applied to reaction if we can retrieve KEGG reaction id in the sbml file (may have another implementation like inheritance). 

Work only for sbml2 legacy.


aebrahim:
If you still want to do something like this, KEGG id's etc. should go in the annotation dictionary.

PR #187, opened 2015-10-25
Fix error message for empty model

JuBra:
Saving a model without a reaction added results in a ValueError from min/max


aebrahim:
Hello, this looks good. I'm going to do the following before merging.

1) Instead of doing a try/except, check for len(cobra_model.reactions) > 0
2) Rebase as a single commit (178aed0)

To do this I will merge pull request #188. Please let me know if there are any issues with this.

JuBra:
Hello, sounds good to me.

aebrahim:
Replaced with #188

PR #189, opened 2015-10-26
Provide the possibility to pass all information in initialization

JuBra:
Hello,
I think it would be nice being able to pass the information to the reaction already during initialization as it is already the case for metabolites.

### DIFF
--- a/cobra/core/Reaction.py
+++ b/cobra/core/Reaction.py
@@ -47,26 +47,27 @@ class Reaction(Object):
 
     """
 
-    def __init__(self, name=None):
+    def __init__(self, name=None, id=None, subsystem='', lower_bound=0.,
+                 upper_bound=1000., objective_coefficient=0.):
         """An object for housing reactions and associated information
         for cobra modeling.
 
         """
-        Object.__init__(self, name)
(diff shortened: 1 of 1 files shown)


aebrahim:
Wouldn't None or 1000 evaluate to True?

I think it makes more sense for the keyword arguments to just have the
default values (i.e. lower_bound=0., upper_bound=1000. etc)

JuBra:
No, or returns the first truthy element of the expression, or in case all are falsy the last one. However giving it a second thought, there is a problem if 0 is passed as upper_bound to the function as it would be replaced.. I changed it to your suggestion.

aebrahim:
TIL. Thanks! And thanks for the pull requests and bug fixes.

Would you mind doing the rebasing this into a single commit, and folding
the function definition (line 50) to 79 characters in a PEP-8 compliant way?

(Sorry I know it's a bit pedantic... if it's a pain I can just do those
things).

JuBra:
Done. I hope I did it the right way - git(hub) noob here ;)

aebrahim:
That was perfect. Thanks!

PR #193, opened 2015-11-10
Import izip from itertools for older python versions

JuBra:
izip is no longer present in python versions >3 as zip now has the same functionality and therefore leads to an ImportError for those versions.

### DIFF
--- a/cobra/solvers/gurobi_solver.py
+++ b/cobra/solvers/gurobi_solver.py
@@ -1,7 +1,11 @@
 # Interface to gurobipy
 
 from warnings import warn
-from itertools import izip
+try:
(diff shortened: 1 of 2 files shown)


cobra/solvers/gurobi_solver.py
```diff
@@ -1,7 +1,11 @@
 # Interface to gurobipy
 
 from warnings import warn
-from itertools import izip
+try:
+    # Import izip for python versions < 3.x
+    from itertools import izip as zip
+except:
```
aebrahim: should be ```except ImportError:```

aebrahim:
Thanks for fixing this

PR #194, opened 2015-11-10
Substitute .iteritems() with the iteritems function from six for pyth…

JuBra:
…on 3 compatibility


aebrahim:
Can you please modify the commit message? The subject line is too long. Sorry to be so particular, but enforcing these rules makes everything easier in the long run.

This subject line could just be ```use six.iteritems for py3 compatibility```

Otherwise, thanks, this looks great!

JuBra:
Sure, no problem.

PR #199, opened 2015-11-17
IO SBML Adds export of rxn/met note field

ChristianLieven:
Fixed: Notes field of reactions was not imported into cobrapy or cameo model as reaction.notes. It is now included as a dictionary, which makes parsing/ mapping easier.

Fixed: Notes field of reactions was not included in legacy sbml export from cobrapy or cameo model.

Fixed: When the formula of a metabolite is simply included as a string it will not be exported in the notes section correctly since the original code looked for the ID of a formula object. I changed this to check for a formula attribute regardless of its type. This should allow both options to be exported correctly.

### DIFF
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -503,14 +503,14 @@ def get_libsbml_document(cobra_model,
         #In a cobrapy model the notes section is stored as a dictionary. The following section turns the key-value-pairs
         #of the dictionary into a string and replaces recurring symbols so that the string has the required syntax for
         #an SBML doc.
-        note_str = str(the_reaction.notes.viewitems())
+        note_str = str(list(cameo_model.reactions.MNXR4410_c.notes.items()))
         note_start_tag, note_end_tag, note_delimiter = '<p>', '</p>', ':'
-        note_str = note_str.replace('dict_items([','<html xmlns="http://www.w3.org/1999/xhtml">')
         note_str = note_str.replace('(\'',note_start_tag)
         note_str = note_str.replace('\']),',note_end_tag)
         note_str = note_str.replace('\',',note_delimiter)
         note_str = note_str.replace('\']','')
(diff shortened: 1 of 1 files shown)


ChristianLieven:
Should be fixed now!

cobra/io/sbml.py
```diff
@@ -291,6 +291,7 @@ def create_cobra_model_from_sbml_file(sbml_filename, old_sbml=False, legacy_meta
         if 'SUBSYSTEM' in reaction_note_dict:
             reaction.subsystem = reaction_note_dict['SUBSYSTEM'][0]   
 
+        reaction.notes = reaction_note_dict
```
aebrahim: We actually don't want to change the model when writing it out.

cobra/io/sbml.py
```diff
@@ -490,13 +491,28 @@ def get_libsbml_document(cobra_model,
...
-                                                                 note_end_tag,
-                                                                 note_start_tag,
-                                                                 the_reaction.subsystem,
-                                                                 note_end_tag))
 
+        #Checks if GPR and Subsystem annotations are present in the notes section and if they are the same as those in
+        #the reaction's gene_reaction_rule/ subsystem attribute
+        #If they are not identical, they are set to be identical
+        if the_reaction.gene_reaction_rule:
+            the_reaction.notes.update({'GENE ASSOCIATION': [str(the_reaction.gene_reaction_rule)]})
```
aebrahim: Should be modifying reaction_note_dict, not the_reaction.notes

aebrahim:
Also, if you do ```from six import iteritems``` you can do ```iteritems(some_dict)``` which works correctly in both python 2 and python 3.

ChristianLieven:
Alright! I'll take care of that!

aebrahim:
Replaced by #205. Please test that and make sure it works for you.

PR #202, opened 2015-12-01
Fixes for #201

### LINKED ISSUE #201: Phenotype Phase Plane plot_matplotlib does not work in Python 3
The example from phenotype_phase_plane.ipynb will fail in Python 3 in `plot_matplotlib` with a ValueError in the `colors` array. The problem here is the initialization of the array:

```Python
colors = empty(self.growth_rates.shape, dtype="|S7")
```

This casts to a normal string in Python 2 but casts to a byte-string in Python 3 which is not supported by matplotlib in Python 3 (expects unicode). This can be fixed by deriving the correct numpy dtype at runtime using:

```Python
from numpy import ..., dtype
....
colors = empty(self.growth_rates.shape, dtype=dtype((str, 7)))
```

A probably bigger problem is the absence of tests for the entire module...

Should I create a pull request?

Cheers
Christian

cdiener:
Okay that fixes it for me. Running the tests also gave me an error in the cplex code. Fixed that as well but might be good if someone with an older cplex (<<12.6) version checks if that does not brake anything...

### DIFF
--- a/cobra/flux_analysis/phenotype_phase_plane.py
+++ b/cobra/flux_analysis/phenotype_phase_plane.py
@@ -1,5 +1,5 @@
 from numpy import linspace, zeros, array, meshgrid, abs, empty, arange, \
-    int32, unravel_index
+    int32, unravel_index, dtype
 from multiprocessing import Pool
 
 from ..solvers import solver_dict, get_solver_name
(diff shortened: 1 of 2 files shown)


aebrahim:
Thanks.

Would you mind separating the changes into 2 commits/pull requests?

cdiener:
Ok done. Cplex fix is now in #203.

Cheers
Christian

aebrahim:
Merged  2565431 instead

aebrahim:
Thanks.

PR #208, opened 2015-12-24
Added method Model.set_effective_bounds

cossio:
The new method Model.set_effective_bounds computes effective lower and upper bounds for all reactions in a model. For a example, in a model in which glucose uptake is restricted to 10, glycolysis reactions may have upper bounds of 1000, but in reality their effective upper bounds will be 10 to 20.

### DIFF
--- a/cobra/core/Model.py
+++ b/cobra/core/Model.py
@@ -365,6 +365,7 @@ class Model(Object):
             self.change_objective(rxn)
             rxn.lower_bound = self.optimize(objective_sense='minimize', **kwargs).f
             rxn.upper_bound = self.optimize(objective_sense='maximize', **kwargs).f
+            # if the following assert fails, the original model was inconsistent. If it never fails, the model is consistent.
             assert rxn.lower_bound <= rxn.upper_bound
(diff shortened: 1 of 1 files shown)


cossio:
The checks are failing because it cannot find Cython.... I'm not sure how these automatic checks work, but I don't think this is due to the code?

aebrahim:
Thanks for sending this. A few issues.

1) Please explain some use cases for this.
2) If it were to be implemented, this should be under cobra.flux_analysis, not cobra.core
3) This will be pretty slow... the [flux_variability_analysis](https://cobrapy.readthedocs.org/en/latest/simulating.html#running-fva) function will be much faster.

aebrahim:
Is there any update on this?

I'm closing this for now.

If this is still a feature you need, please provide the following and I'll reopen merge:
1) An explanation of the use case for this function.
2) Have the function in  cobra.flux_analysis
3) An implementation which wraps the [flux_variability_analysis](https://cobrapy.readthedocs.org/en/latest/simulating.html#running-fva) function (or does another equivalently quick optimization)

PR #214, opened 2016-02-02
Add reactions by formula to model

wbryant:
I have added a method to the cobra.core.Model class to allow the addition of models from a list of iterables, each containing a reaction ID and a reaction formula.

This has required the addition of a parseReactionFormula function, taken directly from the MATLAB COBRA toolbox, and adapted to run in Python.

I have tested this with my own reaction formulae and it appears to work - would any further testing be required?

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -19,7 +19,7 @@ _warnings.formatwarning = _warn_format
 from .version import get_version
 __version__ = get_version()
 from .core import Object, Metabolite, Gene, Reaction, Model, \
-    DictList, Species
+    DictList, Species, parseReactionFormula
 from . import io, flux_analysis, design
 
 try:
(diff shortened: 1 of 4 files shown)


wbryant:
Unit tests have been added and it has been modified to be pep8 compliant.

pstjohn:
This is definitely a useful feature for importing some non-cobra models (i.e., something like [this](http://biomet-toolbox.org/models/rxnGene/ANIGER_RXNS.txt)), but there's already a method `cobra.core.Reaction.build_reaction_from_string` that provides fairly similar functionality.

Any reason you didn't just wrap/update the existing method? Maintaining two reaction parsers would likely be tough going forward.

wbryant:
Yikes, I didn't see that!  I had searched for 'formula' and 'equation' and it didn't show up on either search!

Then I just tried to put it in a sensible place; with the other add_reaction method.

Oh well, still all good programming practice.  I'll look at the 'cobra.core.Reaction.build_reaction_from_string' and see whether it does everything I need (it probably does).  I'll modify the code to use this method.

PR #216, opened 2016-02-05
Add reactions by formula

wbryant:
I have added a method to the cobra.core.Model class to allow the addition of models from a list of iterables, each containing a reaction ID and a reaction formula, making use of the `build_reaction_from_string` method in `cobra.core.Reaction`.  It has required an update of this method to take account of compartments in the formula string.

### DIFF
--- a/cobra/core/Reaction.py
+++ b/cobra/core/Reaction.py
@@ -587,9 +587,29 @@ class Reaction(Object):
         self.lower_bound = 0
         self.upper_bound = 0
 
-    def build_reaction_from_string(self, reaction_str, verbose=True,
+    def build_reaction_from_string(self, reaction_str, verbose=False,
                                    fwd_arrow=None, rev_arrow=None,
                                    reversible_arrow=None, term_split="+"):
+        """Builds reaction from reaction equation reaction_str using parser 
+        
+        Takes a string and using the specifications supplied in the optional 
+        arguments infers a set of metabolites, metabolite compartments and 
+        stoichiometries for the reaction.  It also infers the reversibility
+        of the reaction from the reaction arrow.
+        
+        Args:
+            reaction_str: a string containing a reaction formula (equation)
+            verbose: Boolean setting verbosity of function
+                (optional, default=False)
+            fwd_arrow: re.compile for forward irreversible reaction arrows
+                (optional, default=_forward_arrow_finder)
+            reverse_arrow: re.compile for backward irreversible reaction arrows
+                (optional, default=_reverse_arrow_finder)
+            fwd_arrow: re.compile for reversible reaction arrows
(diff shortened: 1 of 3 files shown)


cobra/core/Reaction.py
```diff
@@ -587,9 +587,29 @@ def knock_out(self):
         self.lower_bound = 0
         self.upper_bound = 0
 
-    def build_reaction_from_string(self, reaction_str, verbose=True,
+    def build_reaction_from_string(self, reaction_str, verbose=False,
```
aebrahim: Default shouldn't be changed

aebrahim:
Thanks for all the hard work on this.

I really don't think the ``add_reactions_by_formula``` should go in core. (Once something goes into core, it's super difficult to remove or change its behavior because the API needs to remain stable). It's a utility function to replace the simple for loop:

```
for reaction_id, reaction_str in reaction_list:
    reaction = Reaction(reaction_id)
    model.add_reaction(reaction)
   reaction.build_reaction_from_str(reaction_str)
```

(it is assuming the defaults on build_reaction for string, and will by default ignore existing reactions, which is a little bit dangerous from a model QA standpoint).

But there's a lot of good stuff in this PR. Would you mind splitting it up into the following?

1. A PR to fix the docstring on the ``` build_reaction_from_string``` function?
2. It looks like you worked around a bug in ``` build_reaction_from_string```. Could you please explain what that bug was so I can understand your fix?

As for the ```add_reactions_by_formula``` function, I wonder if we should make a utilities module (```cobra.util```) or something like that where functions like this can go.

wbryant:
The bug in build_reaction_from_string is that the metabolites are not 
assigned to a compartment.  The 'compartment' variable is set if the 
formula string starts with '[x] ' where 'x' is the compartment ID, and 
this is added to each metabolite ID if found, but the created metabolite 
is never assigned to that compartment.

Also, this does not take into account the use of '_x' and '[x]' 
metabolite ID suffixes which are usually used to specify the compartment 
of the metabolite.

My fix in this pull request still looks for the compartment at the 
beginning of the formula string, but this is over-ridden by any 
individual metabolite suffixes of the types mentioned above.  All 
metabolite IDs are then changed to the '_x' compartment specification 
suffix and added to the model (if they do not already exist in the 
relevant compartment), with the correctly specified compartment.

This would require a slight change to the unit test for adding the 'foo' 
metabolite, and would require additional tests to ensure that 
compartments were assigned correctly.  Finally, the default compartment 
ID (if not specified anywhere in the formula string) would be 'c'.

I would be happy to put this all in a separate PR, just let me know.

By the way, thanks for taking the time to look over my proposed changes!

Cheers,

Will.

PR #221, opened 2016-02-10
Changing Metabolite formula and element methods

pstjohn:
This is a pretty basic change, really just reformulating the code for the metabolite formulas to be a bit cleaner. We really shouldn't be checking for formula errors when we put in a call to "elements", those warnings should probably get raised when the formula is set.

I try to have all my non-boundary reactions be mass-balanced, and for that reason I'll often need non-integer chemical formulas for things like biomass metabolites (I removed that warning, but its trivial to add it back in)

Its also nice to be able to set the formula using a dictionary, which is implemented in elements.setter. Specifically, you can set a combined metabolite directly from something like:
```
biomass.elements = {element : -stoich for element, stoich in reaction.check_mass_balance().iteritems()}
```

### DIFF
--- a/cobra/core/Metabolite.py
+++ b/cobra/core/Metabolite.py
@@ -43,54 +43,51 @@ class Metabolite(Species):
         self._bound = 0.
 
     @property
-    def formula(self):
-        """Describes a Chemical Formula
-        A legal formula string contains only letters and numbers.
-        """
-        try:
-            return self._formula
-        except AttributeError:
-            # Handle loading old pickled classes
-            self._formula = self.__dict__['formula']
-            return self._formula
-
-    @formula.setter
-    def formula(self, formula):
-        self._formula = str(formula) if formula is not None else ''
-
-        if "*" in self._formula:
-            warn("invalid character '*' found in formula '{}'".format(
-                self._formula))
-        if "(" in self._formula or ")" in self._formula:
-            warn("invalid formula (has parenthesis) in '{}'".format(
-                self._formula))
-        for element, stoich in element_re.findall(self._formula):
-            if element not in elements_and_molecular_weights:
-                warn('{} not a valid element'.format(element))
-            try:
-                int(stoich) if stoich else 1
-            except ValueError:
-                warn('{} not a valid element count'.format(stoich))
-
-    @property
     def elements(self):
-        """A dictionary breaking the chemical formula down by element.
-        
-        Will raise a ValueError if the formula contains non-integer
-        stochiometries, or if being set with an atom not present in the
-        elements_and_molecular_weights dict.
-        
-        """
-        return {element: int(stoich) if stoich else 1
-                for element, stoich in element_re.findall(self.formula)}
+        tmp_formula = self.formula
+        if tmp_formula is None:
+            return {}
+        # necessary for some old pickles which use the deprecated
+        # Formula class
+        tmp_formula = str(self.formula)
+        # commonly occuring characters in incorrectly constructed formulas
+        if "*" in tmp_formula:
+            warn("invalid character '*' found in formula '%s'" % self.formula)
+            tmp_formula = tmp_formula.replace("*", "")
+        if "(" in tmp_formula or ")" in tmp_formula:
+            warn("invalid formula (has parenthesis) in '%s'" % self.formula)
+            return None
+        composition = {}
+        parsed = element_re.findall(tmp_formula)
+        for (element, count) in parsed:
+            if count == '':
+                count = 1
+            else:
+                try:
+                    count = float(count)
+                    int_count = int(count)
+                    if count == int_count:
+                        count = int_count
+                    else:
+                        warn("%s is not an integer (in formula %s)" %
+                             (count, self.formula))
+                except ValueError:
+                    warn("failed to parse %s (in formula %s)" %
+                         (count, self.formula))
+                    return None
+            if element in composition:
+                composition[element] += count
+            else:
+                composition[element] = count
+        return composition
 
     @elements.setter
     def elements(self, elements_dict):
         def stringify(element, number):
-            return ''.join((element, str(number) if number != 1 else ''))
+            return element if number == 1 else element + str(number)
 
         self.formula = ''.join(stringify(e, n) for e, n in
-                                sorted(iteritems(elements_dict)))
+                               sorted(iteritems(elements_dict)))
 
     @property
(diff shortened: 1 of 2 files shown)


cobra/core/Metabolite.py
```diff
@@ -41,48 +43,57 @@ def __init__(self, id=None, formula=None,
...
+    def elements(self, elements_dict):
+
+        def formula_items():
+            for element, stoich in iteritems(elements_dict):
+                # Ignore stoichiometric entries close to machine precision.
+                if abs(stoich) < 1E-10:
+                    continue
+                yield ''.join((element, str(stoich) if stoich != 1 else ''))
+
+        self.formula = ''.join(formula_items())
```
aebrahim: Wouldn't it be inefficient to set formula if it has already been parsed into elements? I would think you would want to do self._formula = ''.join(formula_items())

aebrahim:
I that formulas with non-integer stoichiometry are better avoided.

I completely see where you are coming from with wanting to handle mass balance though. However,  I think a much better way is to simply flag a reaction as non-mass-balanced when it isn't, instead of trying to essentially "trick" it with a pseudoformula (I would argue that a biomass reaction really isn't mass balanced - it represents the dilution of biomass).

The function cobra.manipulate.validate.check_mass_balance has marked certain SBO terms as a way to annotate a reaction as non-mass-balanced. So, you could do the following:

```python
biomass_reaction.annotation["SBO"] = "SBO:0000629"
generic_nonbalanced_reaction.annotation["SBO"] = "SBO:0000631"
```

Would that work to avoid having floating point stoichiometry? The syntax could probably be made cleaner by defining constants so you don't have to remember the magic strings for the SBO terms.

aebrahim:
Also I agree with you in theory that the formula should get checked on set. We need to make sure that we handle "bad formulas" correctly though. Unfortunately, even if I don't like a formula because it doesn't follow a certain convention (like no parenthesis etc.), it would still be considered valid and needs to be stored as-is without modification. Lots of (most) models out there don't follow the convention.

The elements-based setter seems like an awesome new feature and pretty straightforward. I see no issues with that.

pstjohn:
On non-integer biomass formulas:
You'll sometimes see things like "assuming a cell composition of CH<sub>2</sub>O<sub>0.5</sub>N<sub>0.2</sub> (24.967 g/mol)". It can be nice to make sure your biomass reaction is representative of the overall cell composition, but I'm certainly happy to keep that functionality on my own fork if it would cause more problems than it solves :)

SBO annotations seem like a way around it, but honestly I never use them.

A lot of the code can get simplified if we only care about integer formulas -- I'll put in another commit with the changes.

aebrahim:
Thanks. If you're going to rewrite, then I would also suggest that the elements dict should be sorted to make the written formula correct (also deterministic).

i.e.

```python
def stringify(element, number):
    <etc>

@elements.setter
def elements(self, element_dict):
    self._formula = "".join(stringify(e, n) for e, n in sorted(elements_dict))
```

pstjohn:
This might be better, all consistency checks are done in the formula assignment.
To avoid duplicating code, the elements setter still calls the formula setter to make sure the passed dictionary is valid.

aebrahim:
Hello,

 Do you still plan on making a PR with just the elements part, or should I implement that?

It's a great idea, thanks for suggesting it

pstjohn:
Sorry about that, must have been a mistake in communication. Is there a problem with the implementation as is currently in the PR?

Formulas get stored as-is without modification, albeit that warnings are issued for non-standard formulas. This seems consistent with the current package which raises warnings about non-consistent GPRs, etc., while still allowing the model parsing to proceed.

The elements setter looks largely as you described above as well. I do invoke the .formula setter again though, in order to raise warnings if the passed atoms or stoichiometry are non-standard.

aebrahim:
Oh I see.

I'll update my comments on the diff

cobra/core/Metabolite.py
```diff
@@ -41,48 +43,58 @@ def __init__(self, id=None, formula=None,
...
+        for element, stoich in element_re.findall(self._formula):
+            if element not in elements_and_molecular_weights:
+                warn('{} not a valid element'.format(element))
+            try:
+                int(stoich) if stoich else 1
+            except ValueError:
+                warn('{} not a valid element count'.format(stoich))
+
+    @property
     def elements(self):
```
aebrahim: This old version is way less clean, which I really don't like about it. Unfortunately, we need the old complexity to handle incorrect formulas, of which there are many :(

Can we just use the old elements property code?

cobra/core/Metabolite.py
```diff
@@ -41,48 +43,58 @@ def __init__(self, id=None, formula=None,
         self._bound = 0.
 
     @property
+    def formula(self):
```
aebrahim: I considered doing this, but opted not to for performance reasons. This means that every time a metabolite is created, this check must run.

If you want the formula getter/setter let's do it in a separate PR. We'd have to do some benchmarking on it to ensure the performance isn't too bad, and probably switch to using metabolite._formula in some parts of the code (such as I/O)

cobra/core/Metabolite.py
```diff
@@ -41,48 +43,58 @@ def __init__(self, id=None, formula=None,
...
+        elements_and_molecular_weights dict.
+        
+        """
+        return {element: int(stoich) if stoich else 1
+                for element, stoich in element_re.findall(self.formula)}
+
+    @elements.setter
+    def elements(self, elements_dict):
+        def stringify(element, number):
+            return ''.join((element, str(number) if number != 1 else ''))
```
aebrahim: This is a bit hard to read.

Maybe ```return element if number == 1 else element + str(number)``` instead?

cobra/test/manipulation.py
```diff
@@ -225,6 +225,17 @@ def test_validate_formula_compartment(self):
         errors = check_metabolite_compartment_formula(model)
         self.assertEqual(len(errors), 2)
 
+    def test_validate_formula_setting(self):
```
aebrahim: Maybe ```test_formula_element_setting``` for the name here?

Thanks for writing tests too!

aebrahim:
Sorry for some of the nitpickier comments, but I've been trying to make the codebase more maintainable.

aebrahim:
Rebased into commit d8f3d50 and merged that.

Thank you for helping out with this!

PR #227, opened 2016-02-26
Updates to cobra.flux_analysis.parsimonious

pstjohn:
From my experience, there's very few times when I wouldn't prefer a parsimonious FBA solution, especially when the computational cost is just one additional LP solve. With the current implementation, however, it can be tedious to use the existing pFBA implementation.
These edits allow users to simply call:
```python
>>> model.optimize(minimize_absolute_flux=1.0)
<Solution 0.87 at 0x1190ed810>
```
to get a parsimonious FBA solution. Additionally, it borrows the concept from flux variability analysis of the fraction of optimum parameter,where passing a value of less than one will find the minimum absolute solution within a given fraction of the optimum.

Sadly If runtime imports are discouraged, this PR may also run into some troubles.

### DIFF
--- a/cobra/flux_analysis/parsimonious.py
+++ b/cobra/flux_analysis/parsimonious.py
@@ -1,8 +1,14 @@
-from ..manipulation import modify
+from six import iterkeys, iteritems
+from warnings import warn
+
+from ..manipulation.modify import convert_to_irreversible, revert_to_reversible
+from ..core import Solution
+from ..solvers import solver_dict, get_solver_name
 
 
-def optimize_minimal_flux(model, already_irreversible=False,
-                          **optimize_kwargs):
+def optimize_minimal_flux(cobra_model, already_irreversible=False,
+                          fraction_of_optimum=1.0, solver=None,
+                          desired_objective_value=None, **optimize_kwargs):
     """Perform basic pFBA (parsimonius FBA) and minimize total flux.
 
     The function attempts to act as a drop-in replacement for optimize. It
@@ -11,53 +17,71 @@ def optimize_minimal_flux(model, already_irreversible=False,
     flux. Finally, it will convert the reaction back to the irreversible
     form it was in before. See http://dx.doi.org/10.1038/msb.2010.47
 
-    model : :class:`~cobra.core.Model` object
+    cobra_model : :class:`~cobra.core.Model` object
 
     already_irreversible : bool, optional
         By default, the model is converted to an irreversible one.
         However, if the model is already irreversible, this step can be
-        skipped.
+        skipped
 
+    fraction_of_optimum : float, optional
+        Fraction of optimum which must be maintained. The original objective
+        reaction is constrained to be greater than maximal_value *
+        fraction_of_optimum. By default, this option is specified to be 1.0
+
+    desired_objective_value : float, optional
+        A desired objective value for the minimal solution that bypasses the
+        initial optimization result.
+
+    solver : string of solver name
+        If None is given, the default solver will be used.
+
+    Updates everything in-place, returns model to original state at end.
     """
-    if not already_irreversible:
-        modify.convert_to_irreversible(model)
-    model.optimize(**optimize_kwargs)
-    # if the problem is infeasible
-    if model.solution.f is None:
-        raise Exception("model could not be solved")
-    old_f = model.solution.f
-    old_objective_coefficients = {}
-    old_lower_bounds = {}
-    old_upper_bounds = {}
-    for reaction in model.reactions:
-        # if the reaction has a nonzero objective coefficient, then
-        # the same flux should be maintained through that reaction
+
+    if len(cobra_model.objective) > 1:
+        raise ValueError('optimize_minimal_flux only supports models with'
+                         ' a single objective function')
+
+    if 'objective_sense' in optimize_kwargs:
+        if optimize_kwargs['objective_sense'] == 'minimize':
+            raise ValueError(
+                'Minimization not supported in optimize_minimal_flux')
+        optimize_kwargs.pop('objective_sense', None)
+
+    # Convert to irreversible, so all reactions will have a positive flux
+    convert_to_irreversible(cobra_model)
+
+    solver = solver_dict[get_solver_name() if solver is None else solver]
+    lp = solver.create_problem(cobra_model, **optimize_kwargs)
+    if not desired_objective_value:
+        solver.solve_problem(lp, objective_sense='maximize')
+        status = solver.get_status(lp)
+        if status != "optimal":
+            raise ValueError(
+                "pFBA requires optimal solution status, not {}".format(status))
+        desired_objective_value = solver.get_objective_value(lp)
+
+    for i, reaction in enumerate(cobra_model.reactions):
+
         if reaction.objective_coefficient != 0:
-            old_objective_coefficients[reaction] = \
-                reaction.objective_coefficient
-            old_lower_bounds[reaction] = reaction.lower_bound
-            old_upper_bounds[reaction] = reaction.upper_bound
-            x = model.solution.x_dict[reaction.id]
-            reaction.lower_bound = x
-            reaction.upper_bound = x
-            reaction.objective_coefficient = 0
-        else:
-            reaction.objective_coefficient = 1
-    # set to minimize flux
-    optimize_kwargs["objective_sense"] = "minimize"
-    model.optimize(**optimize_kwargs)
-    # make the model back the way it was
-    for reaction in model.reactions:
-        if reaction in old_objective_coefficients:
-            reaction.objective_coefficient = \
-                old_objective_coefficients[reaction]
-            reaction.lower_bound = old_lower_bounds[reaction]
-            reaction.upper_bound = old_upper_bounds[reaction]
-        else:
-            reaction.objective_coefficient = 0
-    # if the minimization problem was successful
-    if model.solution.f is not None:
-        model.solution.f = old_f
-    if not already_irreversible:
-        modify.revert_to_reversible(model)
-    return model.solution
+            # Enforce a certain fraction of the original objective
+            target = (desired_objective_value * fraction_of_optimum /
+                      reaction.objective_coefficient)
+            solver.change_variable_bounds(lp, i, target, reaction.upper_bound)
+
+        # Minimize all reaction fluxes (including objective?)
+        solver.change_variable_objective(lp, i, 1)
+
+    solver.solve_problem(lp, objective_sense='minimize', **optimize_kwargs)
+    solution = solver.format_solution(lp, cobra_model)
+
+    # Return the model to its original state
+    cobra_model.solution = solution
+    revert_to_reversible(cobra_model)
+
+    if solution.status == "optimal":
+        cobra_model.solution.f = sum([coeff * reaction.x for reaction, coeff in
+                                      iteritems(cobra_model.objective)])
+
+    return solution
--- a/cobra/solvers/coin.py
+++ b/cobra/solvers/coin.py
@@ -27,7 +27,7 @@ class Coin(CyClpSimplex):
 
 
 def create_problem(cobra_model, objective_sense="maximize", **kwargs):
-    m = cobra_model.to_array_based_model()
+    m = cobra_model.to_array_based_model(deepcopy_model=True)
     lp = Coin()
     v = lp.addVariable("v", len(m.reactions))
     for i, rxn in enumerate(m.reactions):
--- a/cobra/test/flux_analysis.py
+++ b/cobra/test/flux_analysis.py
@@ -65,6 +65,48 @@ class TestCobraFluxAnalysis(TestCase):
             self.assertAlmostEqual(model.solution.f, 0.8739, places=3)
             self.assertAlmostEqual(sum(abs_x), 518.4221, places=3)
 
+            # Test desired_objective_value
+            desired_objective = 0.8
+            optimize_minimal_flux(model, solver=solver,
+                                  desired_objective_value=desired_objective)
+            abs_x = [abs(i) for i in model.solution.x]
+            self.assertEqual(model.solution.status, "optimal")
+            self.assertAlmostEqual(model.solution.f, desired_objective,
+                                   places=3)
+            self.assertAlmostEqual(sum(abs_x), 476.1594, places=3)
+
+            # Test fraction_of_optimum
+            optimize_minimal_flux(model, solver=solver,
+                                  fraction_of_optimum=0.95)
+            abs_x = [abs(i) for i in model.solution.x]
+            self.assertEqual(model.solution.status, "optimal")
+            self.assertAlmostEqual(model.solution.f, 0.95*0.8739, places=3)
+            self.assertAlmostEqual(sum(abs_x), 493.4400, places=3)
+
+            # Make sure the model works for non-unity objective values
+            model.reactions.Biomass_Ecoli_core.objective_coefficient = 2
+            optimize_minimal_flux(model, solver=solver)
+            self.assertAlmostEqual(model.solution.f, 2*0.8739, places=3)
+            model.reactions.Biomass_Ecoli_core.objective_coefficient = 1
+
+            # Test some erroneous inputs -- multiple objectives
+            model.reactions.ATPM.objective_coefficient = 1
+            with self.assertRaises(ValueError):
+                optimize_minimal_flux(model, solver=solver)
+            model.reactions.ATPM.objective_coefficient = 0
+
+            # Minimization of objective
+            with self.assertRaises(ValueError):
+                optimize_minimal_flux(model, solver=solver,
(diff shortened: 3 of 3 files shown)


aebrahim:
Thanks for the suggestion.

I would rather not have ```model.optimize()``` call optimize_minimal_flux, just for the sake of clarity. I'd prefer there only be one "correct" way to call the function.

This really doesn't seem that bad to me as far as typing goes:
```python
from cobra.flux_analysis import optimize_minimal_flux as pfba
pfba(model)
```

However, I am intrigued by the fraction_of_optimum parameter. I would accept a PR which adds a ```fraction_of_optimum``` parameter to ```optimize_minimal_flux```. Out of curiosity, could you explain a potential use case for this?

dsweaver:
Optimality has the most competitive metabolic flux distribution under 
the current set of conditions. The space of flux solutions close to 
optimality contains slightly less competitive alternate flux 
distributions. These slightly suboptimal flux distributions are 
interesting in the context of in-culture heterogeneity and selection 
over time for optimal metabolic flux states.

For examples see Ibarra 02 and Segrè 02 and related work

http://www.ncbi.nlm.nih.gov/pubmed/12432395
http://www.pnas.org/content/99/23/15112.abstract

pstjohn:
As a practical example, you'll sometimes see heavy flux through reaction
loops which generate some slight improvement in biomass yield. Allowing
some sub-optimality in objective score in exchange for a much lower
absolute sum can sometimes generate more realistic flux distributions.

On Feb 26, 2016, at 3:41 PM, dsweaver <notifications@github.com> wrote:

Optimality has the most competitive metabolic flux distribution under
the current set of conditions. The space of flux solutions close to
optimality contains slightly less competitive alternate flux
distributions. These slightly suboptimal flux distributions are
interesting in the context of in-culture heterogeneity and selection
over time for optimal metabolic flux states.

For examples see Ibarra 02 and Segrè 02 and related work

http://www.ncbi.nlm.nih.gov/pubmed/12432395
http://www.pnas.org/content/99/23/15112.abstract

On 2/26/16 2:19 PM, Ali Ebrahim wrote:

> However, I am intrigued by the fraction_of_optimum parameter. I would
> accept a PR which adds a |fraction_of_optimum| parameter to
> |optimize_minimal_flux|. Out of curiosity, could you explain a potential
> use case for this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencobra/cobrapy/pull/227#issuecomment-189506195>.
>

—
Reply to this email directly or view it on GitHub
<https://github.com/opencobra/cobrapy/pull/227#issuecomment-189513487>.

aebrahim:
Oh I meant specifically when you might want to run pFBA with fraction of optimality parameter (there are obviously many other contexts which investigate the pareto tradeoffs between growth and some other parameter). In any case, if the optimality fraction parameter is something people want in optimize_minimal_flux, I'm happy to merge it. I was asking for the use case mainly out of curiosity.

Originally one use case I thought of was using it to explore the tradeoff between growth and total flux. However, with a linear penalty, this would pretty much just scale all the fluxes by the parameter. But, on the other hand, forced fluxes (such as ATPM) would prevent this from happening.
Originally one use case I thought of was using it to explore the tradeoff between growth and total flux. However, with a linear penalty, this would pretty much just scale all the fluxes by the parameter. But, on the other hand, forced fluxes (such as ATPM) would prevent this from happening, so you might observe something interesting.

aebrahim:
Oops I didn't see your response @pstjohn  - pretty much exactly what you said

pstjohn:
Either way -- I'd be happy to take out the hooks from model.optimize and leave it as an update to minimize_absolute_flux. Its definitely an important function, I'm happy to try to clean up the implementation regardless. Although I suppose this should probably get integrated with addressing #228, which I could definitely take a look into.

Its certainly not unheard of in python package to duplicate methods in a functional and object-oriented fashion, for instance `np.max(my_array)` vs `my_array.max()`.

I suppose my main question is: if it was efficient enough, when would we *not* prefer a pFBA solution when the user calls model.optimize()? It would likely improve the solution's uniqueness across solvers and give a more representative description of in vivo fluxes.

aebrahim:
You are right that a pFBA solution is generally going to be better than an FBA solution in terms of the flux state being closer to reality.

However, it is so much less efficient that I would say it's not worth it, especially since even though it's much better, pFBA still isn't giving you a unique flux solution, just a point from a more restricted space. In many many (possibly a majority?) of use cases, calls to ```model.optimize()``` are only looking for feasibility and maximum growth rate anyways, and those shouldn't get slowed down.

Improvements to the optimize_minimal_flux function would be very appreciated though!

pstjohn:
Lets see if these tests pass. On my box at least, these are the updated timings (compared to those I list in #228)

```python
model = cobra.test.create_test_model('textbook')
%timeit optimize_minimal_flux(model)
100 loops, best of 3: 13.7 ms per loop (18.6 ms per loop previously)

%timeit optimize_minimal_flux(model)
1 loops, best of 3: 371 ms per loop (465 ms per loop previously)
```

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,90 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+        for i, reaction in enumerate(cobra_model.reactions):
+
+            if reaction.objective_coefficient != 0:
+                # Enforce a certain fraction of the original objective
+                solver.change_variable_bounds(
+                    lp, i, reaction.x * fraction_of_optimum,
+                    reaction.upper_bound)
+
+            # Minimize all reaction fluxes (including objective?)
+            solver.change_variable_objective(lp, i, -1.)
```
aebrahim: Should actually make this ```1``` and then minimize.

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,90 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
-            reaction.objective_coefficient = 0
-    # if the minimization problem was successful
-    if model.solution.f is not None:
-        model.solution.f = old_f
-    if not already_irreversible:
-        modify.revert_to_reversible(model)
-    return model.solution
+            # Use a passed objective value.
+            dummy_xdict = {
+                cobra_model.objective.keys()[0].id: desired_objective_value,
```
aebrahim: ```list(cobra_model.objective.keys())[0].id``` for python 3 compatibility

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,90 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+    if desired_objective_value:
+        assert len(original_objective) == 1
+
+    # Convert to irreversible, so all reactions will have a positive flux
+    convert_to_irreversible(cobra_model)
+
+    solver = solver_dict[get_solver_name() if solver is None else solver]
+
+    solution = None
+    try:
```
aebrahim: Why is this all in a try except block?

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,90 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+        cobra_model.solution = solution
+
+    finally:
+        # Return the model to its original state
+        revert_to_reversible(cobra_model)
+
+        try:
+            solution.f = sum([coeff * reaction.x for reaction, coeff in
+                              iteritems(cobra_model.objective)])
+        except Exception:
```
aebrahim: Check to see if ```solution.status == "optimal"``` first

cobra/solvers/cplex_solver.py
```diff
@@ -93,7 +93,7 @@ def format_solution(lp, cobra_model, **kwargs):
         x = y = x_dict = y_dict = objective_value = None
 
     return Solution(objective_value, x=x, x_dict=x_dict, status=status,
-                    y=y, y_dict=y_dict)
+                    y=y, y_dict=y_dict, solver=solver_name)
```
aebrahim: would you mind putting ```solution.solver``` in a separate PR? It seems independent, and it's much easier to analyze everything one feature at a time.

cobra/flux_analysis/parsimonious.py
pstjohn: if the optimization fails, we want to make sure the model gets converted back to a reversible model

cobra/flux_analysis/parsimonious.py
pstjohn: changed it to an iterkeys() call, but yeah

cobra/solvers/cplex_solver.py
pstjohn: Sure, I can do that. Honestly I thought I would have needed it to do this PR, but then ended up not needing it.

cobra/flux_analysis/parsimonious.py
aebrahim: It's better to handle this explicitly by using ```if solution.status == "optimal"```

Generally, only code which is expected to error should be put in the try/except block. Otherwise it obscures errors.

aebrahim:
This looks great by the way. But I'm going to nitpick a few things.

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,90 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
-        if reaction in old_objective_coefficients:
-            reaction.objective_coefficient = \
-                old_objective_coefficients[reaction]
-            reaction.lower_bound = old_lower_bounds[reaction]
-            reaction.upper_bound = old_upper_bounds[reaction]
+
+    # Store objective for later use
+    original_objective = dict(cobra_model.objective)
+    if desired_objective_value:
+        assert len(original_objective) == 1
```
aebrahim: I don't think this should be necessary

cobra/solvers/coin.py
```diff
@@ -27,7 +27,7 @@ def objectiveValue_(self):
 
 
 def create_problem(cobra_model, objective_sense="maximize", **kwargs):
-    m = cobra_model.to_array_based_model()
+    m = cobra_model.copy().to_array_based_model()
```
pstjohn: So it looks like I have to change this, otherwise calling the coin solver causes cobra_model.reaction._model to point to this new array_based_model. A `repair` call could work, but I don't really want to put in a conditional call to `model.repair()` if and only if the coin solver was used. Maybe there's a better way to handle this in `to_array_based_model`?

aebrahim: Yeah, that's a bug. It's annoying and I'm not entirely sure how to fix it.

For now this is fine.

pstjohn: sorry about that, deleted + re-pushed a commit I thought didn't require that copy, but turns out it didn't work. Using the deepcopy keyword to clean things up a bit now.

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,86 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+
+    solver = solver_dict[get_solver_name() if solver is None else solver]
+
+    solution = None
+
+    # Maximize the existing objective function.
+    if not desired_objective_value:
+        lp = solver.create_problem(cobra_model)
+        solver.solve_problem(lp, objective_sense='maximize', **optimize_kwargs)
+        solution = solver.format_solution(lp, cobra_model)
```
aebrahim: This will be an unnecessary call. We don't actually need the whole solution here, just the objective value.

You should use ```solver.get_objective_value``` and ```solver.get_status``` instead.

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,86 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+        optimize_kwargs.pop('objective_sense', None)
+
+    # Convert to irreversible, so all reactions will have a positive flux
+    convert_to_irreversible(cobra_model)
+
+    solver = solver_dict[get_solver_name() if solver is None else solver]
+
+    solution = None
+
+    # Maximize the existing objective function.
```
aebrahim: I would rewrite this block as follows
````
lp = solver.create_problem(cobra_model, **optimize_kwargs)
if not desired_objective_value:
    solver.solve_problem(lp, objective_sense='maximize')
    status = solver.get_status(lp)
    if status != "optimal":
        raise ValueError("pFBA requires optimal solution status, not " + solution.status)
    desired_objective_value = fraction_of_optimum * solver.get_objective_value(lp)
```

Then you set the objective to desired_objective_value / coefficient.

Also I think this function should raise an Exception when there is more than one objective, as discussed in this thread: https://groups.google.com/forum/#!topic/cobra-pie/AgbXPIznoIo

cobra/flux_analysis/parsimonious.py
pstjohn: Are we willing to assume there will be only one objective for the input model? I wasn't sure whether or not that was a safe assumption I could make.

cobra/flux_analysis/parsimonious.py
aebrahim: In the future (if it is needed) we can implement his solution and make a pseudo metabolite and reaction to consolidate multiple objective reactions into a single one.

cobra/flux_analysis/parsimonious.py
aebrahim: Yup! See the comment on the diff above.

Sorry I should have mentioned that earlier. It would probably have simplified your implementation quite a bit.

cobra/flux_analysis/parsimonious.py
pstjohn: Sounds great. This will now raise an exception with a multi-objective model.
I just set `desired_objective_value = solver.get_objective_value(lp)`, since I put the `fraction_of_optimum` parameter later.. If the user chooses to set both `desired_objective_value` and `fraction_of_optimum`, they should probably expect the desired value to be scaled appropriately.

pstjohn:
I think I've addressed all the critiques you mentioned, I think its a lot cleaner now -- definitely no worries in nitpicking :).

Let me know if I've missed anything

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,74 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
-        else:
-            reaction.objective_coefficient = 0
-    # if the minimization problem was successful
-    if model.solution.f is not None:
-        model.solution.f = old_f
-    if not already_irreversible:
-        modify.revert_to_reversible(model)
-    return model.solution
+            # Enforce a certain fraction of the original objective
+            solver.change_variable_bounds(
```
aebrahim: This assumes an objective_coefficient of 1.

Can you add in a test with the objective set to 2?

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,74 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+                lp, i, desired_objective_value * fraction_of_optimum,
+                reaction.upper_bound)
+
+        # Minimize all reaction fluxes (including objective?)
+        solver.change_variable_objective(lp, i, 1)
+
+    solver.solve_problem(lp, objective_sense='minimize', **optimize_kwargs)
+    solution = solver.format_solution(lp, cobra_model)
+
+    if solution.status != "optimal":
```
aebrahim: Raising an Exception is unnecessary. Just return the solution with solution.status being "failed" or whatever it is (just to be consistent with the rest of the API).

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,74 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
-    old_f = model.solution.f
-    old_objective_coefficients = {}
-    old_lower_bounds = {}
-    old_upper_bounds = {}
-    for reaction in model.reactions:
-        # if the reaction has a nonzero objective coefficient, then
-        # the same flux should be maintained through that reaction
+
+    if len(cobra_model.objective) > 1:
+        raise RuntimeWarning('optimize_minimal_flux only supports models with'
```
aebrahim: ValueError

cobra/flux_analysis/parsimonious.py
```diff
@@ -11,53 +17,74 @@ def optimize_minimal_flux(model, already_irreversible=False,
...
+                             ' a single objective function')
+
+    if 'objective_sense' in optimize_kwargs:
+        if optimize_kwargs['objective_sense'] == 'minimize':
+            raise RuntimeWarning(
```
aebrahim: ValueError

aebrahim:
This is so good! Thank you so much for your hard work on this feature.

Only a few minor things left. Also, we should test with bad inputs (using self.assertRaises) to ensure they are being raised.

You've already done a lot, and the parts remaining are minor. If you don't feel like doing them, I can add in a commit that fixes them and merges. (Although if you do decide to fix them that would be even better!!).

PR #229, opened 2016-02-28
Save reaction subsystem in sbml3

JuBra:
Hello,
as the subsystem attribute is currently not saved using sbml3 and as I was not able to find a sbml native way to store it, I would propose the usage of a cobra namespace to enable the saving of the subsystem.

### DIFF
--- a/cobra/io/sbml3.py
+++ b/cobra/io/sbml3.py
@@ -50,7 +50,8 @@ except:
 namespaces = {"fbc": "http://www.sbml.org/sbml/level3/version1/fbc/version2",
               "sbml": "http://www.sbml.org/sbml/level3/version1/core",
               "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
-              "bqbiol": "http://biomodels.net/biology-qualifiers/"}
+              "bqbiol": "http://biomodels.net/biology-qualifiers/",
+              "cobra": "https://github.com/opencobra"}
 for key in namespaces:
     register_namespace(key, namespaces[key])
 
@@ -81,6 +82,7 @@ SPECIES_XPATH = ns("sbml:listOfSpecies/sbml:species[@boundaryCondition='%s']")
 OBJECTIVES_XPATH = ns("fbc:objective[@fbc:id='%s']/"
                       "fbc:listOfFluxObjectives/"
                       "fbc:fluxObjective")
+SUBSYSTEM_ATTRIB = ns("cobra:subsystem")
 if _with_lxml:
     RDF_ANNOTATION_XPATH = ("sbml:annotation/rdf:RDF/"
                             "rdf:Description[@rdf:about=$metaid]/"
@@ -295,6 +297,7 @@ def parse_xml_into_model(xml, number=float):
             ns("sbml:listOfReactions/sbml:reaction")):
         reaction = Reaction(clip(sbml_reaction.get("id"), "R_"))
         reaction.name = sbml_reaction.get("name")
+        reaction.subsystem = sbml_reaction.get(SUBSYSTEM_ATTRIB)
         annotate_cobra_from_sbml(reaction, sbml_reaction)
         lb_id = get_attrib(sbml_reaction, "fbc:lowerFluxBound", require=True)
         ub_id = get_attrib(sbml_reaction, "fbc:upperFluxBound", require=True)
@@ -483,6 +486,8 @@ def model_to_xml(cobra_model, units=True):
             reversible=str(reaction.lower_bound < 0).lower())
         set_attrib(sbml_reaction, "name", reaction.name)
         annotate_sbml_from_cobra(sbml_reaction, reaction)
+        # add subsystem
+        set_attrib(sbml_reaction, "cobra:subsystem", reaction.subsystem)
         # add in bounds
         set_attrib(sbml_reaction, "fbc:upperFluxBound",
                    create_bound(reaction, "upper_bound"))


aebrahim:
AFAIK, this won't be standard SBML, and getting an extension into standard SBML is a very long and difficult process. This means I can't accept the pull request.

Some people (@draeger in particular) are working on how to use SBML groups to encapsulate subsystem information. When that is figured out, then I think that is the method to use here.

In the meantime, if you just want to save your models with subsystem information, I suggest using the JSON format.

JuBra:
Well, that was exactly the point for suggesting a different namespace, to keep valid sbml while saving additional information on top. However having a proper sbml extension would of course be favorable.

draeger:
Hi @JuBra, as @aebrahim pointed out, there is no need to introduce something specific to store subsystem information. 

The SBML Level 3 [Groups package](https://sourceforge.net/p/sbml/code/HEAD/tree/trunk/specifications/sbml-level-3/version-1/groups/specification/groups-2016-01-05.pdf?format=raw) is very well suited for this kind of information and already clearly defined. Subsystems can be seen as groups of reactions. The way how we now encode this information in [BiGG Models](http://bigg.ucsd.edu) is:
 * we define one group for each subsystem, e.g., "Transport, Lysosomal" with group kind "partonomy".
 * for each reaction belonging to that subsystem, we add one member element to this group with a reference to the identifier of that reaction.

It is then very straightforward to obtain all reactions belonging to one subsystem. All that needs to be done is to go through the list of groups in the model and to obtain all reaction objects (by resolving the identifier references) within a group of interest.

Similarly, you can build a hash (or dictionary) data structure pointing from each reaction identifier as key to a list or set data structure of all groups where this identifier is used. This will give you the exact information you are looking for.

This solution is simple, backed up with current SBML specifications, and does not require us to define or validate anything extra. The SBML editors are just now in the process of officially accepting the groups package, for which there is already software support, and it will soon also be included in SBML validation.

Here is an example how such a group looks like:
```xml
<groups:listOfGroups xmlns:groups="http://www.sbml.org/sbml/level3/version1/groups/version1">
  <!-- ... -->
  <groups:group groups:name="S_Citric_Acid_Cycle" groups:kind="partonomy">
      <groups:listOfMembers>
        <!-- Each member is a reaction, here referenced via its identifier -->
        <groups:member groups:idRef="R_ACONT"/>
        <groups:member groups:idRef="R_ACONTm"/>
        <groups:member groups:idRef="R_AKGDam"/>
        <groups:member groups:idRef="R_AKGDbm"/>
        <groups:member groups:idRef="R_CSm"/>
        <groups:member groups:idRef="R_CSp"/>
        <groups:member groups:idRef="R_ICDHxm"/>
        <groups:member groups:idRef="R_ICDHym"/>
        <groups:member groups:idRef="R_ICDHyp"/>
        <groups:member groups:idRef="R_ICDHyr"/>
        <groups:member groups:idRef="R_ITCOALm"/>
        <groups:member groups:idRef="R_SUCD2_u6m" />
        <groups:member groups:idRef="R_SUCOASm"/>
    </groups:listOfMembers>
  </groups:group>
  <!-- ... --> 
</groups:listOfGroups>
```

JuBra:
@draeger Thank you for the link to the package! I didn't know that there is already a sbml solution available which of course is the better solution.
@aebrahim 
What I am concerned about is that from the user perspective I think it is not obvious that depending on the format you choose to save your model in, you are loosing information. E.g. take the standard salmonella test model. Saving it to sbml3 and reading it again, you will loose subsystem and media compositions.

matthiaskoenig:
@draeger
Shouldn't be the metabolites belonging to the reactions also be part of the
group. If you want to visualize the subsystem (group) the metabolites
should be part of the group.
A subsystem corresponds in some way to a pathway and part of the pathway
should also be the metabolites. If the metabolites are not part of the
group it is unclear what to visualize if one asks only for the
visualization of the subsystem. From my understanding of a group I only
would display isolated reaction nodes.


On Mon, Feb 29, 2016 at 2:51 PM, JuBra <notifications@github.com> wrote:

> @draeger <https://github.com/draeger> Thank you for the link to the
> package! I didn't know that there is already a sbml solution available
> which of course is the better solution.
> @aebrahim <https://github.com/aebrahim>
> What I am concerned about is that from the user perspective I think it is
> not obvious that depending on the format you choose to save your model in,
> you are loosing information. E.g. take the standard salmonella test model.
> Saving it to sbml3 and reading it again, you will loose subsystem and media
> compositions.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencobra/cobrapy/pull/229#issuecomment-190219907>.
>



-- 
Matthias König
Junior Group Leader LiSym - Systems Medicine of the Liver
Humboldt-University Berlin, Institute for Theoretical Biology
  https://www.livermetabolism.com
konigmatt@googlemail.com
Tel: +49 30 20938450
Tel: +49 176 81168480

aebrahim:
By convention only reactions have been in subsystems. You can get the "metabolites" in each subystem very easily by querying all metabolites in a certain reaction group.

draeger:
In principle, you are right, @matthiaskoenig: the participants do belong to the subsystems. However, as @aebrahim pointed out, if we would also add all reaction participants as group memebers (which we could), the files would become unnecessarily large and the number of in-memory objects would dramatically increase, too. Since it is obvious that reaction participants makeup the reaction, this would also be kind of redundant information. Not including them there saves a lot of resources on all ends and the computational effort of extracting them from the reactions is very little. I hope this clarifies why we decided not to include them there.

matthiaskoenig:
So how do I now that a group is a subsystem representation? Is there an
annotation on the group?
I want to handle groups generically. The subsystem group are special groups
which do not list all members of the group, but expect tools working with
them to infer all the members of the group. This is not difficult like you
mention, but how can I distinguish this special group from a group where
someone just wants some reactions in the group (for instance group kinases,
which lists all kinases in the system). Infering here species would be
wrong.
How do I now from the sbml that I deal with such a non-standard subsystem
group?

On Tue, Mar 1, 2016 at 9:23 AM, Andreas Dräger <notifications@github.com>
wrote:

> In principle, you are right, @matthiaskoenig
> <https://github.com/matthiaskoenig>: the participants do belong to the
> subsystems. However, as @aebrahim <https://github.com/aebrahim> pointed
> out, if we would also add all reaction participants as group memebers
> (which we could), the files would become unnecessarily large and the number
> of in-memory objects would dramatically increase, too. Since it is obvious
> that reaction participants makeup the reaction, this would also be kind of
> redundant information. Not including them there saves a lot of resources on
> all ends and the computational effort of extracting them from the reactions
> is very little. I hope this clarifies why we decided not to include them
> there.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencobra/cobrapy/pull/229#issuecomment-190609535>.
>



-- 
Matthias König
Junior Group Leader LiSym - Systems Medicine of the Liver
Humboldt-University Berlin, Institute for Theoretical Biology
  https://www.livermetabolism.com
konigmatt@googlemail.com
Tel: +49 30 20938450
Tel: +49 176 81168480

draeger:
First of all, XML and hence also SBML, is hierarchically organized. If an element belongs to a group, then it belongs to that group as a whole, i.e., including all of its children. Hence, a subsystem group is not different from any other group. However, you are right that at the moment there is no specific indicator that a group of reactions is meant to represent a subsystem. A dedicated SBO term could help.

matthiaskoenig:
Yes, an SBO would be perfect. Than it is very easy to handle the group as a
subsystem group.
I mainly want to create subgraphs based on groups. In the case of the
subsystem the subgraphs should contain all the species, modifiers and gene
associations. For other groups it should only contain the members listed in
the group.

On Tue, Mar 1, 2016 at 10:19 AM, Andreas Dräger <notifications@github.com>
wrote:

> First of all, XML and hence also SBML, is hierarchically organized. If an
> element belongs to a group, then it belongs to that group as a whole, i.e.,
> including all of its children. Hence, a subsystem group is not different
> from any other group. However, you are right that at the moment there is no
> specific indicator that a group of reactions is meant to represent a
> subsystem. A dedicated SBO term could help.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencobra/cobrapy/pull/229#issuecomment-190629048>.
>



-- 
Matthias König
Junior Group Leader LiSym - Systems Medicine of the Liver
Humboldt-University Berlin, Institute for Theoretical Biology
  https://www.livermetabolism.com
konigmatt@googlemail.com
Tel: +49 30 20938450
Tel: +49 176 81168480

draeger:
Another idea came up during the SBML team meeting: instead of a new SBO term it would just now also be possible to create a new Group kind attribute value "subsystem" that would be a special case of partonomy and could only be applied to reactions. In this way, the SBO term attribute could still be used on the group for expressing some other semantic information. Please let me know what you think about that, because the groups package is nearly finalized; this would be a last minute change.

matthiaskoenig:
Hi Andreas,
I have to read the groups spec to understand what you mean with partonomy.
Is the following the latest version?
https://sourceforge.net/p/sbml/code/HEAD/tree/trunk/specifications/sbml-level-3/version-1/groups/specification/groups-2016-01-05.pdf
I will answer you in the next few hours.
M


On Wed, Mar 2, 2016 at 9:02 PM, Andreas Dräger <notifications@github.com>
wrote:

> Another idea came up during the SBML team meeting: instead of a new SBO
> term it would just now also be possible to create a new Group kind
> attribute value "subsystem" that would be a special case of partonomy and
> could only be applied to reactions. In this way, the SBO term attribute
> could still be used on the group for expressing some other semantic
> information. Please let me know what you think about that, because the
> groups package is nearly finalized; this would be a last minute change.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencobra/cobrapy/pull/229#issuecomment-191402681>.
>



-- 
Matthias König
Junior Group Leader LiSym - Systems Medicine of the Liver
Humboldt-University Berlin, Institute for Theoretical Biology
  https://www.livermetabolism.com
konigmatt@googlemail.com
Tel: +49 30 20938450
Tel: +49 176 81168480

matthiaskoenig:
Here my opinion:
The groups specification is clear about what a group is: A group is the set
of all its members defined in the listOfMembers.
With the subsystem group you want to define a group which does NOT list all
its members. For this special group additional members should be inferred
from the existing list of members (namely species and gene associations
which than together with the reactions form the subsystem/pathway).
So for someone to understand this 'special' group you have to do 2 things:

[1] Provide the rules for the inference of the additional members.
Something like: If the reaction is part of the group, all reactants,
products and modifiers of the reaction are part of the group, ...). These
have to be defined somewhere, so anybody working with such a group is able
to figure out which are the full set of members

[2] Mark the group as such a special group (either via a new SBO or via a
new groupType)
You have to state all the inference rules for the groupType in the
specification or with the SBO term.
This breaks down the very clean way of how groups are handled right now.
And it implements only one special case of an inference group (your
subsystem).
Other people probably want to do similar things like for instance I want to
create a group of rules. Here the inferred members of the group should be
all the parameters and species used in the rules.
A solution would be provide a general new kind inferenceGroup with a way to
state the inference rules. I don't like only handling the one special case
of a much broader problem.

From my perspective the cleanest way is just to list all members of the
group and use partonomy as groupKind (this will inflate all your groups,
but everybody can handle the group easily without the need for implementing
the inference rules (how to state them? where to state them?).
The main reason to only have the reactions in the group is that the COBRA
format did exactly this. The species never where annotated to a subsystem
in cobra. This should be represented 1 to 1 via groups, which is not
possible without some concept like inference rules/inference groups
For me it is unclear what the list of reactions means for visualization of
the subgraph for the group: inferring species is clear, but should I also
display the GeneAssociations for the reactions? probably yes, but I don't
have the inference rules anywhere so I cannot be sure. Everybody working
with this subsystem group will have the same problem.

Matthias

On Thu, Mar 3, 2016 at 9:55 AM, Matthias König <konigmatt@googlemail.com>
wrote:

> Hi Andreas,
> I have to read the groups spec to understand what you mean with partonomy.
> Is the following the latest version?
>
> https://sourceforge.net/p/sbml/code/HEAD/tree/trunk/specifications/sbml-level-3/version-1/groups/specification/groups-2016-01-05.pdf
> I will answer you in the next few hours.
> M
>
>
> On Wed, Mar 2, 2016 at 9:02 PM, Andreas Dräger <notifications@github.com>
> wrote:
>
>> Another idea came up during the SBML team meeting: instead of a new SBO
>> term it would just now also be possible to create a new Group kind
>> attribute value "subsystem" that would be a special case of partonomy and
>> could only be applied to reactions. In this way, the SBO term attribute
>> could still be used on the group for expressing some other semantic
>> information. Please let me know what you think about that, because the
>> groups package is nearly finalized; this would be a last minute change.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/opencobra/cobrapy/pull/229#issuecomment-191402681>.
>>
>
>
>
> --
> Matthias König
> Junior Group Leader LiSym - Systems Medicine of the Liver
> Humboldt-University Berlin, Institute for Theoretical Biology
>   https://www.livermetabolism.com
> konigmatt@googlemail.com
> Tel: +49 30 20938450
> Tel: +49 176 81168480
>



-- 
Matthias König
Junior Group Leader LiSym - Systems Medicine of the Liver
Humboldt-University Berlin, Institute for Theoretical Biology
  https://www.livermetabolism.com
konigmatt@googlemail.com
Tel: +49 30 20938450
Tel: +49 176 81168480

aebrahim:
From a COBRA point of view, only reactions have subsystems. So no matter what is in the file, only reactions will be assigned subsystems when a file is parsed by cobrapy.

I personally don't even like subsystems. I'd rather deprecate them altogether. I think annotation is a much clearer and more explicit way to go. ```groups``` is kind of a hack which I guess we can use for reactions, but I think the best thing to do is really to annotate with standardized pathways. Do you want to annotate all proteins that have Fe-S clusters, or all amino acid metabolites? Great, make MIRIAM-compliant id's for whatever you want and add it into the RDF annotation. This also forces the use of identifiers which must be linked to entities instead of unregulated strings.

draeger:
Ok, yes, @matthiaskoenig, you are right: the way how we group reactions to reflect the COBRA legacy attribute `subsystem` is not ideal because it leaves room for interpretation.

The cleanest way to encode subsystems would to use the hierarchical modeling package (comp). We would then need to decompose the models and define submodels etc. I am not in favor of extending the groups package by defining all the rules as you suggest, because then we would almost redefine the comp package.

By using groups of reactions as subsystems exactly reflects the information given in the original COBRA files. Nothing else is added, no additional assumptions are made. How this information is interpreted and how groups of reactions (or subsystems) are visually displayed is up to the user.

The advantage of using groups for this is that we do not have to change the structure of the models as we would have to do by using comp. We do not need to create yet another annotation element or additional SBML construct with all the hassle it takes from proposing until it gets accepted.

If we would list all metabolites (and maybe also all gene products) that belong to a group of reactions, we would not only significantly enlarge the files, increase the memory consumption of libraries such libSBML and JSBML, but also make up additional information that is not given by the COBRA files as @aebrahim  pointed out.

For these reasons, I propose to keep this simple solution as is. Having this said, the SBO term solution seems best for now, because it would mark the group as a subsystem group and not define what implications this has. If you insist to open this box, I would still not recommend writing all this in the groups specification. On the long term, we could think about hierarchically decomposing models, but this is not easy.

draeger:
A few more facts:

1) All we have from our source of information is the information that reactions belong to subsystems. Anything beyond that is already a question of interpretation.
2) If we would put an annotation on individual reactions, this would also not solve questions raised by @matthiaskoenig, because it would still not be entirely clear if metabolites or genes involved in this reaction also belong to that reaction's subsystem.
3) The groups package doesn't define how to visualize anything. It is a way to group model components together given some rudimentary information if it is a collection, classification, or partonomy.
4) Visual information in SBML is encoded through layout and render extensions. To make clear how a group should be displayed, it would be necessary to attach graphical objects to it.

aebrahim:
Ok. My point is if you are discussing metabolites and reactions in
subsystems it has nothing to do with cobrapy (on whose bug tracker this is
being discussed), which will ignore them.

It seems like you are discussing a separate SBML only related issue...
Perhaps the SBML mailing list is more suitable?

matthiaskoenig:
@Ali sorry for being off topic (last comment from my side)
@Andreas The solution how it is right now works for me. Your points 1-4)
summarize the main points. Unfortunately there is no visual information for
most cobra files so the layout/visualization is open to interpretation.

On Thu, Mar 3, 2016 at 8:30 PM, Ali Ebrahim <notifications@github.com>
wrote:

> Ok. My point is if you are discussing metabolites and reactions in
> subsystems it has nothing to do with cobrapy (on whose bug tracker this is
> being discussed), which will ignore them.
>
> It seems like you are discussing a separate SBML only related issue...
> Perhaps the SBML mailing list is more suitable?
> On Mar 3, 2016 11:24 AM, "Andreas Dräger" <notifications@github.com>
> wrote:
>
> > A few more facts:
> >
> > 1) All we have from our source of information is the information that
> > reactions belong to subsystems. Anything beyond that is already a
> question
> > of interpretation.
> > 2) If we would put an annotation on individual reactions, this would also
> > not solve questions raised by @matthiaskoenig
> > <https://github.com/matthiaskoenig>, because it would still not be
> > entirely clear if metabolites or genes involved in this reaction also
> > belong to that reaction's subsystem.
> > 3) The groups package doesn't define how to visualize anything. It is a
> > way to group model components together given some rudimentary information
> > if it is a collection, classification, or partonomy.
> > 4) Visual information in SBML is encoded through layout and render
> > extensions. To make clear how a group should be displayed, it would be
> > necessary to attach graphical objects to it.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/opencobra/cobrapy/pull/229#issuecomment-191924781>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencobra/cobrapy/pull/229#issuecomment-191926450>.
>



-- 
Matthias König
Junior Group Leader LiSym - Systems Medicine of the Liver
Humboldt-University Berlin, Institute for Theoretical Biology
  https://www.livermetabolism.com
konigmatt@googlemail.com
Tel: +49 30 20938450
Tel: +49 176 81168480

draeger:
Thanks for this valuable discussion. I have just requested the creation of a dedicated SBO term for subsystem groups, see https://sourceforge.net/p/sbo/term-request/113/. I think, we reached an agreement.

Issue #238, opened 2016-03-15
Some weirdness in the new docs for summaries


cdiener:
There are some weird results in the docs for the summary functions.

Looking at [section 4.3.1](http://cobrapy.readthedocs.org/en/latest/simulating.html#Running-FVA-in-summary-methods) the FVA flux for the objective value is "0.000" even though in [section 4.1.1](http://cobrapy.readthedocs.org/en/latest/simulating.html#Analyzing-FBA-solutions) it has been been established that the optimal value is "0.874". Also in "In[16]" below the pfk flux percentages are both 50% even though the reactions have different fluxes but the same stoichiometry...

aebrahim:
Pinging @pstjohn

pstjohn:
Strange, I'll take a look today.

On Mar 15, 2016, at 2:17 PM, Ali Ebrahim <notifications@github.com> wrote:

Pinging @pstjohn <https://github.com/pstjohn>

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/opencobra/cobrapy/issues/238#issuecomment-197003964>

pstjohn:
OK, so the answer to both issues is that the model hasn't been solved since the objective was changed to/from ATP maintenance. Its an easy fix, we just have to re-optimize before calling the summary method again. I'll put in a pull request.

More generally, If we wanted this to throw an error the best way might be to delete the existing solution whenever the objective value is changed. Otherwise we'll have to check whether or not the solution in model.solution indeed corresponds to the current objective value.

aebrahim:
I've created #240 to track some improvements to summary.

cdiener:
Ok thanks!

aebrahim:
Did #239 fix this?

cdiener:
Only the biomass value, not the percentages.

cdiener:
Fixed with #271.

PR #239, opened 2016-03-15
Fixing summary calls in simulating.ipynb

pstjohn:
Previously, the summary calls were made on a solution object that was
optimized for ATP maintenance, while the current objective was for
biomass optimization. Addresses #238

### DIFF
--- a/documentation_builder/simulating.ipynb
+++ b/documentation_builder/simulating.ipynb
@@ -41,7 +41,7 @@
     {
      "data": {
       "text/plain": [
-       "<Solution 0.87 at 0x7f1bb8789550>"
+       "<Solution 0.87 at 0x10ddd0080>"
       ]
      },
      "execution_count": 2,
@@ -106,7 +106,7 @@
     {
      "data": {
       "text/plain": [
-       "0.8739215069684909"
+       "0.8739215069684305"
       ]
      },
      "execution_count": 4,
@@ -172,8 +172,8 @@
       "  %      FLUX   RXN ID                        REACTION                       \n",
       " 41.6%     16     GAPD        g3p_c + nad_c + pi_c <=> 13dpg_c + h_c + nadh_c\n",
       " 24.1%    9.3      PDH     coa_c + nad_c + pyr_c --> accoa_c + co2_c + nadh_c\n",
-      " 13.1%    5.1      MDH              mal__L_c + nad_c <=> h_c + nadh_c + oaa_c\n",
       " 13.1%    5.1    AKGDH    akg_c + coa_c + nad_c --> co2_c + nadh_c + succoa_c\n",
+      " 13.1%    5.1      MDH              mal__L_c + nad_c <=> h_c + nadh_c + oaa_c\n",
       "  8.0%    3.1 Bioma...   1.496 3pg_c + 3.7478 accoa_c + 59.81 atp_c + 0.36...\n",
       "\n",
       "CONSUMING REACTIONS -- Nicotinamide adenine dinucleotide - reduced\n",
@@ -263,7 +263,7 @@
     {
      "data": {
       "text/plain": [
-       "{<Reaction Biomass_Ecoli_core at 0x7f1b92f03470>: 1.0}"
+       "{<Reaction Biomass_Ecoli_core at 0x116510828>: 1.0}"
       ]
      },
      "execution_count": 9,
@@ -292,7 +292,7 @@
     {
      "data": {
       "text/plain": [
-       "{<Reaction ATPM at 0x7f1b92f036d8>: 1}"
+       "{<Reaction ATPM at 0x1165107b8>: 1}"
       ]
      },
      "execution_count": 10,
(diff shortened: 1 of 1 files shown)


documentation_builder/simulating.ipynb
```diff
@@ -832,21 +834,21 @@
...
+    "version": 2
    },
    "file_extension": ".py",
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.4.3"
+   "pygments_lexer": "ipython2",
+   "version": "2.7.10"
```
aebrahim: The notebook has to be run in python 3. See spatialaudio/nbsphinx#27

documentation_builder/simulating.ipynb
```diff
@@ -758,13 +759,14 @@
       "PRODUCING REACTIONS -- Pyruvate\n",
       "-------------------------------\n",
       "  %             FLUX   RXN ID                        REACTION                       \n",
-      " 50.0%   6.13 ± 6.13      PYK                  adp_c + h_c + pep_c --> atp_c + pyr_c\n",
-      " 50.0%   9.76 ± 0.24   GLCpts                     glc__D_e + pep_c --> g6p_c + pyr_c\n",
+      " 85.0%   9.76 ± 0.24   GLCpts                     glc__D_e + pep_c --> g6p_c + pyr_c\n",
```
cdiener: Those percentages still seem weird. Should fluxes of ~9 and ~6 not end up in perecentages of 9/15 (60%) and 6/15 (40%)?

pstjohn: Exactly, the percentages are calculated from the FBA solution. This is probably another item to add to #240.

Its non-trivial to decide how percentages should be assigned in the FVA case. Mainly because the total flux through each node is no longer fixed, and the percentages will necessarily change. 

Finding errors in percentages would probably also be too hard, although that would be a cool feature.

Perhaps they should just not be printed to avoid confusion.

cdiener: Good point here with the median input/output problem. One could use the pFBA solution for that, however that would create another layer of complexity...

pstjohn: True, I used half-spans +/- range/2 because its a familiar format. Probably better to make summary just have "MAX" and "MIN" columns when called with FVA.

aebrahim: You could also do center of min, max

pstjohn: That's actually whats currently done:

```python
        fva_results = pd.DataFrame(
            flux_variability_analysis(met.model, met.reactions,
                                      fraction_of_optimum=fva)).T
        half_span = (fva_results.maximum - fva_results.minimum) / 2
        median = fva_results.minimum + half_span
```

Fluxes then get printed according to `median` ± `half_span`

cdiener: @pstjohn yes, completely agree. Sorry for spamming comments here. Maybe we should open a gitter to discuss stuff like this...

aebrahim: No worries about all the comments. It's all relevant to the discussion of
this bug.

cdiener: Ok :) I would vote for  the min - max format since any center value could be interpreted as the mean/median/most likely flux, all of which are false. Also it could give the impression that taking the solution where each flux takes its center value yields the reported biomass objective, which is also not true...

pstjohn:
This most recent update should hopefully address running the notebook under 3.4, (as well as fixing what seemed to be a newly depricated `round` call that raised errors under my new virtualenv.)

These changes are meant to address the main issue of #238 (model solution not being updated), which seem to be the most pressing. I'd vote to move the discussion on changes to the methods over to #240, if only because then the docs can get updated sooner.

documentation_builder/simulating.ipynb
pstjohn: So right now `summary` methods don't trigger an FBA solution, pFBA or otherwise. You can certainly get pFBA solutions in the summary methods, as long as the model is optimized prior to calling the summaries. This is partly by design, as they are supposed to be quick ways of examining a model's current state.  

This is muddled somewhat by the addition of the FVA options, which do trigger optimizations.

cobra/flux_analysis/summary.py
```diff
@@ -167,10 +164,9 @@ def model_summary(model, threshold=1E-8, fva=None, digits=2):
                               'err': half_span.loc[r.id]}
              for r in in_rxns}).T
 
-        out_fluxes.sort_values(by='x', ascending=False, inplace=True)
-        out_fluxes = out_fluxes.round(digits)
-        in_fluxes.sort_values(by='x', inplace=True)
-        in_fluxes = in_fluxes.round(digits)
+        out_fluxes = out_fluxes.sort_values(
```
aebrahim: I don't see why this is necessary.

It looks to be less efficient than the inplace call.

aebrahim:
Thanks!

I think you can remove the changes to summary and rounding - My guess is that you are trying to fix the same bug I already fixed in 9395b83. Maybe rebase off of the latest master?

pstjohn:
Oops! I missed that, thanks for being on top of the version updates. Should be fixed now.

PR #244, opened 2016-03-19
Fix for memory leak in cglpk change_coefficient

cdiener:
There was a premature return statement that did not clean up the mallocs.


aebrahim:
Nice catch. Thanks.

PR #246, opened 2016-04-05
Added handler for 'near_optimal'. Fixes #245.

### LINKED ISSUE #245: Failed test - test_single_gene_deletion with anaconda and python 3.5
After installing cobrapy using pip, at the python command prompt  I run:

```
from cobra.test import test_all
test_all()
```
I get a single test failure, 5 skipped tests and 2 expected failures!

```
======================================================================
FAIL: test_single_gene_deletion (cobra.test.flux_analysis.TestCobraFluxAnalysis)
----------------------------------------------------------------------

  File "/Users/ra381/anaconda/lib/python3.5/site-packages/cobra/test/flux_analysis.py", line 129, in test_single_gene_deletion
    self.assertEqual(statuses[gene], 'optimal')
AssertionError: 'solsta.near_optimal' != 'optimal'
- solsta.near_optimal
+ optimal
---------------------------------------------------------------------
Ran 141 tests in 13.259s
FAILED (failures=1, skipped=5, expected failures=2)
True
```

cdiener:
Mosek did not handle 'near_optimal' before. Now it is handled as optimal.

### DIFF
--- a/cobra/solvers/mosek.py
+++ b/cobra/solvers/mosek.py
@@ -20,6 +20,7 @@ _SUPPORTS_MILP = True
 status_dict = {
     mosek.solsta.dual_infeas_cer: 'infeasible',
     mosek.solsta.prim_infeas_cer: 'infeasible',
+    mosel.solsta.near_optimal: 'optimal',
     mosek.solsta.optimal: 'optimal',
     mosek.solsta.integer_optimal: 'optimal'}
 


aebrahim:
I'm not 100% convinced this is the way to go. It could obscure the fact that a solver error (only being near optimal) occurred, and the flux state being returned may not be correct. Do you think this might be an issue? I haven't looked into what "near optimal" really means.

Also, I think tuning the mosek parameters to prevent this from happening may be worthwhile.

cdiener:
Going from [here](http://docs.mosek.com/7.1/tools/The_optimizers_for_continuous_problems.html#ch-linopt-eq-term-feas-lo) near optimal seems to apply to the dual methods and means that some feasibility tolerance (primal, dual or duality gap) could not be reached, but relaxing the parameter a bit it can (scaling all three with a factor >1). It basically happens when the feasibility tolerances are yet not met but the algorithm can not improve the solution anymore. It's kind of bad practice of Mosek to roll this off to the user, since it is a problem with algorithm convergence...

Looking at the default parameters in mosek.py some tolerances seem to be very strict (1e-15, 0 etc.) which might enable that problem. Additionally a lot of the parameters are set by hand. Do you know what the logic for that was?

Either way we have to handle the "near_optimal" case somehow. For the example model the near_optimal solution is sufficiently close to the predefined one.  If we treat this as error we have to tune the parameters so the test cases are optimal. Another alternative would be to throw a warning but that would just mean letting the user deal with it again.

cdiener:
Saw that you found another fix. So we could change that pull so that solsta.near_optimal maps to "near optimal" which will trigger the usual sol != "optimal" errors and let the user decide.

aebrahim:
What do you think about having it map to ```failed```?

It means the solver failed to find a solution which meets the specified tolerance. In any case, I think this is a bug which mainly impacts conic optimization (i.e. QP's)

aebrahim:
Although I think we really don't need to do anything here. That status is an error and it should not be transparent.

cdiener:
Ok, I agree. I'll close the PR for now.

aebrahim:
OK, glad we agree. Thanks for making it anyways!

PR #247, opened 2016-04-05
Split up single_gene_deletion tests

cdiener:
Even though it was more elegant, the previous test mixed test cases for
the fba and moma single gene deletion. Those cases are now split up into
two distinct test cases to make debugging easier.


cdiener:
I think the tests should not be included in the coverage analysis. It's kind of funny that coveralls is complaining that the tests are not tested...

aebrahim:
Yeah, it's no big deal that the coverage decreased. I'll merge now.

Issue #248, opened 2016-04-05
Gurobi produces *a lot* of output in the default case


cdiener:
After glpk it seems that gurobi also has the verbosity disease now. Running a simple example

```python
from cobra.test import create_test_model

mod = create_test_model("textbook")
mod.optimize(solver="gurobi")
```

produces

```bash
Parameter OptimalityTol unchanged
   Value: 1e-06  Min: 1e-09  Max: 0.01  Default: 1e-06
Changed value of parameter IntFeasTol to 1e-09
   Prev: 1e-05  Min: 1e-09  Max: 0.1  Default: 1e-05
Changed value of parameter Method to 0
   Prev: -1  Min: -1  Max: 4  Default: -1
Changed value of parameter LogFile to 
   Prev: gurobi.log  Default: 
Parameter FeasibilityTol unchanged
   Value: 1e-06  Min: 1e-09  Max: 0.01  Default: 1e-06
Out[3]: <Solution 0.87 at 0x7fe4ecaebb38>
```

This becomes very ugly in the deletion workers or FVA....

aebrahim:
What OS and what version of gurobi? This doesn't happen on my machine (Ubuntu 14.04, python 2.7.6, gurobi version 6.0.0)

cdiener:
Ah sorry. It's Gurobi 6.5 and Python 3.4.4 on Debian 8. This is an Anaconda installation.

aebrahim:
Thanks for filing the bug. No idea why this is happening. Lots of solvers don't respect the verbosity parameter, and there are already a ton of workarounds for them. Looks like we'll need to find a way to add Gurobi to the club.

cdiener:
I agree. I liked your [context manager code](https://github.com/opencobra/cobrapy/blob/master/cobra/solvers/cglpk.pyx#L39) for cglpk (very elegant solution). Maybe we can define `quiet` in a way that it is accessible for all solvers and just use it for the solvers that won't shut up.

cdiener:
Ok, so the problem seems to be in create_problem. It seems that Gurobi 6.5 will give output by default when setting parameters. Since the parameters are set in sorted order, "verbose" is set very late which gives all this output. So we have to make sure the verbosity parameter is set first....

PR #249, opened 2016-04-05
Repair verbosity for Gurobi 6.5 (fix for #248)

cdiener:
Now sets "verbose" first. Should I also check for "output_verbosity" which is currently not set in the default parameters (and which are the only ones relevant for create_problem)?


aebrahim:
Love the branch name lol

PR #256, opened 2016-06-13
Allow add_reactions to accept a generator

pstjohn:
This commit allows Model.add_reactions to accept a generator input. Also
changes a unittest to pass a generator rather than an equivalent list.


pstjohn:
Is there anything else you'd like to see that's missing in this PR?

PR #259, opened 2016-07-17
make the build reproducible

afif-elghraoui:
This patch ensures a stable ordering for the list of dependencies generated
in setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,7 @@ extras = {
 all_extras = {'Cython>=0.21'}
 for extra in extras.values():
     all_extras.update(extra)
-extras["all"] = list(all_extras)
+extras["all"] = sorted(list(all_extras))
 
 # If using bdist_wininst, the installer will not get dependencies like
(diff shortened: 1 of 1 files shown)


afif-elghraoui:
Some background on the significance of reproducible builds: https://reproducible-builds.org/

aebrahim:
Thanks.

Should this be list(sorted( instead?

afif-elghraoui:
I don't think it matters. This particular patch was applied to the Debian package for version 0.4.1 and it's been passing the reproducibility tests as a result. Why do you suspect switching the function calls would make a difference?

aebrahim:
On Python 3 does reversed return a generator instead of a list?

afif-elghraoui:
Looks like it still returns a list in python3: https://docs.python.org/3/library/functions.html#sorted
We might not even need the type cast since `sorted`` takes any interable.

aebrahim:
Oh, good to know. I'll merge then.

PR #260, opened 2016-07-26
doc: clarify Metabolite docstring regarding compartment arg

phantomas1234:
`None or a dictionary` was probably intended be put in as the documentation for `Model.compartements`.


phantomas1234:
@aebrahim does this make sense? At some point we'll need to start to make the judgment calls but you're probably still around. I create this because of https://groups.google.com/forum/#!topic/cobra-pie/nKuVDxgH-pk

PR #263, opened 2016-07-28
Parsimonious fFBA should return consistent model

pstjohn:
This is a single-line change to optimize_minimal_flux to revert the
model to a irreverisble form before throwing a ValueError. Since these
operate in-place, the model would otherwise be changed after an erroneous
pFBA call.


hredestig:
nice catch, thanks for the pull request!

Issue #266, opened 2016-08-13
mistake in docs/comments for model.summary()


shlicshlac:
In the docs/comments for cobra.core.model.py, the argument for specifying decimal precision is listed as "round", but in cobra.flux_analysis.summary.py, method model_summary() instead takes "digits" as the parameter to do this. 

For example, using model.summary(round=4) will still give 2 digits after the decimal, whereas using model.summary(digits=4) will give 4.

Line 370 in cobra.core.model.py should be changed from `round: int` to `digits: int` so that the docs generated will be correct.

pstjohn:
Can we close this now?

Issue #240, opened 2016-03-15
improvement to summary with fva


aebrahim:
The following improvements would be useful:

 - [x] FVA only runs for reactions which might be displayed (input/output/biomass)
 - [x] solver options should be accepted (and the ability to specify a solver to override the default)
 - [ ] A new optimal flux state is computed (this one can be discussed further)

pstjohn:
* FVA only runs for reactions which might be displayed (input/output/biomass)

This one shouldn't be hard, we'll just move the `.query(lambda x: x, 'boundary')` to above the FVA call. No real point to call FVA on objectives, the range is specified by the function call.

* solver options should be accepted (and the ability to specify a solver to override the default)

Also easy, and currently only relevant to FVA solutions. I'll pass kwargs from the summary call through to the FVA call.

* A new optimal flux state is computed (this one can be discussed further)

So yeah, this one we'd need to think about. They summary methods are designed to print out nicely to a terminal window, so I'll usually call them quickly in series to follow material flow through the network after solving w/ pFBA. Triggering a new optimization at each stage would slow this down, but I agree that as shown in #238 using the existing solution can occasionally lead to unexpected behavior.

cdiener:
I think the third point could be mitigated by redesigning the output format. For instance if the summary output for the FVA case would be something like that:

```bash
PRODUCING REACTIONS -- Pyruvate
-------------------------------
  %             FLUX   RANGE   RXN ID                             REACTION
 40.0%            4        2-8        PYK                  adp_c + h_c + pep_c --> atp_c + pyr_c
 ...
```

Here the percentage and shown flux value will come from the presolved solution and FVA just adds a range. When run without FVA the RANGE column would be missing. This would solve #238 since and would be clear since the presented flux values are "real" (not the mean of the range) and the percentages can be calculated from them.

hredestig:
This all appears fixed with 20d3de8 - recalculating the optimal flux for every call to summary appears heavy and indeed mitigated by new summary output.

PR #268, opened 2016-08-16
WIP: Fix to model summary methods

### LINKED ISSUE #267: list index error in model summary
Using for example the model at https://github.com/cdiener/corda/blob/master/corda/data/cemet.xml calling

```python
model.optimize()
model.summary()
```

gives 
```bash
  File "/home/cdiener/.local/lib/python3.5/site-packages/cobra/core/Model.py", line 376, in summary
    return model_summary(self, **kwargs)
  File "/home/cdiener/.local/lib/python3.5/site-packages/cobra/flux_analysis/summary.py", line 132, in model_summary
    out_fluxes = pd.Series({r.reactants[0]: r.x for r in out_rxns})
  File "/home/cdiener/.local/lib/python3.5/site-packages/cobra/flux_analysis/summary.py", line 132, in <dictcomp>
    out_fluxes = pd.Series({r.reactants[0]: r.x for r in out_rxns})
IndexError: list index out of range
```

The error occurs since it is assumed that all reactions always have reactants. However, that does not have to be the case, particularly if not using boundary metabolites. I could set up a fix but do not understand the logic of using only the first reactant for the summary, could you give me a hint what this is supposed to do @pstjohn? 

Cheers

pstjohn:
This fix resolves #267, in which models that do not conform to the
standard boundary reaction format would cause the `model_summary` method
to fail. Now the summary method will print the ID of the first
metabolite and raise a warning if a boundary reaction exchanges more
than one species.

### DIFF
--- a/cobra/flux_analysis/summary.py
+++ b/cobra/flux_analysis/summary.py
@@ -1,5 +1,6 @@
 from six.moves import zip_longest
 from six import iterkeys, print_, text_type
+from warnings import warn
 
 import pandas as pd
 
@@ -129,8 +130,20 @@ def model_summary(model, threshold=1E-8, fva=None, digits=2, **solver_args):
             lambda rxn: rxn.x < -threshold, None
         ).query(lambda x: x, 'boundary')
 
-        out_fluxes = pd.Series({r.reactants[0]: r.x for r in out_rxns})
-        in_fluxes = pd.Series({r.reactants[0]: r.x for r in in_rxns})
+        def get_boundary_species(cobra_reaction):
+            """For a given cobra_reaction, find out which species is being
+            produced """
+            mets = list(iterkeys(cobra_reaction.metabolites))
+            if len(mets) != 1:
+                warn('The boundary reaction {} exhanges more than one '
+                     'metabolite'.format(cobra_reaction.id))
+
+            return mets[0]
+
+        out_fluxes = pd.Series({
+            get_boundary_species(r): r.x for r in out_rxns})
+        in_fluxes = pd.Series({
+            get_boundary_species(r): r.x for r in in_rxns})
 
         # sort and round
         out_fluxes.sort_values(ascending=False, inplace=True)


pstjohn:
Good catch -- I'll work on that.
You're right, splitting coupled boundary reactions shouldn't be too tough either

cdiener:
Awesome. If you need help with anything you know where to find me :)

pstjohn:
While I'm at it we might as well get to #266, #238, and #240.

How would we feel about exporting table formatting to a third party? It might make the code a bit more extendable / maintainable. [tabulate](https://pypi.python.org/pypi/tabulate) looks promising, but I'm certainly happy to look at others if you have suggestions. (Or if we want to avoid additional dependencies)

cdiener:
I agree. This kind of formatting usually has hundreds of edge cases one has to consider ("what happens if the name of the reaction is super long", etc.). I like how it's done now with converting everything to a pandas data frame and then writing some formatting for that data frame. `tabulate` looks good, we just have to check how it deals with long strings and how it plays with pandas  (for instance the reaction strings). In terms of dependencies I think everything that does not have compiled code is fine.

hredestig:
pardon the premature merge, the bug fix was already good, please open new pr for the additional issues you mentioned. Anyway, tabulate looks like good option, prettytable is another but tabulate directly takes pandas df so maybe the better option.

pstjohn:
Yikes, as @cdiener mentioned there was still an issue with assigning input and output fluxes..
I'll try to send in another PR soon that fixes the current master then.

hredestig:
ouch, yes mea culpa :( This also reminds me I think it would be good to start doing merges to devel first, and only upon release to master..

cdiener:
I think there is no problem here. The bug I reported initially was already fixed here and the rest can be in another pull req. I think bug fix pull requests are fine in master, otherwise master would just be a glorified tag for the latest release :)

pstjohn:
Sounds great. Regardless, it shouldn't be too long until I have a newer implementation ready.
Tabulate seems to be working great, and should make maintenance for these methods a lot easier going forward.

PR #270, opened 2016-08-17
Updated assess functions

oveoyas:
Made the assess functions in flux_analysis/reaction.py work with the latest cobrapy version and added solver as a keyword argument.

### DIFF
--- a/cobra/flux_analysis/reaction.py
+++ b/cobra/flux_analysis/reaction.py
@@ -4,7 +4,7 @@ from ..core.Reaction import Reaction
 from six import iteritems
 
 
-def assess(model, reaction, flux_coefficient_cutoff=0.001):
+def assess(model, reaction, flux_coefficient_cutoff=0.001, solver=None):
     """Assesses the capacity of the model to produce the precursors for the
     reaction and absorb the production of the reaction while the reaction is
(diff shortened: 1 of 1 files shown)


cdiener:
Hi, it seems like pep8 is complaining that some of the lines are now over 80 characters with the the new solver argument. I think a few line breaks for those should fix it.

Cheers

cobra/flux_analysis/reaction.py
```diff
@@ -36,7 +41,8 @@ def assess(model, reaction, flux_coefficient_cutoff=0.001):
         return results
 
 
-def assess_precursors(model, reaction, flux_coefficient_cutoff=0.001):
+def assess_precursors(model, reaction, flux_coefficient_cutoff=0.001,
```
cdiener: There is a trailing white space after the comma PEP8 is complaining about, sorry :(

oveoyas:
Thanks Christian :) Looks like there aren't any tests for these functions currently.

cdiener:
No, that is something we have to get to eventually, but it is not required here. The rest is up to @phantomas1234  or @hredestig. I don't know whether they might want the commits squashed into one but there shoudln't be a a problem. Thanks.

PR #274, opened 2016-09-02
Implement automated building of python manylinux and mac wheels

hredestig:
Fixed since first commit in devel which didn't properly fail when unit-tests failed.

Also adds dependency tabulate needed by improved model_summary.

This new feature/fix came with a couple of (minor) drawbacks
- change to codecov instead of coveralls as difficulty getting coveralls to report for installed wheel (can't use coverage run setup.py test on library package)
- (for the moment) drop support for testing matplotlib functions (as difficulty installing matplotlib in non-framework mode in docker container)
- readme files switch to native rst mode to avoid dependency on pypandoc


phantomas1234:
We need to check that the OS X wheels properly package all libs by uninstalling all third party dependencies before the wheels are tested.

hredestig:
`delocate-wheel` appears to do its job for cobrapy. Built packages include glpk library 
https://testpypi.python.org/pypi?name=cobra&version=0.4.2b2.post39&:action=files

PR #275, opened 2016-09-02
make copies of attributes in reaction

hredestig:
can be proper objects so copy by reference of e.g. dictionary `notes`
leads to unexpected behaviour as shown in #272

### DIFF
--- a/cobra/core/Model.py
+++ b/cobra/core/Model.py
@@ -128,7 +128,7 @@ class Model(Object):
             new_reaction = reaction.__class__()
             for attr, value in iteritems(reaction.__dict__):
                 if attr not in do_not_copy:
-                    new_reaction.__dict__[attr] = value
+                    new_reaction.__dict__[attr] = copy(value)
             new_reaction._model = new
             new.reactions.append(new_reaction)
             # update awareness


pstjohn:
The alternative here to to just replace `model.copy()` with `copy.deepcopy()`

We don't really have any profiling tools setup to check for changes

aebrahim:
This sounds like a pretty easy fix without replacing the entire function.
The function is useful because it's much much faster than deepcopy for
large models. (I recommend profiling with iJO or another large model by
using the timeit magic in jupyter)

aebrahim:
I just wanted to caution against premature removal of the function. You are
right - the lower level LP interface was written exactly for that reason,
and has many other benefits like basis reuse as well. However, some older
code in other repositories relied on copy, and replacing this function will
create unexpected performance regressions for those use cases.

If the current developers really want to get rid of this (I wouldn't blame
you guys, it's been a pain of a function to maintain), my suggestion would
be to mark it deprecated and warn users to use deepcopy directly instead
for at least a version.

For the record i think this PR is great (a simple fix), and I didn't mean
to hold it up (in case this discussion is).

cdiener:
Thanks for the detailed reply. Completely agree. Cheers and all the best with your new job.

PR #278, opened 2016-09-07
Additional summary test improvements

pstjohn:
This commit adds tabulate as a requirement for summary methods tests,
and re-runs the tests for each of the solvers.

waiting for solver info from @matthiaskoenig in #277


pstjohn:
Don't merge this yet.

I'm thinking that in general, the summary method tests probably shouldn't depend on the solver we use. But if we can't guarantee the solutions will be equal from solver to solver, maybe we need to pass a pre-computed solution.

cdiener:
I agree that we should not depend on a particular flux solution since there may be many flux distributions that guarantee optimality. How hard would it be to assign the fluxes by hand to the model (setting all reaction.x to 1 for instance) and calling the summary after that? Another easy option would be just to check the table headings...

pstjohn:
So I tried a couple different options. Setting them all to 1 doesn't really work, since we check the mass balance for metabolite summaries. I also want to make sure FBA gets run.

This will hopefully be the best option: I save a cglpk'd parsimonious flux balance analysis solution in the test/data directory, and then load up that as the model solution.

cdiener:
I think that's a good idea and makes it independent of future solver bugs/changes as well.

hredestig:
@pstjohn is this ready for merging?

pstjohn:
Yeah, I think its ready.

PR #281, opened 2016-09-09
Adding bounds property to reaction

pstjohn:
This PR adds a `bounds` attribute to `cobra.Reaction`, which allows
bounds to be checked and changed with a single line:

```
model.reactions.PGI.bounds = (0, 1000)
```

```
>>> model.reactions.PGI.bounds
(0, 1000)
```

It's a pretty small change, but it parallels the MATLAB toolbox's `changeRxnBounds` function, which had a `b` option to set both bounds at once.

### DIFF
--- a/cobra/test/unit_tests.py
+++ b/cobra/test/unit_tests.py
@@ -357,6 +357,9 @@ class TestReactions(CobraTestCase):
         pgi = model.reactions.get_by_id("PGI")
         pgi.reaction = "g6p_c --> f6p_c"
         self.assertEqual(pgi.lower_bound, 0)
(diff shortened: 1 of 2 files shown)


cobra/test/unit_tests.py
```diff
@@ -357,6 +357,9 @@ def test_build_from_string(self):
         pgi = model.reactions.get_by_id("PGI")
         pgi.reaction = "g6p_c --> f6p_c"
         self.assertEqual(pgi.lower_bound, 0)
+        pgi.upper_bound = 1000
```
hredestig: `pgi.bounds = (0, 1000)` to test the setter?

Issue #265, opened 2016-08-12
problem with phenotype_phase_plane and matplotlib on OSX


shlicshlac:
Using calculate_phenotype_phase_plane() from cobra.flux_analysis.phenotype_phase_plane followed by plot_matplotlib() and matplotlib.pyplot.show() on OS X raises the error: 

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/matplotlib/tight_layout.py:222 **UserWarning**: tight_layout : falling back to Agg renderer

The resulting plot appears fine, but it can't be rotated, zoomed in/out, etc. This problem didn't happen when I did the same in linux.

This is due to [an issue with matplotlib on some operating systems](https://github.com/matplotlib/matplotlib/issues/1852). One post in the github page in the above link suggests using 

`figure.set_tight_layout(True)`

instead of

`figure.tight_layout()`

I made this change to penotype_phase_plane.py (line 101) and it completely fixed the problem on OS X. 

Maybe this will be helpful for OS X users of cobrapy until the problem is fixed in matplotlib?

hredestig:
fixed #283

PR #283, opened 2016-09-19
fix broken plot on mac osx #265

hredestig:
avoid 'Fall back to Agg renderer' as suggested here
https://github.com/matplotlib/matplotlib/issues/1852


hredestig:
not tested on linux / win to try
```{python}
import cobra
import cobra.test
data = cobra.flux_analysis.calculate_phenotype_phase_plane(model, "EX_glc__D_e", "EX_o2_e")
data.plot()
```
and see if plot is interactive

hredestig:
tested in linux min, works as designed

PR #285, opened 2016-10-05
Fixed a bug in ArrayBasedModel bounds setting in Python 3

cdiener:
See https://groups.google.com/forum/#!topic/cobra-pie/-Yfx7atjII8

This was a fun one.
Python 3 calls only `__setitem__` and not `__setslice__` as Python 2
and there was an error in the LinkedArray class in `__setitem__`
where it would assign all the values to each attribute in place
of assigning each value to the correct attribute.

Also added a unit test so that behavior is checked in the future and removed an 
unused assignment in the unit tests detected by Flake8.

### DIFF
--- a/cobra/core/ArrayBasedModel.py
+++ b/cobra/core/ArrayBasedModel.py
@@ -321,7 +321,7 @@ class LinkedArray(ndarray):
         ndarray.__setitem__(self, index, value)
         if isinstance(index, slice):
             for i, entry in enumerate(self._list[index]):
-                setattr(entry, self._attr, value)
+                setattr(entry, self._attr, value[i])
         else:
             setattr(self._list[index], self._attr, value)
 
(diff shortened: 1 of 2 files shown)


cdiener:
Don't merge yet. Still want to test some edge cases for slicing.

hredestig:
great catch this! the failed build should be fixed if you pull the update build-plan from master

cdiener:
Okay fixed some bugs for additional edge cases in slice assignment. Also refactored a bit because `__setitem__` and `__setslice__` basically did the same thing (both handled slices) but with different code and different bugs. Now all the action is happening in `__setitem__`.

If the CI passes this should be done now.

cdiener:
Sorry for that. Now it's okay. Fell free to squash the commits if you want to.

PR #294, opened 2016-10-18
Improvements to the SBML3 validator

cdiener:
Those are some suggestions on how to improve the SBML3 validator.

1. errors are now grouped into "validator", "warnings", "SBML errors" and "other"
2. the function now has a fall through for other exceptions and does not crash anymore as discussed in #287.
3. tests have been adjusted for the new structure
4. the function now has docstring in numpy style

Example:

```python
In [1]: from cobra.io.sbml3 import validate_sbml_model

In [2]: mod, errors = validate_sbml_model("/home/cdiener/Downloads/model_cobra.xml")

In [3]: mod is None
Out[3]: True

In [4]: errors
Out[4]: 
{'SBML errors': [],
 'other': ["'NoneType' object has no attribute 'startswith'"],
 'validator': ['species missing id'],
 'warnings': []}
```


pstjohn:
Thanks for putting the work into this! I wonder though if this is the best approach. The `'NoneType' object has no attribute 'startswith'` error is still largely unintelligible: the real info here is the `species missing id`, but we don't know what species.

Also, a new cobrapy user is going to have to know to call `validate_sbml_model` rather than just `cobra.io.read_sbml_model`. 

I think the best option would be to have `read_sbml_model` implement something like

```python
try:
    parse_sbml_model()
except Exception:
   validate_sbml_model()
```

But there's currently the circular loop where `validate_sbml_model` calls `parse_sbml_model`...

If we don't mind that, then this PR works. We can just add something like the above later. My opinion is that we probably don't want to be messing around with maintaining our own tools for SBML validation, when it seems like there are existing tools for that 😄

cdiener:
I agree with most of the comments. The idea for separating the errors into groups was to give the possibility to show the most important ones. For example I would suggest a normal user to check the comments in "SBML Errors" and "validator", whereas as "warnings" and "other" would be meant for our consumption. 

I am not sure whether we want to call the validator everytime model parsing fails. There are some people that read a lot of models at once and they might not want to call the validator on all of them. I think I would prefer a version like this:

```python
try:
    parse_xml_model(...)
except:
    raise CobraSBMLError("Something went wrong reading the model. You can get a detailed report \
                                          using the `validate_sbml_model` function or using the online validator at \
                                          http://...")
```
And than let the user decide what to do. 

As for the nondescriptive error, we could check for empty IDs in `parse_xml_model` and raise a `CobraSBMLError` with the species ID that is caught by `validate_sbml_model`... 

Let me know what you think. I will happily implement your suggestions (the PR is also open for maintainer commits).

cobra/io/sbml3.py
```diff
@@ -548,7 +548,13 @@ def read_sbml_model(filename, number=float, **kwargs):
                 xmlfile.write(outfile, encoding="UTF-8")
             filename = outfile.name
         return read_sbml2(filename, **kwargs)
-    return parse_xml_into_model(xml, number=number, **kwargs)
+    try:
+        return parse_xml_into_model(xml, number=number, **kwargs)
+    except Exception:
```
cdiener: I think just `except:` is fine here.

cdiener: Uh, didn't know about that. In that case nevermind.

cdiener: I can add a test for the new try/except and maybe add an id checker to `parse_xml_model` until tomorrow if that's fine.

cdiener:
Interestingly `io/sbml3.py` had a `get_attrib` function that can check for required attributes. It just was not used inside the `clip` calls. Changed that so that missing IDs raise the correct error in `parse_xml_model`. Now the errors are a bit more informative:

```python
In [1]: import cobra

In [2]: cobra.io.sbml3.validate_sbml_model("/home/cdiener/Downloads/model_cobra.xml")
Out[2]: 
(None,
 {'SBML errors': ["required attribute 'id' not found in tag '{http://www.sbml.org/sbml/level3/version1/core}species' with name '<--[c]'"],
  'other': [],
  'validator': ['species missing id'],
  'warnings': []})
```

Also added a test for the new "Something went wrong" error.

hredestig:
if you change the base branch to devel and pull from it the ci should pass again

cdiener:
Closing since the history is messed up. Will send another pull

PR #299, opened 2016-10-21
Changes to SBML3 validator.

cdiener:
This replaces #294 and includes the changes discussed there on top of the repaired CI to have a cleaner history.


cdiener:
Thanks @hredestig! Next one will be to devel I promise ^_^'

PR #305, opened 2016-10-31
Feat/pytest

hredestig:
this switches cobrapy from unittest to pytest and adds benchmarking - comments/review welcome!


hredestig:
cheers for the review @pstjohn I added a couple of words on pytest and pytest-benchmark in the contribution guidelines. Those main landing pages linked there are also really good quick intros imo

PR #309, opened 2016-11-08
feat: add optlang solver interface to model

hredestig:
- add optlang solver interface to the model class to enable direct
  updates of the solver upon changes to the model

this replaces the entire defunct optlang_model branch (to clean up the entangled history)

note that the PR is to devel-2 and not to devel to isolate these big changes until they are ready for prime time.


cdiener:
Might the appveyor failure be due to the last temp file changes in optlang (not being kept around that much anymore)?

hredestig:
yes problem with the tempfiles for windows.. is being fixed here https://github.com/biosustain/optlang/pull/34

cobra/core/Model.py
```diff
@@ -266,9 +421,49 @@ def optimize(self, objective_sense='maximize', **kwargs):
...
+                    {'minimize': 'min', 'maximize': 'max'}[objective_sense]
+                self.solver.optimize()
+                self.solver.objective.direction = original_direction
+            else:
+                self.solver.optimize()
+            solution = solution_type(self)
+            self.solution = solution
+            return solution
+
+    def solve(self, solution_type=LazySolution, *args, **kwargs):
```
pstjohn: This seems confusing to me. Why do we need a `model.optimize` and a `model.solve`?

pstjohn:
Is there any reason we're keeping the legacy solvers around in this branch?

cdiener:
I think its for backwards compatibility. The idea was to slowly deprecate the old interface... Also optlang currently (officially) only supports glpk and cplex.

cobra/core/Model.py
```diff
@@ -46,18 +57,65 @@ def __init__(self, id_or_model=None, name=None):
...
+                interface = config.solvers[value]
+            except KeyError:
+                raise not_valid_interface
+        elif isinstance(value, types.ModuleType) and hasattr(value, 'Model'):
+            interface = value
+        else:
+            raise not_valid_interface
+        for reaction in self.reactions:
+            reaction._reset_var_cache()
+        self._solver = interface.Model.clone(self._solver)
```
KristianJensen: That was how solvers were changed previously in cameo. The problem is though, that any additional constraints or variables (e.g. flux ration constraints or auxiliary variables) will be lost by repopulating the solver. Cloning the solver keeps these extra variables and constraints but is a bit slower. Since you would not usually change the solver back and forth a lot, we decided that a few extra seconds were ok to add this functionality.

cobra/core/Model.py
```diff
@@ -230,9 +308,85 @@ def add_reactions(self, reaction_list):
...
 
+        # from cameo ...
+        self._populate_solver(reaction_list)
+
+    def _populate_solver(self, reaction_list, metabolite_list=None):
+        """Populate attached solver with constraints and variables that
+        model the provided reactions.
+        """
+        constraint_terms = AutoVivification()
+        if metabolite_list is not None:
```
KristianJensen: Each metabolite has a constraint, corresponding to a row in the Sv = 0 equation.

cobra/core/Reaction.py
```diff
@@ -76,14 +78,282 @@ def __init__(self, id=None, name='', subsystem='', lower_bound=0.,
...
+    def _reset_var_cache(self):
+        self._forward_variable = None
+        self._reverse_variable = None
+
+    @property
+    def objective_coefficient(self):
+        """ Get or set the objective coefficient (float)
+        """
+        if self.model is not None and self.model.solver.objective is not None:
+            coefficients_dict = \
```
KristianJensen: One of the advantages of using optlang is that you can set the objective to an arbitrary expression of reactions using `model.objective = expression`. It would be possible to update the reaction coefficients in the model.objective setter, but this would require the same sympy expression parsing. For cameo did not intend reaction.objective_coefficient to be used, so we didn't want to parse the sympy expression every time the objective was changed and this getter was implemented instead to maintain cobra-compatibility.

cobra/core/Model.py
hredestig: agreed, that's also legacy, added deprecationwarning to indicate this

cobra/core/Metabolite.py
```diff
@@ -40,6 +39,29 @@ def __init__(self, id=None, formula=None, name="",
...
+        self._id = value
+        self.model.metabolites._generate_index()
+
+    @property
+    def constraint(self):
+        if self.model is not None:
+            return self.model.solver.constraints[self.id]
+
+    @property
+    def constraint(self):
```
pstjohn: did you mean to repeat this three times? Also, could we get docstrings for new methods like this?

cobra/core/Metabolite.py
```diff
@@ -167,6 +211,21 @@ def summary(self, **kwargs):
         except ImportError:
             warn('Summary methods require pandas/tabulate')
 
+    def _repr_html_(self):
```
pstjohn: If its possible, could we make stuff like this a different pull request? This one seems like a pretty monolithic 'cobra becomes cameo' one, where I think it would be easier to digest if we did one for solvers, one for `___str__` methods, etc.

cobra/config.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2014 Novo Nordisk Foundation Center for Biosustainability, DTU.
```
cdiener: Apache and Gpl are not compatible. We would have to change the license to GPL/LGPL here.

cobra/core/Model.py
cdiener: Constraints for Sv=0 are set below but here you just set an empty expression "0" for each metabolite. Is this supposed to be the right side of Sv=0?

cobra/core/Model.py
cdiener: Switching out solvers is a pretty common operation in cobrapy (seeing which is fastest, switching from a LP to QP solver etc) and had no overhead at all previously. With clone switching becomes much slower than reading the model from SBML and solving it together. I think the users wouldnt like that....

cobra/core/Model.py
cdiener: I think optimize is actually the correct term here. Solve is pretty ambivalent and might even imply that there is a unique solution. But I have no strong feelings on that...

cobra/core/Model.py
KristianJensen: This line just initializes the Constraint objects (one for each metabolite). Line 385 sets the coefficients of the Constraint expressions (lhs, basically the S-matrix). The reason that the coefficients are not set here is for performance, since the `set_linear_coefficients` method operates directly on the solver and circumvents all sympy operations.

cobra/core/Model.py
KristianJensen: I've added some functionality in optlang that clones via an LP format intermediate. This can leverage the solver's own IO functions (usually written in C) so changing solvers will be much faster (it takes ~400 ms to change the solver of iJO1366)

cdiener: Well that sounds awesome! Is it in one of the optlang branches. I have to check that out.

cobra/core/Model.py
cdiener: Ah ok got that. But is it really necessary, since the for loop also adds constraints with metabolite.id only if they are not in the constraints (see the if/else before setting coefficient_terms)? I'm probably wrong but deleting the first lines in the function adding the metabolite constraints shouldnt change anything. Of course _populate_solver always has to be called with a non-empty reaction list.

KristianJensen: Usually that's true, but in the case where you have a metabolite that is not in any of the reactions in reaction_list you need to add the empty constraint separately. (So my previous answer was actually not correct. I was referring to the Constraint construction in line 363, this one is only there to make sure all metabolites get a constraint, even if they don't have any reactions)

cdiener: But why. The solver will ignore all zero rows anyways. It just gives the presolver a little more work. If it eventually does participate in a reaction this will trigger _populate_solver again which will add the constraint in the first reaction where it is used. Metabolites that do not participate in any reaction should not affect the solver at all...

KristianJensen: The metabolite cannot be added to a reaction using the `reaction.add_metabolites` method if it doesn't have a constraint. Having metabolites that don't participate in reactions is not that common, and the extra work for the solver is pretty close to nothing, so the most straightforward way to handle it was to just make sure that all metabolites have a corresponding constraint.

cdiener: Ok I understand. I think it would be nicer to have only one function that updates the solver object (like _populate solver or another one) and that only works with a reaction_list as this will make the lazy solver update easier as well. For instance the set_linear_coefficients code block in add_metabolites is pretty much the same that happens in _populate_solver only that is doesn't create the variables (but you coud add a check in populate_solver) and this happens in a lot of functions and will make it hard to find bugs or react to breaking changes in the optlang API.

cobra/core/Reaction.py
cdiener: I agree. Objective coefficients for reactions do not make sense for arbitrary objectives. They should be deprecated entirely since the objective is a property of the model anyways...

cobra/core/Model.py
cdiener: Thanks for walking me through that though! Realky appreciate it. I think for now we could pull it like that and try make our suggestions for refactoring via pull requests to devel-2.

pstjohn:
On deprecating the old solvers: my feeling is that if we can get all the internal references to be consistent (pFBA, FVA, MOMA, etc.), we don't necessarily need to keep up the old solver interface... I would expect it might cause more problems than it would solve. 

I think its safe to assume the majority of users don't rely on the underlying solver interface (with a few exceptions as we talked about in the video chat), those that do could always keep using an older version until they port their code over. Maybe put in some hooks to give informative errors if someone tries to use a coin solver or gurobi, for instance.

cdiener:
The more I look at the code, the more I agree. Gurobi is already in
optlang. Maybe one of the optlang collaborators can comment on its state. I
think if we can get mosek into optlang in a near future it would cover most
used solvers I have seen on the mailing list and in the Github issues... If
somebody really needs the old interface we can still point them to the
wheels for 0.5.4.

cobra/config.py
hredestig: just an oversight, cleaning out the apache mentions, no intention to change licenses anywhere

hredestig:
I totally agree we shouldn't keep the old solver interface around, however, updating the current algorithms to use optlang interface is probably more effort than just moving the cameo implementations to cobra - but let's handle/compare/discuss that in a different pr

cobra/core/Metabolite.py
hredestig: oops, must be a rebasing error, docstring added

cdiener:
Thanks @hredestig. Will send my suggestions in form of PRs to devel-2 in the future. I think some stuff can be moved from cameo, but there is some use in revising them. Like I said I think MOMA can be simplified even from the cameo implementation and FVA in cameo still uses the slower version (switching between min and max instead of doing all mins first, then the maxs). I think we should also think about what functions from cameo we could move to cobrapy and vice versa. LMOMA and ROOM would be good ideas to get to cobrapy. Maybe RobustOptKnock would be better suited for cameo given that cameo *is* the strain design package but I don't know what @zakandrewking thinks about that...

hredestig:
Yup, completely agree on all those points

zakandrewking:
I would love to see the strain design algorithms moved to cameo. I never finished implementing RobustKnock because I hit a bug that I suspect is related to the solver interface. Hopefully an OptLang implementation will fix this.

cdiener:
Awesome, I am actually in need of a dualising method in optlang as well and as far as I have seen it seems to be on the way. This should make bi-level methods much easier to implement.

phantomas1234:
This is going really! Thanks all of you. With all these additions I think we'll go way beyond what cameo tried to achieve.

PR #310, opened 2016-11-08
Fix/enable inline pytest

hredestig:
without explicitly importing the fixtures, pytest don't find them when running outside the package repo


hredestig:
thanks for checking! when building wheels you may want to test the actually installed package a safe way to do that is to move out of the directory. Also, previously you could do 
```
import cobra.test
cobra.test.test_all()
```
but that was actually broken before, this makes it work pretty much like before pytest

PR #312, opened 2016-11-12
feat: (wip) adding context managers for model manipulation

pstjohn:
Here's a first attempt at an implementation for context managers in cobrapy.
I stripped down the TimeMachine class from cameo, (I'm not sure we'll use all the features). Big thanks to @cdiener and @phantomas1234 for the suggestions.

We'll have to mix this in with the optlang interface, but I was having some issues there so I figured I'd hash out a working version on devel-2 and worry about merging in to the optlang once that gets merged.

One thing I realized we'll want is to make sure we can have nested contexts. I.e from the test,
```python
bounds0 = model.reactions[0].bounds
with model:
    model.reactions[0].bounds = bounds1
    with model:
        model.reactions[0].bounds = bounds2
        assert model.reactions[0].bounds == bounds2
    assert model.reactions[0].bounds == bounds1
assert model.reactions[0].bounds == bounds0
```
It might seem superfluous, but this way we can feel free to use contexts in cobra methods without worrying whether a model that gets passed in is already inside a context.

i.e., `func_that_uses_context(model)` would be fine if called as
```python
with model:
    func_that_uses_context(model)
```

Right now I'm only doing reaction bounds. We'll want to make sure this also does objective functions, but that will require some more thought with optlang. For now at least, this should serve as a rough outline of how we could implement context managers.


hredestig:
Would you be okay with merging this **after** we merge #309 and suggesting changes to TimeMachine instead? We can of course replace that class with your HistoryManager but I think it would be great if cameo didn't have to have a near-identical class as well, it would recycle more code and facilitate for developers that work on both cobrapy and cameo

pstjohn:
Yeah thats fine, lets try to do another video chat prior to merging. That might help get all of these aligned.

pstjohn:
I don't think it would be too hard to check to see we're not resetting a
previous state... we'd just use 'bookmarks' that are unique to each
attribute, and have historymanager handle it appropriately. _history is an
orderedDict, so lookups are O(1).

With populate_solver, I wonder if we want the full history... I'm mainly
wondering about adding/removing reactions and how that impacts the order of
operations

On Nov 12, 2016, at 7:42 AM, Christian Diener <notifications@github.com>
wrote:

*@cdiener* approved this pull request.

Awesome work. Looks much cleaner now.

I think checking for unnecessary undos would be slow since you have to walk
through the entire history. For that you would a state based approach,
which I think is not necessary here. That approach, however, would be
necessary for the lazy solver update. Here I would just log the ids of
reactions and metabolites that have to be added and let _populate_solver
figure out what has to be done.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/opencobra/cobrapy/pull/312#pullrequestreview-8298955>,
or mute the thread

cdiener:
I don't think we need the history for the lazy solver. If you just know
that reaction 1 changed you do the following on update:

- reaction in solver but not model: delete
- reaction in model but not solver: add
- reaction in both: update reaction coefs and bounds (might create a
performance hit here)

The alternative would be to have basic functions to add, delete, change
coefs, change bounds for a reaction/variable in the solver and manage that
with the time machine/history context by calling execute on solver update.
All the basic operations have an associated decorator so that Reaction's
methods trigger the appropriate modification.

Actually rereading alternative 2 does not sound too bad and could probably
be used in the flux_analysis functions.

Adding and deleting is a problem. For now we should probably only ignore
repeated calls. For bounds it's much easier because you only use the last
call for a particular bound. Coefficients would be difficult as well.

pstjohn:
TODO:

- [ ] objective_coefficient? Not sure how this is currently done with optlang.
- [ ] Add/remove reactions
- [ ] Add/remove metabolites

cdiener:
`objective_coefficient` and the old way of returning `model.objective` do not make sense anymore with optlang. They should be deprecated.

cobra/util/context.py
```diff
@@ -0,0 +1,112 @@
...
+
+    try:
+        return obj._model._contexts[-1]
+    except (AttributeError, IndexError):
+        pass
+
+    return None
+
+
+def resettable(f):
```
hredestig: if we had resettable(f, attr=None) could we not get resettability for model.solver.objective as well then using
```{python}
@resettable(attr="solver")
@setter
def objective:
   ...

def wrapper(self, new_value):
   ...
   if attr:
      old_value = getattr(getattr(self, attr), f.__name__)
   ...
...
or something like that?

File: cobra/util/optlang.py [hunk 2]
```diff
@@ -0,0 +1,46 @@
+def split_bounds(lower_bound, upper_bound):
```
pstjohn: I do use make_forward_and_reverse_bounds (split_bounds) in cobra.core.Model, so having that separated made some sense. I can certainly move the other over, but it seemed odd to have it floating in `Reaction.py`.
Happy to do whichever though

cobra/core/Model.py
```diff
@@ -16,9 +15,10 @@
 import types
 import optlang
 from sympy.core.singleton import S
-from functools import partial
-from cobra.util import AutoVivification
-from cobra import exceptions, config
+from ..util import AutoVivification
```
pstjohn: They appeared in other places in the package, I don't really have a preference (didn't know they were discouraged). I'll change them

cobra/util/optlang.py
hredestig: The separation is good, though nothing wrong with extra 'floating' function definitions in a Reaction module (although correction that module name to reaction will reduce the feeling that it's only for the Reaction class, for a different cleanup PR)

hredestig:
I think this looks good now, is it ready to merge?

cobra/util/context.py
```diff
@@ -42,7 +42,7 @@ def __exit__(self, type, value, traceback):
 
     def execute(self, bookmark=None):
         """ Execute up to the given entry in the history.
-        If no entry is specifed, executes the last entry.
+        If no entry is specified, executes the last entry.
```
pstjohn: Honestly we could remove the bookmark argument if we're not going to use it.
I included it because I wondered if we'd need it for the nested managers, but ended up using independent stacks instead.

pstjohn:
I'd say we're RTM. We'll need to clean up some things in devel-2 in the future, but lets see where we can start using these.

Thanks for wrapping this up!!

cobra/core/Model.py
```diff
@@ -552,6 +552,7 @@ def objective(self):
         return self.objective_reactions
 
     @objective.setter
+    @resettable
```
hredestig: No not a problem but still tinkering with objective_reactions in #324 so might as well leave it like this for now.

hredestig:
@pstjohn there coverage of context.py is not perfect, could you have a look and then remove the bookmark argument if we're not gonna use it?

cobra/util/context.py
pstjohn: I'm on board with this. Is this still something you'd use after the recent objective change?

pstjohn:
This is a pretty major gutting of the `HistoryManager`, its now just a stack model.
I suppose its elegant, `model._contexts` is not a stack of stacks 😄

Issue #295, opened 2016-10-19
add functionality for flux sampling


cdiener:
Ongoing progress can be seen in https://github.com/cdiener/cobrapy mostly in `flux_analysis/sampling.py`. 

For now the provided algorithms will be Artificial Centering Hit-and-Run and [optGPSampler](http://dx.doi.org/10.1371/journal.pone.0086587).

As soon as I have some benchmarks they will appear here. 

Any features you consider important? Let me know.

cdiener:
Docs and tests (98%) are now pushed.

HAHerrmann:
Hi, I was hoping to use the cobra.flux_analysis.sampling functionality you described above but can't seem to find it anywhere. Could you please let me know where I can find the flux_analysis/sampling.py file?

cdiener:
Hi @HAHerrmann, flux sampling is currently still in development, that is why is has not been merged into cobrapy yet. The branch that contains is can be found [here](https://github.com/cdiener/cobrapy/tree/sampling). To set up a local test environment where you can use it, use the follwoing steps:

1. `git clone` the branch to a local directory
```bash
git clone -b sampling https://github.com/cdiener/cobrapy
cd cobrapy
```

2. Install locally with `pip` (I dont your setup, should be `pip` or `pip3` for you). This will require `cython` in order to generate the glpk bindings. If you have cplex, gurobi or mosek this step can be skipped entirely.

```bash
pip install --editable .
```

3. Staying in the same directory you can now use flux sampling (if you leave the directory your "normal" cobrapy install will be used).

```python
from cobra.flux_analysis import sample
```

For usage instructions [see the docs in the sampling branch](https://github.com/cdiener/cobrapy/blob/sampling/documentation_builder/sampling.ipynb).

HAHerrmann:
@cdiener , that worked fine. Thanks for your quick and thorough reply!

cdiener:
Glad it worked. If you find any bugs please post them here :)

phantomas1234:
Would this be ready now for the devel-2 branch in cobrapy?

cdiener:
Yeah definitely will open a pull request. This will also help me to see what is needed to make it work on Windows.

cdiener:
Flux sampling has now been merged into `devel-2`. Any new bugs, feature requests and comments are better addressed in individual issues. Thanks!

HAHerrmann:
[...]

PR #315, opened 2016-11-15
[WIP] Add flux sampling

cdiener:
This is the initial PR for flux sampling as described in #295. Do not merge until appveyor passes please. Also, can somebody that knows more about Sphinx check whether I added the docs correctly?


cdiener:
Travis failures are not caused by sampling as far as I can see ("no solvers found"). Having some problem setting up tests for sampling on appveyor since it requires scipy which does not provide wheels for windows. As I see it the options are:

1. leave it like it is and sampling tests will skipped (disadvantage is that we don't see failures in sampling on windows).
2. Use numpy and scipy wheels from [here](http://www.lfd.uci.edu/~gohlke/pythonlibs/) (disadvantage is that this requires specifying numpy andf scipy versions in appveyor which have to be changed by hand everytime numpy and scipy get updated to work).
3. Substitute/add miniconda (already installed on appveyor) in the build matrix (disadvantage is the use of conda which is not the same as base python)

Midnighter:
If you're going the miniconda route, I've recently set up template repositories for setting up a build matrix on [AppVeyor](https://github.com/Midnighter/python-conda-appveyor) and [TravisCI](https://github.com/Midnighter/python-conda-travis). Maybe those can help you quickly set it up.

cdiener:
Ok thanks. How much overhead is there in switching python versions on conda in appveyor) When I used conda on my machine getting changing the python version triggered an update of all packages which took quite a while. Maybe one could just use conda for the entries that correspond to the "native" conda python versions (2.7 and 3.5).

Midnighter:
At the moment I do call an update of the conda environment ([see this codeline](https://github.com/Midnighter/python-conda-appveyor/blob/master/ci/appveyor/conda_setup.py#L18)) which could easily be skipped. Overall, doing the update and setting up a specific Python version and installing the required packages takes about 1.5 min.

hredestig:
are these unofficial wheels http://www.lfd.uci.edu/~gohlke/pythonlibs/ an option?

cdiener:
Those are the ones i linked to in point 2.

hredestig:
We should test as close to the recommended way to install cobrapy, and with that currently not being conda, I'd go with option 2 even if we have to update the appveyor.yml from time to time. If this turns out to be too much work we can always change strategy.

cdiener:
If we go with 2 I think we should set up a mirror repo with the wheels inside (or put them on zenodo or something). Gohlke goes through some lengths to not show direct download links, so we should respect that. This would also protect against breaking CI builds if the versions change.

pstjohn:
How do we think the majority of windows users would install python/scipy? If thats via conda, I’m fine with option 3. I use conda on my mac, its gotten much better in recent years. (I still think we should distribute via pip: its trivial to pip install a package with conda's python)

We'll test against base python on travis, so I doubt there will be issues that we'll miss.

cdiener:
Good question. Don't have official numbers, but I think conda is the standard for scientific Python on Windows those days...

hredestig:
If we build wheels using conda python, are these then equivalent to the wheels that are built using regular python?

cdiener:
@hredestig They shoudl be but we could also just add additional entries in the build matrix that use conda but do not build the wheels, so that conda is only used for testing...

hredestig:
Using Gohlke's binaries appears less of a change to the current tested and working setup than switching to conda. We can host the binaries here https://github.com/opencobra/pypi_cobrapy_travis to get a static download link. Acceptable? I can implement it.

cdiener:
Unfortunately the the Gohlke wheels include mkl and are thus to big (113 MB) for normal Github (max. 100 MB), so we would need to think about something else. Zenodo would be an option.

Midnighter:
Maybe git-lfs is an option?

* [Collaborate](https://help.github.com/articles/collaboration-with-git-large-file-storage/)
* [Configure LFS](https://help.github.com/articles/configuring-git-large-file-storage/)

cdiener:
There is no Git LFS client for Windows...

hredestig:
alright, zenodo or dropbox then

cdiener:
Last question before I start implementing it. Do we want numpy and scipy for all entries in the build matrix or only for some (maybe 2.7 and 3.5)?

hredestig:
Alright, cheers for putting it together, if we anyway install scipy/numpy then might as well do for all entries.

cdiener:
Hm, I can't get it to work with the wheels from Gohlke. Maybe some file corruptions because I downloaded them on Linux? Either way it does not seem to be a good solution.

cdiener:
Okay this is a battlefield here. I think we all agree that this won't be merged as is. I'll let it open to test some strategies for appveyor and will close and send a new rebased version when everything works...

PR #317, opened 2016-11-24
Fix for failing test on Windows

cdiener:
`LazySolution` uses the current time stamp to judge whether a solution has changed. On windows `time.time` has lower resoultion than on Linux (officially 1/60 of a second). If resolving takes less than the solution gets the same timestamp. Introducing a sleep period > 1/60 s fixes the test, however for the future we might not want to rely on the time stamp for that. Maybe one could hash the solution and compare hashes rather than timestamps.

### DIFF
--- a/cobra/test/test_solver_model.py
+++ b/cobra/test/test_solver_model.py
@@ -49,8 +49,10 @@ def solved_model(request, model):
 
 class TestLazySolution:
     def test_self_invalidation(self, solved_model):
+        from time import sleep
         solution, model = solved_model
         assert abs(solution.f - 0.873921506968431) < 0.000001
+        sleep(0.05)
         model.optimize()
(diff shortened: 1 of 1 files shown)


phantomas1234:
Can we maybe use time.clock() on windows? http://stackoverflow.com/questions/1938048/high-precision-clock-in-python

cdiener:
Yes you can but have to add an OS test. Either way you probably want to
leave the sleep statement in the test just to make sure.

cdiener:
Would you like me to add the "use clock() on Windows" part to the PR?

cdiener:
Hi I'm also in favor of merging the test fix for now. We can still figure out later what to do with the hash/timestamp since it is not critical (very hard to trigger the bug in real use cases).

PR #318, opened 2016-11-25
feat: entities serialisation

sgalkina:


cdiener:
Thanks, looks cleaner. Do you think it makes sense to make those methods private or refactor them to its own utils module as proposed by @pstjohn?

sgalkina:
@cdiener thanks for the suggestions. These methods were refactored out because of the use for them in a third-party service which needs conversions of reactions, genes and metabolites from/to dictionaries, thus they need to be public. I'm not sure about creating utils module for them, `io` sounds like a logical place for serialisation tools.

cdiener:
Ok get it. If they are public we should also add documentación for them. May I ask what third party service we are talking about?

PR #319, opened 2016-11-25
fix: model copy and dict search

hredestig:
model.notes and model.annotatation object were only copied by reference
which is unintuitive, let's copy those

querying dictlist with a unicode string triggered bug, use six

### DIFF
--- a/cobra/core/Model.py
+++ b/cobra/core/Model.py
@@ -96,10 +96,13 @@ class Model(Object):
         than deepcopy
         """
         new = self.__class__()
-        do_not_copy = {"metabolites", "reactions", "genes"}
+        do_not_copy = {"metabolites", "reactions", "genes", "notes",
+                       "annotation"}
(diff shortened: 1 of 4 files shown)


cobra/core/Model.py
```diff
@@ -96,10 +96,13 @@ def copy(self):
         than deepcopy
         """
         new = self.__class__()
-        do_not_copy = {"metabolites", "reactions", "genes"}
+        do_not_copy = {"metabolites", "reactions", "genes", "notes",
+                       "annotation"}
         for attr in self.__dict__:
             if attr not in do_not_copy:
                 new.__dict__[attr] = self.__dict__[attr]
+        new.notes = copy(self.notes)
```
sgalkina: deepcopy would probably fit better, because nested objects would also need copies, not references

hredestig: okidok

PR #322, opened 2016-11-28
Add flux sampling

cdiener:
This time it has been debugged on windows using my own test branch that already included #321. So this one should work.

Cheers


cdiener:
Hi @hredestig, I think this one is working for now. Maybe we could merge to prevent future merge conflicts in `test_flux_analysis.py`.

PR #324, opened 2016-12-06
(wip) Fix/refactor objective coefficient

hredestig:
reaction.objective_coefficient makes less sense when the objective can
be made up of more than just reactions and their coefficients. However,
we still want the option to easily assign the model.objective a set of
reactions and their coefficients.

This is an attempt to reduce this inconsistency by moving the property
reaction.objective_coefficient to model.set_reaction_coefficient and
model.reaction_coefficient where model._objective_reactions is a set
that keeps track of the reactions that have been assigned as objectives
and the coefficient being extracted from the solver.


cdiener:
Since it adds new behavior some tests might be good.

hredestig:
Still not so sure about this.. Now model.reaction_coefficient still needs to interact with the solver.objective, a perhaps better alternative would be to keep _objective_reactions a dictionary and let Model maintain these until _populate_solver

cdiener:
What happens if the objective is quadratic or non-linear?

hredestig:
I was thinking that reactions and their coefficients is syntactic convenience currently widely used for linear objectives and we need to support that functionality but could leave out this convenience for other objectives, just need to get rid of inconsistency with model.objective, and the use of reaction.objective_coefficient. Needs more thinking, feel free to change or make a separate contribution.

hredestig:
need to do this differently, closing for now

Issue #328, opened 2016-12-12
Flux sampling issues


h8dinh:
Hi,

I have tried the new cobrapy features cobra.flux_analysis.sampling (flux sampling, used sampling.py file, version in branch devel2 on Wednesday, December 7). There are two issues I have seen when running the simulation with a modified E. coli genome scale model iJO1366, glucose uptake -20, anaerobic. I tries to run simulation for the same case two times to compare the two output distribution.

1) First issue, I tried to see if I use the sampling method two times, is there any different between the two cases? Using method='optgp', processes=4, sample=1000 or 20000 yield exact same distribution, which worries me. However, using 'arch' sample=1000 yield different distribution.
(e.g. code: sampling.sample(model, 20000, processes=4, method='optgp', solver='gurobi'))

2) Second issue, when I tried to use 'arch' sample=10000 or 20000 two times, the first time it runs through ok, but the second time it yields an error:
RuntimeError: maximum recursion depth exceeded
(e.g. code: sampling.sample(model, 10000, method='arch', solver='gurobi'))

Additional info:
- I implement the method by simply copying the sampling.py python file to my local updated cobrapy master branch. Then, I just use sampling module by import cobra.flux_analysis.sampling as sampling.

h8dinh:
For the second issue, when I delete the object containing the distribution before assigning the sampling output to that object, somehow the error of recursion depth go away (not sure if this work around is replicable).

E.g.
del wt
(simulation)
wt = sampling.sample(model, 10000, method='arch', solver='gurobi')

h8dinh:
For the first issue, there is something wrong with setting multiple core. When I run optgp with 1 core, everything is ok. When I run with 4 cores, I got two exact same distribution.

For sample=1000, method='optgp', solver='gurobi'
1 core:
![image](https://cloud.githubusercontent.com/assets/15752681/21121471/772a7bb8-c081-11e6-94fb-7565af328375.png)


4 cores:
![image](https://cloud.githubusercontent.com/assets/15752681/21121468/6b3189d2-c081-11e6-950e-de622b12d352.png)

cdiener:
Hi, thanks for the testing! I pretty much know what's up with the second option (has to do how we reset the step when sampling produces bound violations). The first one is interesting though. It would help if you could give me the following information:

1. Your Operating System (Windows, Mac, Linux)
2. Your Python version

Also it would be great if you could lend me your model. For instance if you have your model already in code and it is named `model` you could do

```python
[...]

PR #329, opened 2016-12-13
Fixes to sampling

cdiener:
Various fixes related to flux sampling. Resolves issues in #328.

### DIFF
--- a/cobra/flux_analysis/sampling.py
+++ b/cobra/flux_analysis/sampling.py
@@ -10,6 +10,7 @@ from ..solvers import solver_dict, get_solver_name
 from copy import deepcopy
 from multiprocessing import Pool, Array
 import ctypes
+from time import time
 
 BTOL = np.finfo(np.float32).eps
 """The tolerance used for checking bounds feasibility."""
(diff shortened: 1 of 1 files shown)


cobra/flux_analysis/sampling.py
```diff
@@ -355,10 +359,11 @@ def _sample_chain(args):
 
     center and n_samples are updated locally and forgotten afterwards.
     """
-    sampler, n = args       # has to be this way to work in Python 2.7
+    sampler, n, idx = args       # has to be this way to work in Python 2.7
     center = sampler.center
+    np.random.seed(int(time() * idx + idx) % np.iinfo(np.int32).max)
```
hredestig: so simply `np.random.seed()` doesn't do what its supposed to according to the docs?

cdiener: Honestly, it would probably would in most cases. The thing is that it may use system time and processes may be spawned at the same time. So if the clock resolution is low all processes might be seeded to the same value in the worst case. The code here just ensures that this is impossible even though I agree it's a bit convoluted...

PR #338, opened 2017-01-11
Add helper functions for optlang solvers that work with new context

cdiener:
This is a proposal to pave the way for the optlang port of `flux_analysis.py`. The functions defined in `cobra.util.solver` help to temporarily add or remove variables and constraints to the solver. It also includes helpers to extract all available solver interfaces from `optlang`.

This implied some minor changes in `cobra.Model` in setting `Model.solver` and `Model.optimize`. 

The "solver" keyword is now respected in `Model.optimize` but requires a "optlang-" prefix for the optlang solvers.

### DIFF
--- a/cobra/util/solver.py
+++ b/cobra/util/solver.py
@@ -1,0 +1,153 @@
+"""This module includes additional helper functions for the optlang
+solver object that integrate well with the context manager, meaning that
+all operations defined here work well with the model context manager.
+
+The functions defined here together with the existing model functions should
+allow you to implement custom flux analysis methods with ease."""
+
+from cobra.util.context import get_context
+from functools import partial
+import optlang
+
+
+class SolverNotFound(Exception):
+    None
+
+
+solvers = {match.split("_")[0]: getattr(optlang, match)
+           for match in dir(optlang) if "_interface" in match}
+
+
+def get_solver_name(mip=False, qp=False):
+    """Selects a solver for a given optimization problem in a reproducible
+    manner.
+
+    Parameters
+    ----------
+    mip: string
+    Does the solver require mixed integer linear programming capabilities?
+    qp: string
+    Does the solver require quadratic programming capabilities?
+
+    Returns
+    -------
+    name: string
+    The name of feasible solver.
+
+    Notes
+    -----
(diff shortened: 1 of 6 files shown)


cdiener:
The codecov thing is incorrect, see comment on gitter.

cobra/util/solver.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    context = get_context(model)
+
+    if len(what) > 0:
+        model.solver.add(what)
+        if context:
+            context(partial(model.solver.remove, what))
+
+
+def remove_from_solver(model, what=[]):
```
pstjohn: Would we ever want to *not* check for a context? I.e., should we just be replacing all calls to model.solver.add() and model.solver.remove() with a context'ed version?

cobra/util/solver.py
```diff
@@ -0,0 +1,170 @@
...
+    # variable > -expression
+    constraints = [
+        # positive value constraint
+        model.solver.interface.Constraint(expression - variable, ub=0,
+                                          name="abs_pos_" + name),
+        # negative value constraint
+        model.solver.interface.Constraint(expression + variable, lb=0,
+                                          name="abs_neg_" + name)
+        ]
+    model.solver.add(constraints + [variable])
```
pstjohn: couldn't you just call `add_to_solver` above, and not duplicate the context code?

cobra/util/solver.py
cdiener: model.solver is an `optlang.Model` and `add` is defined and implemented in optlang :/

cobra/util/solver.py
cdiener: You're right! Silly me :) will change it

cobra/util/solver.py
pstjohn: right, but we could replace all calls to `model.solver.add()` with `add_to_solver`. We could even consider making it a cobra.Model method, `model.add_to_solver(self, ...)`

cdiener: Ah ok, I misunderstood. Yes, that would be feasible. Also thought about making it methods of `Model` but remembered that you were against adding too many new functions in `Model.py` :smile:

PR #340, opened 2017-01-12
Changing default query behavior

pstjohn:
I mentioned this a while back in the gitter, figured it was an easy enough change I'd just put the PR in to ask how other people feel about this.

I personally don't query on 'id' much, I'll usually use the query function to filter the reactions or metabolite list with a lambda function.

```python
>>> model.reactions.query(lambda r: len(r.reactants) > 1)
```

or 

```python
>>> model.metabolites.query(lambda m: m.charge)
```

Having to pass 'None' each time seems like a bit of a strange default behavior. Would anyone's code be significantly broken by a change like this?

### DIFF
--- a/cobra/core/DictList.py
+++ b/cobra/core/DictList.py
@@ -54,7 +54,7 @@ class DictList(list):
         """
         return [getattr(i, attribute) for i in self]
 
-    def query(self, search_function, attribute="id"):
+    def query(self, search_function, attribute=None):
         """query the list
 
         search_function: used to select which objects to return
@@ -66,7 +66,7 @@ class DictList(list):
             * a function which takes one argument and returns True
               for desired values
 
-        attribute: the attribute to be searched for (default is 'id').
+        attribute: the attribute to be searched for (default is None).
                    If this is None, the object itself is used.
 
         returns: a list of objects which match the query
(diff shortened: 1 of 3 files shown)


cobra/core/DictList.py
```diff
@@ -54,7 +54,7 @@ def list_attr(self, attribute):
         """
         return [getattr(i, attribute) for i in self]
 
-    def query(self, search_function, attribute="id"):
+    def query(self, search_function, attribute=None):
```
hredestig: While already have a quite a few statements like `model.reactions.query('EX')` in my notebooks, they would all break with this modification.. Perhaps a good solutions for both would be to add that if search_functions is a RE, and attribute is None then attribute is 'id'?

pstjohn: Yeah I was wondering that -- I'll think about it a bit more. Maybe if its a string, you search ID, if its a function, you use the object..

My main concern is that it seems odd to have to pass `None` as an explicit argument.

Not that it really matters, but I think the more generalizable way to get boundary reactions would be 
```python
>>> model.reactions.query(lambda x: x.boundary)
```

That's how I try to do it in the cobrapy methods (i.e., `summary`).

hredestig: yes exactly, that's what I meant, to cover the situation when you only give a string there should be a sensible behaviour. If you give a function it makes sense to have to set the attribute.

hredestig:
Nice improvement, if you agree with my changes to the doc I'll merge this?

pstjohn:
Good catch! I did forget to update the docs. This looks great, thanks for the addition.

PR #343, opened 2017-01-12
Deprecate ArrayBasedModel

### LINKED ISSUE #330: Deprecate ArrayBasedModel
This was brought up on the gitter and I just wanted to make sure it gets its own issue.

I'm in favor of dropping this class in favor of a `model.S` property to generate a stoichiometric matrix. The proposed changes of 0.6 would probably break a lot of the ArrayBasedModel's functionality, and I'm not sure the speed improvements offered are worth the development effort.

Definitely chime in if you think its worth keeping? The plan I'm guessing would be to raise some deprecation warnings in a future version (0.6?) and remove it some time later. 

Would finally close #160

pstjohn:
Deprecates ArrayBasedModel in favor of separate functions in
cobra.util.array. Still will likely need to implement similar functions
to get bounds arrays for cobra.io.mat

Would close #330

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -8,11 +8,6 @@ from .core import Object, Metabolite, Gene, Reaction, Model, \
     DictList, Species
 from . import io, flux_analysis, design
 
-try:
-    from .core import ArrayBasedModel
-except ImportError:
-    None
-
 __version__ = get_version()
 del get_version
 
--- a/cobra/core/ArrayBasedModel.py
+++ b/cobra/core/ArrayBasedModel.py
@@ -27,6 +27,10 @@ class ArrayBasedModel(Model):
             Specifies which type of backend matrix to use for S.
 
         """
+        warn("ArrayBasedModel is deprecated, use "
+             "`cobra.util.array.create_stoichiometric_matrix` instead",
+             DeprecationWarning)
+
         if deepcopy_model and isinstance(description, Model):
             description = description.copy()
         Model.__init__(self, description)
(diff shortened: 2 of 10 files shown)


cdiener:
As discussed this should also add a property `Model.S` which calls the helper function.

pstjohn:
I was wondering about that.. I suggested it earlier but am now less sure.

How does everyone feel about a model.S property? My reasons for hesitating now are:

* The function implementation gives you a few options on how the array is constructed (sparse, dense, dtype) that would get missed by model.S
* The function makes it explicitly clear that the array is being constructed on-the-fly and users won't think to try to write to model.S the way you could with `ArrayBasedModel`.
* Less clutter in `cobra.Model` for a feature that may only get used on occasion.

cdiener:
I would agree with point 1. Point 2 can be addresses easily by making S a read-only property. Since there is already a function calculating it it would not create too much clutter (maybe 5 more lines + docs).

zakandrewking:
Hey all- I'm not sure if this is the place to ask, but we still use optGpSampler, and it relies on the ArrayBasedModel. Has anyone thought about the future approach to MCMC sampling?

http://cs.ru.nl/~wmegchel/optGpSampler/

cdiener:
Cobrapy now has its own sampling module that implements optgp and achr.
Granted it's a bit slower than the c++ implementation (which I think is
abandoned...) but it has decent convergence.

pstjohn:
Its only deprecated for now, things will only start breaking when we remove the old code.. (0.7?)

Also, it looks from a quick browse that optGPsampler only wants the stochiometric matrix and vectors for upper/lower bounds and objectives, which would be pretty easy to get from the newer functions..

If the package gets an update to keep it current (last update seems to be in 2014), it shouldn't be hard to switch over

zakandrewking:
Thanks! We'll just adapt it to use the new functions.

PR #349, opened 2017-01-17
Cleaning up reaction.boundary

pstjohn:
This simplifies the implementation of `reaction.boundary`, and removes the (imo) duplicated / undocumented `model.exchanges`. 

I would suggest users in the future to get exchange reactions with 
```python
model.reactions.query(lambda x: x.boundary)
```

Thoughts?


pstjohn:
Probably pretty low: From the blame chart, it looks like `model.exchanges` got added as part of the cameo merge.

pstjohn:
Yikes, had an extra commit in there I didn't want for this PR (removing old media code).

Issue #303, opened 2016-10-26
design / implement better management of growth media


mmundy42:
I'd like to start a discussion on how to implement management of growth media. I can think of two approaches:
1. Add a new method to cobra.Model that takes either a dictionary or a json file with the exchange reaction IDs and lower bound for the reaction. The method updates the exchange reactions in the model as specified in the input.
2. Add a new cobra.Media object with methods to add and remove metabolites and set bound for metabolite. Add a new method to cobra.Model that takes a cobra.Media object that updates or adds exchange reactions in the model.

In either approach exchange reactions that are not in the input would have the bounds set to 0.

I also think there should be a way to save media to a file so they can be shared with others.

Thoughts? Other ideas?

cdiener:
Uh nice. I have two things close to my heart ;)

1. Most experimentalists and even theoretician don't think about growth media in terms of reactions but rather in terms of metabolites. The formulation into reactions is convenient for us but I think we could go a step further and allow defining media in terms of metabolites together with bound on the import of that metabolite. This would than be mapped to the respective boundary reactions automatically. Problems could arise with coupled transporters. We could ignore those imports for now. 

2. An analysis I have seen in a publication was to calculate the minimal medium for a model. Basically there are two methods, both depending on a minimum growth threshold: use pFBA with a weighted (molecular weight?) sum of import fluxes and keep the ones larger than zero or define a MILP with an indicator for each import and minimize the sum of indicator to get the medium with the fewest components.

pstjohn:
One of the ways I've been thinking about implementing this was overwriting Model.media to be a property where we could set the getters and setters.

In the get method, I would imagine calling a function we write in cobra.manipulation which looks for boundary reactions that are not knocked out. We'll have to make some decisions on directionality... for instance, if a model has a reaction ` CO2 --> ` with bounds (0, 1000), is that part of the media? If it is, we'll have big dictionaries to specify simple media, if its not, we'll have a tough time distinguishing between media where certain exchanges are `knocked out`.

In the set method, I'd imagine just the reverse of the getter. Iterate over the dict-like object, maybe something like 
```python
{'EX_glc_e': (0, 1000),
 'EX_h_e': (-1000, 1000),
}
```
and then set the bounds for each reaction as appropriate. In terms of _storing_ media along with the model, I would think the best place to do this would be in the model.notes dictionary. Just have a key for `media`, give each media dict a name, and then we could have `model.media`'s setter also accept string keys which would pull from the notes. so,
[...]

PR #350, opened 2017-01-18
Implement `Model.media`

pstjohn:
A first step towards closing #303, this PR implements a model.media property which returns a dictionary of bounds for the exchange reactions:
```python
>>> model.media
{'EX_ac_e': (0.0, 1000.0),
 'EX_acald_e': (0.0, 1000.0),
 'EX_akg_e': (0.0, 1000.0),
 'EX_co2_e': (-1000.0, 1000.0),
 'EX_etoh_e': (0.0, 1000.0),
 'EX_for_e': (0.0, 1000.0),
 'EX_fru_e': (0.0, 1000.0),
...
```

The setter method for model.media will check to see if the given media is a key present in `model.media_compositions`, and if so, will set it accordingly. Otherwise will iterate through the media, setting the bounds when they differ from the existing properties.

I've also removed the untested essentiality and initialize_growth_media functions, since these seemed to use an older media implementation.

Not sure how we'll want to handle
* if a reaction appears in the media but not in the model, or 
* a boundary reaction in the model is omitted from the media


mmundy42:
One initial comment - Ali Ebrahim commented in issue #303 that boundary metabolites don't make sense in a flux based model and shouldn't be used. And there are existing models where a reaction meets the definition of a boundary reaction but is not an exchange reaction. So I'm not sure if the media property can return the boundary reactions. Maybe it would make sense for a media to always be named and stored in media_compositions.

pstjohn:
> And there are existing models where a reaction meets the definition of a boundary reaction but is not an exchange reaction.

Do you have a good example for this? Media would flag any reaction without reactants or products, which would seem to catch all source and sink reactions. My bigger fear would be 'coupled import' mechanisms without an explicit external compartment. I.e., `glc + atp --> adp`, which we can't capture with `Reaction.boundary`.

On the boundary metabolites, I'm not really proposing we implement them. This is just a shortcut way of managing the bounds on 'exchange' reactions.

pstjohn:
If `media_compositions` is hidden, we'll need a way for users to save media. As it stands now, we're asking them to write directly to
```python
model.media_compositions['glucose'] = model.media
```
Constraining imports/exports, that might be tricky. I'm specifically thinking about changing something like bicarbonate concentration, where maybe you now want to regulate the upper bound of CO2 production

cdiener:
Ah sorry, misunderstood your API. You want something like `model.media = "glucose"` right? I thought that `model.media` would be another dict of media. Maybe rather rename to `model.medium` (singular) in that case?

mmundy42:
> Do you have a good example for this?

Yes, in the models provided with the "Anoxic conditions promote species-specific mutualism between gut microbes in silico" paper, several of the models have demand and sink reactions where Reaction.boundary() returns that is a boundary reaction because of the reaction's definition. For example,

```
bt.reactions.DM_4HBA.reaction
'4hba[c] --> '
bt.reactions.sink_chols.reaction
'chols[c] <=> '
```

It doesn't make sense to me that a reaction with metabolites in the cytosol compartment would be in the list of reactions for the media.

I agree with @cdiener that it should be an error if there is a reaction in the media that is not in the model.

I'm fine with reactions not in the media having their import set to 0. But it would be nice for there to be an option to get a list of those reactions. Just thinking it would give users a way to confirm that the media was applied in the way that they expected.

pstjohn:
So I suppose the first point is whether or not we want to mess around with compartments... Would it be problem if `model.medium` returned, for the reactions you mentioned:
```python
>>> bt.medium
{...
 'DM_4HBA': (0, 1000),
 'sink_chols': (-1000, 1000),
 ...
```

I guess this is also linked to the second point, which is what should `medium.setter` do when the dictionary is missing a reaction key. I would probably lean towards doing nothing to that reaction, so that as in the case above, a media dict missing `'DM_4HBA'` wouldn't knock out the sink reaction...

Mike, how would you imagine the API working? Did you use the previous media methods that explicitly asked for a 'boundary_compartment' kwarg?

mmundy42:
Talking with some users here to get their thoughts, will post more later.

mmundy42:
No, we did not use the previous media methods.

Talking to biologist users, it doesn’t quite make sense to them to have the media compositions be a part of the model.  They think of the media conditions as something separate from the model of the organism and that you apply a media condition to a model.

One use case is that you have a media defined (the dictionary format in this pull request is good) and then apply the media condition to a set of models. In our case, the set of models is for organisms living in the human gut. The media is a superset of all of the exchange reactions in the models. If there is an reaction in the media that is not in a model, that should not be an error. I know I just contradicted myself :-)

For the case where there is an exchange reaction in the model but not in media, we have been setting the import to 0.

A few other concerns:

1. Models could become large if each one has the media conditions included inside it. Probably not a big deal for studying single organisms but becomes a concern when studying a community across several media conditions.
2. Are the media conditions lost when saving the model to SBML or matlab formats? Assuming that saving to json would work.

I’m getting hung up on attempting to infer that a reaction is a boundary reaction by the definition. It seems like that should be an explicit attribute of the reaction but maybe that makes things too complicated.

For the API, how about a setter that simply takes a dictionary of reactions and possibly a name for the media.

pstjohn:
So you make some good points on not storing the media with the model. I'm happy to get rid of that functionality, I mainly included it to match the older media methods. If it doesn't seem like anyone was using them, and there's no real great place to document it, let's just remove that.

On boundary reactions, the determination in `Reaction.boundary` has been around for a while. I hesitate to make it a settable property, largely because I'll probably forget to set it 😄 . As long as users adhere to the 'boundary reaction' philosophy ('met_e --> '), I think they'll be OK.


So I think we're all on board for the current API setter, I think we just need to hammer out what happen in the following cases, `model.medium = new_medium`:
1. A reaction_id is present in the `new_medium` but *not* present in the model
2. A boundary reaction is present in the model but *not* present in the medium

I think (1) should at least raise a warning.. If it was a non-error warning, you could certainly still use the method in the manner you described above. But you wouldn't get caught off-guard applying a media to the wrong model.

I think (2) should probably leave the missing reaction alone. Just to protect against the case where a reaction (i.e., sink) is flagged as a boundary reaction when you weren't expecting it. It could certainly also raise a warning though.

cdiener:
If media are not stored within the model would they become their own object wrapping a dict? Could be nice to have a custom print method etc. 

I think the original argument for having media within the model is due to naming conventions. Metabolite names between different models can be different and so can reaction names. So a medium in cobrapy will always have to be specific for a given model.

pstjohn:
We could certainly have media inherit from dict, what kind of print method were you envisioning?

There were a few __repr_html__ methods from cameo we pulled out and could re-consider for cobrapy, maybe we do that in a separate PR?

cdiener:
I was thinking about having name and description for a media as well and having a print method that shows those and maybe some summary (how many imports/exports) for instance

pstjohn:
Lets see if this passes CI.
One concern for getting too fancy with the Media object is that there's no current way to save the media, either alone or as part of a model.

I'd be happy with the PR as-is for now at least, under the assumption that we could duck-type `medium` later to be a more fully-fledged object.

cdiener:
I agree. I think the `medium_from_metabolites` has higher priority anyways since it enables creating media from metabolite specifications like "real" growth media.

cobra/core/Model.py
```diff
@@ -118,6 +117,45 @@ def description(self, value):
...
+        Parameters
+        ----------
+        medium: dictionary-like
+            The medium to initialize. medium should be a dictionary defining
+            `{rxn_id: bounds}` pairs.
+
+        """
+
+        # Check the symmetric difference to see if reactions are missing
+        boundary_rxn_keys = set([rxn.id for rxn in self.reactions.query(
```
hredestig: use `self.medium.keys()` ?

cobra/core/Model.py
```diff
@@ -118,6 +117,45 @@ def description(self, value):
         self.name = value
         warn("description deprecated", DeprecationWarning)
 
+    @property
+    def medium(self):
+        return {rxn.id: rxn.bounds for rxn in
```
hredestig: how will this be with all export reactions, I didn't quite get the conclusion from the discussion, I note that  e.g. oleate export from yeast will show in up in the medium which perhaps isn't so intuitive?

pstjohn: I don't really see a way around this, short of requiring that users manually specify the reactions they're interested in.

One option could be to only look for boundary reactions with reaction.bounds that allow metabolites to be created? but then we loose the ability to set, for instance, the max. CO2 production rate. 

My approach was to return the bounds for all of the reactions on the boundary, which will necessarily include export reactions, and then let the user handle them appropriately.

cobra/core/Model.py
pstjohn: that works

cobra/core/Model.py
```diff
@@ -118,6 +117,45 @@ def description(self, value):
...
+
+        # Check the symmetric difference to see if reactions are missing
+        boundary_rxn_keys = set([rxn.id for rxn in self.reactions.query(
+                        lambda x: x.boundary)])
+        medium_keys = set(medium.keys())
+        missing_in_medium = boundary_rxn_keys - medium_keys
+        missing_in_model = medium_keys - boundary_rxn_keys
+
+        for rxnid in missing_in_medium:
+            warn("{} is missing from the medium".format(rxnid))
```
pstjohn: So there's really two issues here it seems,

1) I don't think we should be zero'ing out reaction bounds which arn't explicitly included in the passed medium. Largely as a result of my point above, which is that if model.medium returns export reactions a user isn't expecting, then it would zero those. So the warning on missing reactions was essentially because the desired behavior in this instance isn't obvious.

2) There's some back-and-forth on whether having extraneous reactions in the passed media should cause a warning or error. On one hand, it could just be a misspelling of the desired reaction (should let the user know), on the other, the media dict is being applied to many models, some of which have extra exchange reactions (see @mmundy42's above). Hence the warning here.

I would obviously prefer an approach which didn't use the warnings. But that would definitely require a firm consensus on what model.medium is for 😄 .

hredestig: So I think of it like this - when calling e.g. `model.set_medium()`, boundary reactions that are mentioned in that dict should be updated to allow import. If a flag is set boundary reactions that are not mentioned should be updated to be blocked as they are not available in the medium. I feel this would be consistent with thinking of medium as 'what you let your cells grow in' whereby not being able to set CO2 production perhaps isn't so intuitive use anyway. Warnings are then not necessary as the behaviour is explicit. Reactions not mentioned at all in the model but in the medium should raise a warning to cover the spell-error issue.

My first use-case in mind here is switching the C-source in my medium from glucose to galactose, in this case glucose import will have to be blocked and galactose enabled.

Would it be useful with a live discussion? I could organise another meeting.

mmundy42:
Lena and I think it would be useful to have an online meeting to finish working through this pull request.

hredestig:
Participate in the doodle if you would like to join a discussion on this http://doodle.com/poll/sv83htenvufvb4va

pstjohn:
I'm also happy to join a video chat.

On the glucose / galactose switching, I agree that's what I'm also largely using this for. But I'm not sure deleting the glucose key from the dictionary is really the easiest way to knock out glucose uptake... 

Remember, these media dictionaries are currently *all* the boundary reactions in model. Even for the textbook model, that's 20 reactions. For the ecoli test model, its a 330 key dictionary. So writing 329 entries to specify that glucose shouldn't be imported doesn't seem like a useful function 😄 .

```python
>>> import cobra.test
>>> model = cobra.test.create_test_model('textbook')
>>> model.medium
{'EX_ac_e': (0.0, 1000.0),
 'EX_acald_e': (0.0, 1000.0),
... (20 total entries)
 'EX_pyr_e': (0.0, 1000.0),
 'EX_succ_e': (0.0, 1000.0)}

>>> model = cobra.test.create_test_model('ecoli')

>>> model.medium
Out[6]:
{'DM_4CRSOL': (0.0, 1000.0),
 'DM_5DRIB': (0.0, 1000.0),
... (330 total entries)
 'EX_xylu__L_e': (0.0, 1000.0),
 'EX_zn2_e': (-1000.0, 1000.0)}
```

So the way I've been using these is doing something like this, to make glucose, glucose + xylose, and xylose only media dicts:

```python
glc_media = model.medium

glcxyl_media = dict(glc_media)
glcxyl_media['EX_xyl_e'] = (-10, 1000)

xyl_media = dict(glcxyl_media)
xyl_media['EX_glc_e'] = (0, 1000)
```


If we want to make missing keys knock out reactions, I think we'll need to make some more restrictive decisions on how we're going to classify boundary reactions. If we *only* wanted to consider import reactions, we could put in some helper scripts to return something like this for the textbook media:

```python
>>> textbook_model.medium
{'EX_co2_e': -1000,
 'EX_glc__D_E': -10,
 'EX_h20_e': -1000,
 'EX_h_e': -1000,
 'EX_nh4_e': -1000,
 'EX_o2_e': -1000,
 'EX_pi_e': -1000}
```

Where it only returns the active bound in the direction creating a metabolite, and only if that import reaction is active.

Then I could agree that deleting a key should knock out that import: it would set all the non-mentioned bounds in the direction of metabolite creation to 0. But we would loose the ability to specify excretion bounds with model.media. Does anyone really want that feature?

hredestig:
Alright so we have meeting Friday 6th at 16  (Copenhagen time) unless we would reach consensus here before that. I think perhaps where I first stumble with the approach currently in the pr is the wording, excretion being part of media. If it was named e.g. 'boundaries' instead then the behaviour would not be surprising. However, with 'media' I think your second proposal is pretty much exactly what I would expect. Maybe it would be warranted to have both implemented with 'media' just a wrapper to 'boundaries' but only concerning import reactions?

pstjohn:
That's an option, but I'm also wondering how useful looking at all the exchanges would be..

Theoretically, we could broaden this to the point where a user could just pass a dictionary of any reaction bounds they want updated. But that really goes against the current API of setting bounds directly through `model.reactions._my_reaction_key_.bounds`

@mmundy42, what would you think about the medium property as described above? (I.e., only looking at the bounds in direction of metabolite creation, and only for reactions which are active).

mmundy42:
@hredestig Looking at the Doodle, I think you want to meet on Monday, Feb 6.

> So I think of it like this - when calling e.g. model.set_medium(), boundary reactions that are mentioned in that dict should be updated to allow import. If a flag is set boundary reactions that are not mentioned should be updated to be blocked as they are not available in the medium. I feel this would be consistent with thinking of medium as 'what you let your cells grow in' whereby not being able to set CO2 production perhaps isn't so intuitive use anyway.

Lena and I agree with @hredestig here.

>  (I.e., only looking at the bounds in direction of metabolite creation, and only for reactions which are active)

And this makes sense to us too.

pstjohn:
Here's an implementation along the lines of what I mentioned above:

```python
>>> model.medium
{'EX_co2_e': 1000.0,
 'EX_glc__D_e': 10.0,
 'EX_h2o_e': 1000.0,
 'EX_h_e': 1000.0,
 'EX_nh4_e': 1000.0,
 'EX_o2_e': 1000.0,
 'EX_pi_e': 1000.0}
```

Each media key is the absolute value of the bound in the direction of metabolite creation. Deleting a key will cause the `medium.setter` to 0 out the bound associated with metabolite creation for that reaction.

pstjohn:
Just to put @cdiener's comments from gitter here, we talked about whether or not this should use metabolite keys vs. reaction keys for the medium dictionary.

It would certainly be trivial to let the user use either if we could assume that each exchange reactions had only one metabolite. If for some reason we can't, then it gets a lot more complicated. (In terms of changing the API down the road, I think the biggest question is should model.medium return reaction keys or metabolite keys.)

I'm happy with the current implementation for now.. does anyone else feel strongly one way or the other?

cdiener:
I think this (metabolite keys) should be a helper function rather than part of the setter. This way it could also get arguments on how to deal with certain edge cases.

pstjohn:
so I think we're RTM on this then

PR #351, opened 2017-01-18
pep8 fixes for sbml.py

pstjohn:
Quick fix, this should let us drop the exclusion for pep8 checks of sbml.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ install:
   - build_wheel . $PLAT
 
 script:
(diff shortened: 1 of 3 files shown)


cdiener:
Sorry, didn't see it. Also made pep8 fixes in #352. Please remove those if this PR get pulled in before.

pstjohn:
Lol, oops, we were probably doing this at the same time.
We could merge yours and close this one, but you should also remove the pep8 exclusions from .travis and tox.ini..

cdiener:
Since you already I thinks it would be better to update mine so it does not include the pep8 fixes. Will do it now.

PR #361, opened 2017-01-30
Allow fixed random number seed in flux sampling

cdiener:
From E-mail request. 

This PR allows the random number seed to be specified explicitly thus allowing for reproducible sampling code. For ARCH it will use the given seed as is. OptGP will give each process a seed of `seed + idx` where `idx` is the index of the process ranging from `0 .. n_processes-1`.

### DIFF
--- a/cobra/flux_analysis/__init__.py
+++ b/cobra/flux_analysis/__init__.py
@@ -3,11 +3,6 @@ try:
 except:
     numpy = None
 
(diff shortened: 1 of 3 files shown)


cdiener:
No Python 3.6 wheels for swiglpk yet (https://github.com/biosustain/swiglpk/pull/20).

PR #362, opened 2017-01-30
Revamp the website

cdiener:
This adds the new website to the `gh-pages` branch. It might still not be perfect, but I think it's already a bit more informative than the old (beloved) website.

[Preview](https://cdiener.github.io/cobrapy_website)

This is currently a manually generated version. As soon as we figure out where to keep the sources we can discuss deployment options.


pstjohn:
looks great!

mmundy42:
Nice job Christian!

PR #367, opened 2017-02-02
Port single deletions to optlang

cdiener:
Sorry that this took awhile, there were a few hickups to sort out. Particularly there were quite some bugs in our initial optlang code and also small issues with optlang. Due to this there are also several changes in the Model code as well as tests. Also waiting for a new optlang release since optlang/devel includes a fix that helps our `Solution` class.

Single deletions are now fully working with optlang (FBA and MOMA). The FBA deletions are more or less twice as slow with optlang (see appended text file) whereas MOMA is as fast as before or even a little faster.

The new interfaces are much nicher in terms of customizability. For instance triple+ deletions with MOMA were impossible with the old interface. Now you can just use:

```Python
from cobra.flux_analysis.moma import moma_model

with model as m:
     add_moma(m)     # adds constraints and new objective
     m.reactions.r1.bounds = (0, 0)
     m.reactions.r2.bounds = (0, 0)
     m.reactions.r3.bounds = (0, 0)
     m.optimize(objective_sense=None)
     print("New growth rate: ", m.solution.f)
```

Test output
--------------- 
[test_output.txt](https://github.com/opencobra/cobrapy/files/749034/test_output.txt)

Some performance tips
-------------------------------

Currently creating the solution actually takes much longer than solving the model (since optlang.Model.primal_values is not that fast), thus it's a good strategy to avoid calling model.optimize() if you only need a few fluxes. Rather use:

```
model.solver.optimize()
model.solver.variables.my_reaction.primal
```

This is much faster.

### DIFF
--- a/cobra/flux_analysis/moma.py
+++ b/cobra/flux_analysis/moma.py
@@ -1,6 +1,61 @@
 from scipy.sparse import dok_matrix
 
 from ..solvers import get_solver_name, solver_dict
+from cobra.util.solver import add_to_solver
+import sympy
+
+
+def moma_model(model):
+    """Changes a model's objective and bounds to represent a MOMA
+    (minimization of metabolic adjustment) model.
+
+    Parameters:
+    -----------
+
+    model : a cobra model
+
+    Returns:
+    --------
+    Nothing.
+
+    Notes
+    -----
+
+    In the original MOMA specification one looks for the flux distribution
+    of the deletion (v^d) closest to the fluxes without the deletion (v).
+    In math this means:
+
+    minimize \sum_i (v^d_i - v_i)^2
+    s.t. Sv^d = 0
+         lb_i <= v^d_i <= ub_i
+
+    Here, we use a variable transformation v^t := v^d_i - v_i. Substituting
+    and using the fact that Sv = 0 gives:
+
+    minimize \sum (v^t_i)^2
+    s.t. Sv^d = 0
+         v^t = v^d_i - v_i
+         lb_i <= v^d_i <= ub_i
+
+    So basically we just re-center the flux space at the old solution and than
+    find the flux distribution closest to the new zero (center).
+    """
+    model.optimize()
+    prob = model.solver.interface
+    v = prob.Variable("moma_old_objective")
+    c = prob.Constraint(model.solver.objective.expression - v,
+                        lb=0.0, ub=0.0, name="moma_old_objective_constraint")
+    to_add = [v, c]
+    new_obj = sympy.Float(0.0)
+    for r in model.reactions:
+        flux = model.solution.fluxes[r.id]
+        dist = prob.Variable("moma_dist_" + r.id)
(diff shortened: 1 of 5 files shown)


phantomas1234:
Two comments,

1. I was hoping that we could stay somewhat close to the `cameo.flux_analysis` implementations of the simulation functions in order to minimize the amount of refactoring we'll have to do in cameo once cobrapy has been fully refactored. So, I guess my original plan was to simply move `cameo.flux_analysis.simulation` to `cobra.flux_analysis.simulation` (refactoring the code to use the new HistoryManage etc.) and then change all dependent downstream functions like `single_gene_deletions` to use the new simulation function. I guess the problem with that is that we're trying to maintain backwards compatibility with the old solvers. I guess I am favoring more and more to drop backwards compatibility for the sake of getting the migration over with. I think a lot of the performance issues would also disappear. Of course, we could swap in better problem formulations (like your new MOMA formulation).

2. Regarding the performance issue with creating the solution, can't we start using `cameo.core.solution.LazySolution` to avoid these problems? https://github.com/biosustain/cameo/blob/devel/cameo/core/solution.py#L186

cdiener:
Hmm, those tests should not be run. Will push a fix...

cdiener:
Just to comment on LazySolutions. I think the problem we had is it's aggresive erroring. In many cases we just want to get the solution change the model and access the old solution. So making it the default would break a lot of things. For that particular case it would be a good solution though, I agree.

cdiener:
Coverage decreases because there is now more code in blocks that require QP solvers and are thus not tested on CI...

cobra/test/test_solver_model.py
```diff
@@ -700,30 +704,39 @@ def test_invalid_objective_raises(self, model):
...
     def test_solver_change(self, model):
         solver_id = id(model.solver)
         problem_id = id(model.solver.problem)
         solution = model.optimize(solution_type=LazySolution).x_dict
-        model.solver = 'glpk'
+        model.solver = "cplex"
         assert id(model.solver) != solver_id
         assert id(model.solver.problem) != problem_id
-        new_solution = model.optimize(solution_type=LazySolution)
+        new_solution = model.optimize(solution_type=LazySolution).x_dict
```
hredestig: `x_dict` renamed to `fluxes`

cobra/flux_analysis/moma.py
```diff
@@ -1,6 +1,70 @@
 from scipy.sparse import dok_matrix
 
 from ..solvers import get_solver_name, solver_dict
+import cobra.util.solver as sutil
+import sympy
+
+
+def moma_model(model):
```
hredestig: I like the proposal `add_moma`

cdiener: Ok, implemented. Also removed a function deprecated in 0.4 and converted the docstrings to numpy format. Have you seen the changes in `test_flux_analysis.py`? I substituted some benchmarks with smaller functions since the whole benchmarking tok 30 min + with cplex and even longer if you also check with scipy (which is sloooow).

PR #368, opened 2017-02-03
fix: context manager not properly resetting changes to model.objective

phantomas1234:
https://gitter.im/opencobra/cobrapy?at=5893ae2ad99b1dd03d3247b2

### DIFF
--- a/cobra/core/Reaction.py
+++ b/cobra/core/Reaction.py
@@ -170,9 +170,8 @@ class Reaction(Object):
     @objective_coefficient.setter
     def objective_coefficient(self, value):
         if self.model is not None:
-            coef_difference = value - self.objective_coefficient
-            self.model.solver.objective += \
-                coef_difference * self.flux_expression
+            if self.flux_expression is not None:
+                self.model.solver.objective.set_linear_coefficients({self._forward_variable: value, self._reverse_variable: -value})
         self._objective_coefficient = value
 
     def __copy__(self):
--- a/cobra/core/Reaction.py
+++ b/cobra/core/Reaction.py
@@ -171,7 +171,9 @@ class Reaction(Object):
     def objective_coefficient(self, value):
         if self.model is not None:
             if self.flux_expression is not None:
-                self.model.solver.objective.set_linear_coefficients({self._forward_variable: value, self._reverse_variable: -value})
+                self.model.solver.objective.set_linear_coefficients(
+                    {self._forward_variable: value,
+                     self._reverse_variable: -value})
         self._objective_coefficient = value
 
     def __copy__(self):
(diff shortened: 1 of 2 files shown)


cdiener:
Hi, could you explain why that works, since the bug was really cryptic.

pstjohn:
Also, why is `@resettable` on model.objective? It would seem that's the highest part of the call stack.

Could we follow something like @cdiener's `add_to_solver()` wrapper, but `modify_solver_objective()`? Then we could check contexts there.

cdiener:
Yeah that's true Reaction.objective_coefficient and model.objective_coefficient do need a reset as well. The problem with just adding `resettable` to Reaction.objetive_coefficient is that it won't work if forward_coef != - backward_coef since it returns 0 in that case and resetting to zero leaves you with both coefs as zero. Currently, `Reaction/Model.objective_coefficient` is tightly integrated with the objective and they are updated automatically upon changes on the low level objective so I think a better solution would be a new decorator `reset_objective` which simply resets the objective accordingly. Than we just decorate the affected 3 functions and done.

pstjohn:
Yeah I've tried a few edits myself but this is definitely a mess 😄 . For instance, calling
`model.objective = model.reactions.ATPM` seems to set `model.solver.objective` twice:
once in `Reaction.objective_coefficient` and again directly, see lines 595:596 in Model.py.

pstjohn:
Hope you don't mind the edits.
Still doesn't seem to be working, this way we can check on CI when we fix the issue.

cdiener:
The gitter bug is fixed. What's failing are some IO tests now...

hredestig:
had a look at this and turns our that IO problem is rather tricky to get right as it boils down to the split-head between what actually decides the objective, `model.solver.objective` or `reaction.objective_coefficient` with something during (un)pickling that doesn't get set correctly. I think the solution is to drop `reaction._objective_coefficient` completely (working on it..)

pstjohn:
Maybe we could roll this in with a larger change to address #371 and the like? I.e., make sure that cobrapy methods are all just thin wrappers around the underlying optlang object?

I just think that would be easier to keep everything consistent.

cdiener:
I agree with @pstjohn. There are several small things that should get adapted. The one where `Reaction.flux` should gets its value from is a complicated one. For instance its possible that LazySolution deems a solution "fresh" but you still get an error accessing the primal value from optlang, since some solvers like cplex have their onw invalidation methods.

pstjohn:
I think its OK if Reaction.flux raises an error.. the old Reaction.x raised errors.
But it would likely allow us to standardize the errors we expect from each method.

hredestig:
fixing this in #377 instead

PR #373, opened 2017-02-09
fix: avoid immutables in defaults

hredestig:
suppose not intentional and can cause odd phenomena. correct numpy
docstring syntax


cdiener:
Hi, I think that `what` shoud not have any default actually. It does nothing sensible when called without `what` set. That was more a relict of some edits where I overlooked deleting the default for what. The rest looks good to me.

PR #374, opened 2017-02-09
Better resets for objectives

cdiener:
This adds a working and much faster reset for objective manipulations. Basically it will reset the objective using `model.solver.objective` if it detects that the modification came from `Model` also correctly resetting former quadratic objectives (which was not the case before). When it detects that the change cam from `Reaction` it will rather use `set_linear_coefficients` to reset only coefs for that reaction.

This speeds up leaving the context by quite a bit. For example in a naive implementation of FVA I saw a difference of 3 orders of magnitude in performance for the "textbook" model.

### DIFF
--- a/cobra/core/Model.py
+++ b/cobra/core/Model.py
@@ -17,7 +17,7 @@ from sympy.core.singleton import S
 from cobra.util.util import AutoVivification
 from cobra.util.context import HistoryManager, resettable
 from cobra.util.solver import solvers, SolverNotFound, interface_to_str,\
-                              get_solver_name
+                              get_solver_name, reset_objective
 import optlang
 
 
@@ -650,7 +650,7 @@ class Model(Object):
         return self.objective_reactions
 
     @objective.setter
-    @resettable
+    @reset_objective
     def objective(self, value):
         if isinstance(value, six.string_types):
             try:
--- a/cobra/core/Reaction.py
+++ b/cobra/core/Reaction.py
@@ -13,6 +13,7 @@ import hashlib
 
 from cobra.util.util import Frozendict, _is_positive
 from cobra.util.context import resettable
+from cobra.util.solver import reset_objective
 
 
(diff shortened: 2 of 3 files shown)


pstjohn:
Why can't we do something similar to `add_to_solver`: where the inner objective setter in each of those different methods puts in a call to `set_objective(model, new_objective)`? Then you just context manage within the setter and you're good to go?

cdiener:
Yeah sure, works as well. You just have to handle a lot of different input types (optlang.Objective, sympy expression, dict).

pstjohn:
Well my thought is that the `set_objective` would only things that would go directly into optlang.Objectives: the code to convert from the dicts, sympy expressions would stay in the various objective setting functions. (I.e., objective reactions would call reaction.objective_coefficient, which would setup an optlang objective to pass to `set_objective`.)

My thought is that we want to keep the context management as low-level as possible, so that when we implement new features we don't have to worry about context management...

Sortof like model.medium: there's no explicit context management there, it all gets handled by reaction.lower_bound and reaction.upper_bound.

cdiener:
The thing is that optlang supports two ways to set the objective:

1. Using a sympy expression and `Objective`
2. Setting linear coefficients directly via `set_linear_coefficients`

2 is about 100x faster since it avoids sympy alltogether. So where possible you want to use 2 and not 1 since setting objectives is something you loop a lot in many of the flux analysis methods (for instance FVA using 1 takes 2 secs for the textbook model and 20 ms with 2). We could write a function that takes either a dict or an `Objective` or two functions, one for each. However, it's important that `Model.objective_coefficients` and `Reaction.objective_coefficient` use strategy 2 for setting and resetting.

pstjohn:
ahh ok, cool, thanks for explaining that. Then I'm on board with either strategy.
I have a bunch of questions about the implication of that though, I can ask them during the next meeting :)

PR #377, opened 2017-02-10
fix: refactor objective_coefficient

hredestig:
make reaction.objective_coefficient a wrapper for
model.solver.objective. Contains the breaking change that
reaction.objective_coefficient can only be set on reactions that
actually have a model and objective_coefficient cannot be set when
creating a new reaction.


cobra/core/Model.py
```diff
@@ -720,3 +654,39 @@ def __exit__(self, type, value, traceback):
         """Pop the top context manager and trigger the undo functions"""
         context = self._contexts.pop()
         context.reset()
+
+
+def get_reactions(model, reactions):
```
cdiener: Awesome. I wrote sooo much code where I have to convert reaction ids to the reactions :+1:

cobra/core/Reaction.py
```diff
@@ -154,26 +158,24 @@ def _reset_var_cache(self):
...
     def objective_coefficient(self, value):
-        if self.model is not None:
-            coef_difference = value - self.objective_coefficient
-            self.model.solver.objective += \
-                coef_difference * self.flux_expression
-        self._objective_coefficient = value
+        if self.model is None:
+            raise AttributeError('cannot assign objective to a missing model')
+        if self.flux_expression is not None:
+            self.model.solver.objective.set_linear_coefficients(
```
cdiener: That should use `set_objective_*` from the solver utils so we don't have to worry about the context manager.

cobra/io/sbml.py
```diff
@@ -108,7 +108,7 @@ def create_cobra_model_from_sbml_file(sbml_filename, old_sbml=False,
             "convert fbc to cobra", True, "Convert FBC model to Cobra model")
         result = model_doc.convert(conversion_properties)
         if result != LIBSBML_OPERATION_SUCCESS:
-            raise(Exception("Conversion of SBML+fbc to COBRA failed"))
+            raise (Exception("Conversion of SBML+fbc to COBRA failed"))
```
cdiener: What do the brackets around the Exception do? Doen't `raise Exception(...)` work?

cobra/core/Model.py
pstjohn: So, I agree this is good functionality, but would it be better suited for `dictlist`?
Something like `get_by_id`, but would accept an iterable, would check for an `id` attribute and search on that if appropriate, etc.? Then we wouldn't need a `get_metabolites` down the road.

Just a consideration, I'm fine with this as is

cdiener: I have no preference where that will live in the end :smile:

cobra/io/sbml.py
hredestig: my editor added the space, the parenthesis was just old lint

cobra/core/Reaction.py
hredestig: good point, cheers

cobra/core/Model.py
hredestig: very good point, added to dictlist

cobra/util/solver.py
```diff
@@ -26,6 +27,63 @@ class SolverNotFound(Exception):
...
+    for rxn in reactions:
+        forward_coefficient = coefficients.get(rxn.forward_variable, 0)
+        reverse_coefficient = coefficients.get(rxn.reverse_variable, 0)
+        if forward_coefficient != 0:
+            if forward_coefficient == -reverse_coefficient:
+                linear_coefficients[rxn] = float(forward_coefficient)
+    return linear_coefficients
+
+
+def set_objective_from_coefficients(model, coefficients, additive=False):
```
hredestig: Have implemented this now, not super pretty but seems to work

cobra/util/solver.py
```diff
@@ -26,6 +29,108 @@ class SolverNotFound(Exception):
...
+    additive: bool
+        If true, add the terms to the current objective, otherwise start with
+        an empty objective
+    """
+    by_expression = (isinstance(value, sympy.Basic) or
+                     isinstance(value, model.solver.interface.Objective))
+    by_coefficient = isinstance(value, (dict, int, list, six.string_types,
+                                        cobra.Reaction))
+    not_supported = (
+        additive and (not model.objective.is_Linear and not by_expression))
```
cdiener: It kind of works since optlang keeps linear and quadratic coefficients apart. So setting linear coefficients in a mixed objective will just add those to the entire objective.

For instance, the following works:

```Python
In [28]: mod = create_test_model("textbook")

In [29]: mod.solver = "cplex"

In [30]: mod.objective = mod.solver.variables.ATPM**2

In [31]: mod.solver.objective.expression
Out[31]: 1.0*ATPM**2

In [32]: mod.solver.objective.set_linear_coefficients({mod.solver.variables.ATPM: 2})

In [33]: mod.solver.objective.expression
Out[33]: 1.0*ATPM**2 + 2.0*ATPM
```

cobra/util/solver.py
```diff
@@ -26,6 +29,108 @@ class SolverNotFound(Exception):
...
+
+
+def set_objective(model, value, additive=False):
+    """ Set the model objective
+
+    Parameters
+    ----------
+    model: cobra model
+       The model to set the objective for
+    value: model.solver.interface.Objective such as
```
cdiener: Ah sorry, I just meant that the function would only accept `optlang.Objective` or a dict. All the conversion to those objects is than implemented in the Model and Reaction objects. At least this was the proposal of @pstjohn and I concurred :) But do as you see fit.

cobra/util/solver.py
```diff
@@ -49,9 +154,9 @@ def get_solver_name(mip=False, qp=False):
 
     Parameters
     ----------
-    mip: string
```
cdiener: Thanks for the fixes!

cobra/util/solver.py
pstjohn: Honestly I suppose either way works. Making the function accept only a few types makes it simpler to implement, but if it helps remove duplicating code in different Model and Reaction objects, it could make sense to do the conversion in the function.

cobra/util/solver.py
```diff
@@ -26,6 +29,108 @@ class SolverNotFound(Exception):
...
+            value[reaction] = reaction.objective_coefficient
+            model.solver.objective.set_linear_coefficients(
+                {reaction.forward_variable: coef,
+                 reaction.reverse_variable: -coef})
+    else:
+        raise TypeError(
+            '%r is not a valid objective for %r.' % (value, model.solver))
+    context = get_context(model)
+    if context:
+        context(partial(set_objective, model=model, value=value,
```
pstjohn: Will this work? I guess I'm confused by how `value` becomes the old value we want to reset to. It might be clearer if we could use a new variable (`old_value`?) to keep track of what should be undone with the context manager.

cobra/util/solver.py
hredestig: Thanks for the clarification, I went for the middle ground, set_objective takes Objective or sympy expression and model.objective sorts out the other object conversions. Seems okay?

cobra/util/solver.py
hredestig: I noted that but wasn't sure there won't be corner cases and misinterpretation so went for what I felt was the safer option reckoning that if you have a complex objective, you probably don't want to use reaction.objective_coefficient and if you did, it was probably a mistake. We can always adjust this later on when we have more test cases?

cobra/util/solver.py
hredestig: Good point, for clarity, I now use the `reverse_value` object to make it easier to read, works?

cobra/util/solver.py
pstjohn: Good catch @cdiener, that's definitely an important point.

pstjohn:
> How does that not create an infinite recursion when exiting the context manager? 

Technically we won't since `model.__exit__` pops the top-most context before calling the reset methods. But if there was a nested context, things might get unnecessarily verbose.

PR #387, opened 2017-02-14
Lazy solution [WIP]

### LINKED ISSUE #380: Improve LazySolution behavior/interface
* Let LazySolution record requested values and make those available even after being invalidated.
* Add `__getitem__` that works just like `get_primar_by_id`.

Midnighter:
fixes #380
Based no our discussion in #380, I made the following changes:

* one `Solution` class only (plus `LegacySolution`)
* deprecation warnings for most of the old interface

Problems at the moment:
1. Our tests use the old interface quite heavily. Luckily pytest swallows those warnings so it's not a big deal.
2. There's quite a bit of setting of values going on. The logic of which escapes me. In my own world, the solution instance should not be modifiable at all but completely depend on the model instance. If you change the model and optimize, you request a new solution.


pstjohn:
> There's quite a bit of setting of values going on. 
Do you have a example of that somewhere? I'm assuming it would be done to make sure the tests are repeatable across different solvers.. 

I'm certainly OK with making `Solution` read-only. Maybe we'll need a test solution (like the test model) that has a reproducible state for the unit tests.

Midnighter:
> Do you have a example of that somewhere? I'm assuming it would be done to make sure the tests are repeatable across different solvers..

I was exaggerating a little. There is a case in the tests but since that was the public interface it makes me wonder how many users would rely on setting a solution attribute?

pstjohn:
Gotcha, the place I expected to find that behavior was in my summary method tests, but I'm remembering now we used a pickled solution: https://github.com/opencobra/cobrapy/blob/devel-2/cobra/test/conftest.py#L42-L48

I wonder if the writing to solution objects is to allow Reaction.x to be stored with the json-serialized model? 

I do think there should be some way to store and load solution objects, whether its through pickling or json serialization. Not sure how hard that would be or if that's a question better suited for the optlang issue board.

cdiener:
> There's quite a bit of setting of values going on.

I think some instances where that happened were methods like pFBA which update the solution of model to the pFBA result.

cobra/core/solution.py
```diff
@@ -0,0 +1,359 @@
...
+from future.utils import python_2_unicode_compatible
+
+from cobra.exceptions import UndefinedSolution
+
+__all__ = ("Solution",)
+
+LOGGER = logging.getLogger(__name__)
+
+
+@python_2_unicode_compatible
```
hredestig: not used anywhere else, perhaps skip for now for consistency and then correct in the whole package if needed, reduce inconsistency?

Midnighter:
Unless we decide to have pandas in there now this is good to go.

Closes #380

hredestig:
pandas is a great idea but not needed to reach a first 0.6 release

PR #391, opened 2017-02-14
Loopless solutions for optlang [fix #385]

### LINKED ISSUE #385: refactor loopless model

cdiener:
Up to now only implements the original method from Schellenberger. First time the optlang interface really shines. I could basically implement the original method as is. This requires less constraints than the prior version in cobrapy and brings a speed up (twice as fast).

### Missing:

- [x] a posteriori version (should be much faster)
- [x] faster loopless FVA using the method from CycleFreeFlux (semi a posteriori) 

As expected *a posteriori* methods are faster. For FVA the loopless version is still slower than the non-loopless one for large problems (kind of expected). However it is much faster than using `add_loopless` or `construct_loopless_model` followed by calling "normal" FVA. Faster meaning 2 seconds vs 2 minutes for textbook and a much larger gap (~1000x) for ecoli.

### API

This one I am still not sure about. So fire away with your comments

A priori:
```Python
with model:
    add_loopless(model)
    model.optimize()
    single_gene_deletion(model)
```

A posteriori:
```Python
with model:
    model.optimize()
    fluxes = loopless_solution(model)
```

Maybe that one should take a flux distribution as argument alongside with the model? That way you could use it to convert flux samples or saved flux distributions to loopless ones as well...

### Benchmarks

[test_output.txt](https://github.com/opencobra/cobrapy/files/778088/test_output.txt)

### DIFF
--- a/cobra/flux_analysis/loopless.py
+++ b/cobra/flux_analysis/loopless.py
@@ -11,8 +11,9 @@ from cobra.manipulation.modify import convert_to_irreversible
 try:
     import numpy
     from cobra.util import nullspace
-except:
+except ImportError:
     numpy = None
+    nullspace = None
 
 
 def add_loopless(model, zero_cutoff=1e-12):
@@ -22,7 +23,7 @@ def add_loopless(model, zero_cutoff=1e-12):
 
     Parameters
     ----------
-    model : a cobra model
+    model : cobra.Model
         The model to which to add the constraints.
     zero_cutoff : positive float, optional
         Cutoff used for null space. Coefficients with an absolute value smaller
@@ -35,15 +36,16 @@ def add_loopless(model, zero_cutoff=1e-12):
     References
     ----------
     .. [1] Elimination of thermodynamically infeasible loops in steady-state
-           metabolic models.
-       Schellenberger J, Lewis NE, Palsson BO.
-       Biophys J. 2011 Feb 2;100(3):544-53.
-       doi: 10.1016/j.bpj.2010.12.3707.
-       Erratum in: Biophys J. 2011 Mar 2;100(5):1381.
+       metabolic models. Schellenberger J, Lewis NE, Palsson BO. Biophys J.
+       2011 Feb 2;100(3):544-53. doi: 10.1016/j.bpj.2010.12.3707. Erratum
+       in: Biophys J. 2011 Mar 2;100(5):1381.
     """
+    if not numpy:
+        raise ImportError("add_loopless requires numpy :(")
+
     internal = [i for i, r in enumerate(model.reactions) if not r.boundary]
-    Sint = model.S[:, numpy.array(internal)]
-    Nint = nullspace(Sint).T
+    s_int = model.S[:, numpy.array(internal)]
+    n_int = nullspace(s_int).T
     max_bound = max(max(abs(b) for b in r.bounds) for r in model.reactions)
     prob = model.solver.interface
 
@@ -67,13 +69,14 @@ def add_loopless(model, zero_cutoff=1e-12):
     add_to_solver(model, to_add)
 
     # Add nullspace constraints for G_i
-    for i, row in enumerate(Nint):
+    for i, row in enumerate(n_int):
         name = "nullspace_constraint_" + str(i)
         nullspace_constraint = prob.Constraint(S.Zero, lb=0, ub=0, name=name)
         add_to_solver(model, [nullspace_constraint])
         coefs = {model.solver.variables[
-            "delta_g_" + model.reactions[ridx].id]: row[i]
-            for i, ridx in enumerate(internal) if abs(row[i]) > zero_cutoff}
+                 "delta_g_" + model.reactions[ridx].id]: row[i]
+                 for i, ridx in enumerate(internal) if
+                 abs(row[i]) > zero_cutoff}
         model.solver.constraints[name].set_linear_coefficients(coefs)
 
 
@@ -84,11 +87,8 @@ def loopless_solution(model):
 
     Parameters
     ----------
-    model : a cobra model
+    model : cobra.Model
         The model to which to add the constraints.
-    obj_val : a float
-        Fix the objective value to this value. Uses the current optimum if
-        None.
 
     Returns
     -------
@@ -111,10 +111,9 @@ def loopless_solution(model):
     References
     ----------
     .. [1] CycleFreeFlux: efficient removal of thermodynamically infeasible
-           loops from flux distributions.
-       Desouki AA, Jarre F, Gelius-Dietrich G, Lercher MJ.
-       Bioinformatics. 2015 Jul 1;31(13):2159-65.
-       doi: 10.1093/bioinformatics/btv096.
+       loops from flux distributions. Desouki AA, Jarre F, Gelius-Dietrich
+       G, Lercher MJ. Bioinformatics. 2015 Jul 1;31(13):2159-65. doi:
+       10.1093/bioinformatics/btv096.
     """
     # Need to reoptimize otherwise spurious solution artifacts can cause
     # all kinds of havoc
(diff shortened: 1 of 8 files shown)


cobra/flux_analysis/loopless.py
```diff
@@ -1,6 +1,71 @@
...
+    ----------
+    .. [1] Elimination of thermodynamically infeasible loops in steady-state
+       metabolic models.
+       Schellenberger J, Lewis NE, Palsson BØ.
+       Biophys J. 2011 Feb 2;100(3):544-53.
+       doi: 10.1016/j.bpj.2010.12.3707.
+       Erratum in: Biophys J. 2011 Mar 2;100(5):1381.
+    """
+    internal = [i for i, r in enumerate(model.reactions) if not r.boundary]
+    Sint = model.S[:, np.array(internal)]
```
pstjohn: Do we think this is an operation we'll do many times? Just thinking that we could speed things up if I had the array-generating function accept an optional argument to only give you certain columns...

In this case though, probably not a speed-limiting factor.

cobra/flux_analysis/loopless.py
```diff
@@ -1,6 +1,71 @@
...
+
+    # Add nullspace constraints for G_i
+    for i, row in enumerate(Nint):
+        name = "nullspace_constraint_" + str(i)
+        nullspace_constraint = prob.Constraint(S.Zero, lb=0, ub=0, name=name)
+        add_to_solver(model, [nullspace_constraint])
+        coefs = {model.solver.variables[
+            "delta_g_" + model.reactions[ridx].id]: row[i]
+            for i, ridx in enumerate(internal) if abs(row[i]) > zero_cutoff}
+        model.solver.constraints[name].set_linear_coefficients(coefs)
```
pstjohn: is this something we'll want to context-manage eventually?

cobra/test/test_flux_analysis.py
```diff
@@ -291,22 +291,46 @@ def construct_ll_test_model(cls):
...
+
     def test_loopless_benchmark(self, benchmark):
         test_model = self.construct_ll_test_model()
-        benchmark(lambda: construct_loopless_model(test_model).optimize())
 
-    def test_loopless(self):
+        def _():
+            with test_model:
+                add_loopless(test_model)
+                test_model.optimize(solver="optlang-glpk")
```
pstjohn: so I'm confused: glpk can handle the MILP formulation?

cobra/flux_analysis/loopless.py
cdiener: For now it's only used here...

cobra/flux_analysis/loopless.py
cdiener: Already is due to the add_to_solver above. Had to do it like that because you can't set linear coefficients for a Constraint that is not part of an optlang.Model yet.

cobra/test/test_flux_analysis.py
cdiener: Yes! Absolutely. GLPK includes a MIP solver. It does not support indicator variables like cplex or gurobi so you have to declare them by hand.

cobra/flux_analysis/loopless.py
```diff
@@ -1,6 +1,71 @@
...
+
+    Returns
+    -------
+    Nothing
+
+    References
+    ----------
+    .. [1] Elimination of thermodynamically infeasible loops in steady-state
+       metabolic models.
+       Schellenberger J, Lewis NE, Palsson BO.
```
Midnighter: Could this not be fixed by using `u"""` rather than reverting to ASCII? Internationalization is a good thing in general, we're not in the 80s anymore :)

cdiener: I think in this case it's not that critical since Palsson is cited like that pretty often. If it would in actual code I would agree. But really have no preference. I can add the `u"""` if you would like.

Midnighter: You're right, he is cited with O fairly often. Let's leave it then.

cobra/flux_analysis/variability.py
```diff
@@ -143,7 +158,7 @@ def _fva_optlang(model, reaction_list, fraction):
         fva_old_objective = prob.Variable(
```
hredestig: do you need `m.optimize()` above instead to save the solution? also, I reverted `m.optimize()` to raise an exception when doing the pfba, was not intended as a final solution there but work-in-progress, with that change I guess you don't need to raise `ValueError` here

cobra/flux_analysis/loopless.py
```diff
@@ -1,6 +1,232 @@
...
+except ImportError:
+    numpy = None
+    nullspace = None
+
+
+def add_loopless(model, zero_cutoff=1e-12):
+    """Add variables and constraints to make a model thermodynamically
+    feasible. This removes flux loops. The used formulation is described
+    in [1]_.
+
```
hredestig: perhaps could give an example of when to use `add_loopless` versus `loopless_solution` to help understand the reason for both of them?

pstjohn:
So here's a random question:

what if I did:

```python
add_loopless(model)
model.add_reaction(my_new_reaction)
model.optimize()
```

or even
```python
add_moma(model)
model.add_reaction(my_new_reaction)
model.optimize()
```

My point is that if we want these routines to return/modify model objects, maybe we should be superclassing model, and then either providing add_reaction wrappers that handle the new constraints, or return `NotImplementedErrors`?

but then we might lose speed and the ability to context manage.

Either way, probably not a huge concern... Maybe a simple docstring warning would be sufficient

cdiener:
@pstjohn That is a good point. For now mentioning it in the docstrings and the future docs is the easiest thing. We could add a check in add_reactions but I wouldn't know for what to check right now...

cdiener:
Returning to WIP status since the changes in the Solution class broke the loopless functions.

cdiener:
So here is the gist of the last commits:

Model.objective still had the `@resettable` decorator. That together with the new `set_objective` caused some bugs filling up the contexts. Taking `@resettable` off `Model.objective` uncovered some bugs with `set_objective` where it would first zero the objective and than read the reset value (which is surprisingly zero in this case :) ).

hredestig:
Looks great now, is it still wip?

cdiener:
@hredestig Still figuring out whether loopless solution should have an argument `fluxes` that takes a dict of fluxes. 

Might be useful for flux sampling (convert sample to a loopless solution) or saved flux solutions. The problem there is how to get the objective value from a dict alone. Easy if `Model.objective_coefficients` gives the correct objective, but what happens if I set an objective in the optlang interface directy (non-linear, uses other variables, etc.)?

cdiener:
Ok, should be fine now. Added the fluxes arg as well. 

Note that this also fixes two bugs I found on the way (#416 and another one in util.solver_valid_atoms). Let me know if you want those in a separate PR.

cobra/flux_analysis/loopless.py
```diff
@@ -1,6 +1,262 @@
...
+            if flux >= 0:
+                rxn.lower_bound = max(0, rxn.lower_bound)
+                model.objective.set_linear_coefficients(
+                    {rxn.forward_variable: 1, rxn.reverse_variable: -1})
+            else:
+                rxn.upper_bound = min(0, rxn.upper_bound)
+                model.objective.set_linear_coefficients(
+                    {rxn.forward_variable: -1, rxn.reverse_variable: 1})
+        model.solver.objective.direction = "min"
+        model.optimize(objective_sense=None)
```
hredestig: this raises a SolveError if status is not optimal - be in try/catch?

cdiener:
Okay rebased onto devel-2 :sweat_smile: and added fixes and tests. Also implemented the review from @hredestig. For now this PR also fixes #416 and #418.

PR #392, opened 2017-02-15
skip mac and 3.4

hredestig:
we are troubled by spurious crashses on mac and it's stalling
development so, for now on devel-2

- no mac builds
- no py 3.4
- only build master, devel, devel-2 and pr's to those

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,12 @@ addons:
       - libfreetype6-dev
       - libpng12-dev
 
+branches:
+ only:
+ - master
+ - devel
+ - devel-2
+
 env:
   global:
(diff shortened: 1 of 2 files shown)


hredestig:
Moving further, of course we have to find the solution to `abort trap 6` but let's not let that block progression on other topics

cdiener:
Will be hard to fix by us. I suspect that there is problem with calling external libs so it's probably a problem with travis and/or multibuild/manylinux.

hredestig:
I get it locally too now though, just sometimes, but python crashes every now and then :(

cdiener:
Yikes. Still could be a problem with multibuild. Maybe the Mac wheels are corrupted. Wouldn't be the first time either. Do you get the same problem when building from source?

cdiener:
Not sure if relevant but I just remembered that I had problems with libsbml and newer gcc version where it would compile fine but create a package that would segfault when imported. Maybe just getting rid of libsbml might be something we want to test.

hredestig:
sounds like something to try, a clue is surely that we have never seen this before but it suddenly started happening this last weekend.

PR #395, opened 2017-02-15
Fix for objective setting in nested contexts

### LINKED ISSUE #393: Nested objective resets are broken
Reproduce:

```Python
In [14]: mod = create_test_model("textbook")

In [15]: mod.objective.expression
Out[15]: -1.0*Biomass_Ecoli_core_reverse_2cdba + 1.0*Biomass_Ecoli_core

In [16]: with mod:
    ...:     with mod: 
    ...:         mod.objective = mod.reactions.ATPM
    ...:     

In [17]: mod.objective.expression
Out[17]: 0
```

pstjohn:
adding 'check_context' kwarg to set_objective, fixes #393

Also refactors if statements as suggested by @cdiener

### DIFF
--- a/cobra/util/solver.py
+++ b/cobra/util/solver.py
@@ -62,7 +62,7 @@ def linear_reaction_coefficients(model, reactions=None):
     return linear_coefficients
 
 
-def set_objective(model, value, additive=False):
+def set_objective(model, value, additive=False, check_context=True):
     """ Set the model objective
 
     Parameters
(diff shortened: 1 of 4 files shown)


cobra/util/solver.py
```diff
@@ -62,7 +62,7 @@ def linear_reaction_coefficients(model, reactions=None):
     return linear_coefficients
 
 
-def set_objective(model, value, additive=False):
+def set_objective(model, value, additive=False, check_context=True):
```
pstjohn: yeah this makes sense

cobra/util/solver.py
```diff
@@ -82,45 +82,47 @@ def set_objective(model, value, additive=False):
...
     reverse_value = None
-    if by_objective:
-        if not additive:
-            if isinstance(value, sympy.Basic):
-                value = model.solver.interface.Objective(value, sloppy=False)
-            model.solver.objective = value
-        else:
-            if isinstance(value, model.solver.interface.Objective):
-                value = value.expression
-            model.solver.objective += value
```
pstjohn: should we just plan on dropping `additive`? Probably would make this a lot easier.
Additive objectives can just be handled in the upstream methods

cdiener: Or substitute it for a reset/zero args that zeros the objective before applying the requested changes. This can simply be implemented with a 

```Python
model.solver.objective = model.solver.interface.Objective(S.Zero)
```

at the beginning of the function.

hredestig: remember now why I didn't do like that, if making it always an addition (to S.Zero if not additive) I get this from the moma test 
`E           cplex.exceptions.CplexSolverError: CPLEX Error  5002: Q in objective is not positive semi-definite.`

hredestig:
Hope you don't mind the edits, there was another bug in there which I needed fixed for pfba, I also adjusted according to suggestions. I'm fine with dropping the additive behaviour for non-dicts though if you like.

cobra/util/solver.py
cdiener: That's because you don't set Objective.direction. I commented on that above :)

cobra/util/solver.py
```diff
@@ -83,44 +83,52 @@ def set_objective(model, value, additive=False):
...
-    elif isinstance(value, dict):
+    if isinstance(value, dict):
+        if not model.objective.is_Linear:
+            raise ValueError('can only update non-linear objectives '
+                             'additively using object of class '
+                             'model.solver.interface.Objective, not %s' %
+                             type(value))
         if not additive:
             model.solver.objective = model.solver.interface.Objective(
                 sympy.S.Zero, direction='max')
```
cdiener: Wouldn't it be better if that would respect the direction from the previous objective?

hredestig: haha yes of course, thanks!

cobra/util/solver.py
hredestig: Can't be the only reason, I get it also when I correcting that...

pstjohn:
@hredestig, not at all, thanks for wrapping this up!

Issue #384, opened 2017-02-13
remove reporter metabolites


hredestig:

cdiener:
I would take that one if nobody else has started already. I would like that to be similar to MOMA so you can do something like:

```python
with model:
  loopless(model)
  model.optimize()
  single_reaction_deletion(model)
```

So you can plug-and-play it with other methods.

pstjohn:
I don't even know what cobra.topology is doing 😄 . A lot of that code is ~6 years old at this point. Thumbs up if you want to take a stab at it!

cdiener:
Ah sorry wrong issue :(

hredestig:
Also seems like something better fitting in a different package, shall we instead just drop this one?

PR #400, opened 2017-02-17
Fix/remove defunct untested code

### LINKED ISSUE #383: refactor gapfilling / update  hints

hredestig:
fixes #383 #384


pstjohn:
Would you want to also include those `__add__` methods we talked about removing? In model.py, for instance.

hredestig:
I'd prefer to see them moved out from the Model class I think but fits just as well in a different PR

PR #404, opened 2017-02-18
Context management for Reactions and Metabolites [fix #389]

### LINKED ISSUE #389: fix/refactor add/remove genes,metabolites,reactions
breakout from discussion in #311
these should allow resetting using history manager

pstjohn:
Got some decent progress.. 

Still todo:

- [x] `Model.remove_reactions`
- [x] Implement `Model.remove_metabolites(method)` kwarg.. (will call `Model.remove_reactions`?)
- [x] `Reaction.metabolites`
- [ ] `Reaction.gene_reaction_rule`?

definitely running into the issue of wanting to call other context-managed methods, but not wanting to trigger the issue seen in #393...

@cdiener, I'll probably ask how you do those benchmarks 😄 , we should also make sure we're not torpedoing performance.

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -468,6 +468,90 @@ class Model(Object):
         # from cameo ...
         self._populate_solver(reaction_list)
 
+    def remove_reactions(self, reactions, delete=True,
+                         remove_orphans=False):
+        """remove reactions from the model
+
+        reactions: [:class:`~cobra.core.Reaction.Reaction`] or [str]
+            The reactions (or their id's) to remove
+
+        delete: Boolean
+            Whether or not the reactions should be deleted after removal.
+            If the reactions are not deleted, those objects will be
+            recreated with new metabolite and gene objects.
+
+        remove_orphans: Boolean
+            Remove orphaned genes and metabolites from the model as well
+
+        """
+        if isinstance(reactions, string_types) or hasattr(reactions, "id"):
+            warn("need to pass in a list")
+            reactions = [reactions]
+
+        context = get_context(self)
+
+        for reaction in reactions:
+            try:
+                reaction = self.reactions[self.reactions.index(reaction)]
+            except ValueError:
+                warn('%s not in %s' % (reaction, self))
+            else:
+                forward = reaction.forward_variable
+                reverse = reaction.reverse_variable
+                remove_from_solver(self, [forward, reverse])
+                self.reactions.remove(reaction)
+                reaction._model = None
+
+                if context:
+                    context(partial(setattr, reaction, '_model', self))
+                    context(partial(self.reactions.add, reaction))
+
+                for x in reaction._metabolites:
+                    if reaction in x._reaction:
+                        x._reaction.remove(reaction)
+                        if context:
+                            context(partial(x._reaction.add, reaction))
+                        if remove_orphans and len(x._reaction) == 0:
+                            self.remove_metabolites(x)
+
+                for x in reaction._genes:
+                    if reaction in x._reaction:
+                        x._reaction.remove(reaction)
+                        if context:
+                            context(partial(x._reaction.add, reaction))
+
+                        if remove_orphans and len(x._reaction) == 0:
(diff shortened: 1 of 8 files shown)


pstjohn:
Alright, some questions for you guys.

When would we want to call `Reaction.remove_from_model()`? It copies all the metabolite and gene objects associated with it and creates a free-floating reaction. My opinion is that we likely wouldn't want to have a reaction floating without an associated model... if we wanted to add a reaction from one model to another, we should be able to just directly model.add_reactions([reactions_from_another_model]) and have the metabolite / gene copying be handled for us, right?

Midnighter:
As outlined elsewhere, I agree with you. `Model` should have the methods to add and remove. I would scrap `X.remove_from_model` completely.

pstjohn:
Well I think we'll need to keep it for backwards-compatibility? At least for now.
But my main question is do we want the Model.remove_reactions() function to mimic `delete` or `remove_from_model`.. I probably think `delete`, since I'm not sure of the use case for remove_from_model.

cdiener:
Thanks for getting started with this @pstjohn :) One question: what about `Reaction.add_metabolites`, `Reaction.subtract_metabolites`, and the `Reaction.reaction` setter?

pstjohn:
Yeah, those will definitely have to be context-managed...
Reaction.reaction will probably be able to use the resettable decorator. The other ones might be a bit trickier. Slammed this week sadly, so this will probably have to wait until next week for me.

pstjohn:
yikes, that rebase didn't go according to plan.
I might as well rethink how I'm going to do this

cobra/core/model.py
```diff
@@ -556,68 +693,6 @@ def optimize(self, objective_sense='maximize', solution_type=Solution,
...
-                reaction = self.reactions[self.reactions.index(reaction)]
-            except ValueError:
-                warn('%s not in %s' % (reaction, self))
-            else:
-                if delete:
-                    reaction.delete(remove_orphans=remove_orphans)
-                else:
-                    reaction.remove_from_model(remove_orphans=remove_orphans)
-
-    def repair(self, rebuild_index=True, rebuild_relationships=True):
```
pstjohn: I don't know why this repair change is getting wrapped up with the remove reactions change... No intention to edit here

pstjohn:
This may be ready if we pass CI.

We could probably context manage the genes, but tbh I'm not sure how useful that would be.
Should probably get to it eventually, but in the meantime this seems fine.

cobra/test/test_model.py
```diff
@@ -267,6 +295,51 @@ def benchmark_add_reaction():
...
+
+        assert test_metabolite._model is model
+        assert test_metabolite in model.metabolites
+        assert test_metabolite.id in model.solver.constraints
+
+    def test_remove_metabolite_destructive(self, model):
+        test_metabolite = model.metabolites[4]
+        test_reactions = test_metabolite.reactions
+        with model:
+            model.remove_metabolites(test_metabolite, method='destructive')
```
cdiener: If `method` is a binary choice maybe better convert it to a bool arg? Less danger of passing wrong type.

cobra/util/solver.py
```diff
@@ -294,7 +296,7 @@ def add_to_solver(model, what):
     """
     context = get_context(model)
 
-    model.solver.add(what)
+    model.solver.add(what, **kwargs)
```
cdiener: Good idea!

cobra/test/test_model.py
pstjohn: Good point. Kept it for compatibility, but it was never in the tests. In fact, `metabolite.remove_from_model(method='destructive')` was broken until this fix (at least on python 3), so I doubt anyone was using it.

cdiener: Ah sorry, my bad. Though you added the arg here. In this case it should stay as is for backward compat :)

pstjohn:
Great idea, thanks for the docstring updates!

PR #413, opened 2017-02-21
Remove version

Midnighter:
Endless annoyances with `coba.version.py`... switch to bumpversion.


cdiener:
What problems are there? I usually prefer versioneer over bumpversion since it allows to manage releases completely through Github...

setup.cfg
```diff
@@ -0,0 +1,29 @@
+[bumpversion]
+current_version = 0.5.11
+commit = True
+tag = True
+
+[bumpversion:file:setup.py]
+search = version="{current_version}"
+replace = version="{new_version}"
+
+[bumpversion:file:memote/__init__.py]
```
cdiener: What is memote?

hredestig:
I like bumpversion due to it's simplicity and transparency. `cobra.version` keeps editing the `VERSION` file when you don't ask for it and it's easy to make mistakes formatting the tag wrongly. Make a mistake in the order of commits and suddenly you have released x.y.zpost234 to pypi. With bumpversion the tag get's placed right after the commit so doesn't happen.

`cobra.version` and `versioneer.py`'s advantages appear to be to solve problems that I personally never experienced..

setup.cfg
Midnighter: Whoops, thanks for the catch.

cdiener:
I think that is preference. For me bumpversion never solved any problems I had :) but I agree that the custom version.py was not optimal.

PR #417, opened 2017-02-22
Feature: data frame as return value [fix #412]

### LINKED ISSUE #412: Make analysis functions return data frames
Since there was some agreement on making pandas a hard requirement, shall we make FVA return a data frame? Other methods as well, where applicable?

hredestig:
- add a `return_frame` argument to pfba, fva, single_x_deletion
- reaction / gene id is in the index but perhaps could think of further standardisation of the columns but also that this will be easier once we start thinking of plotting options (in a different package)
- bit ugly whilst we still work on the updated Solution class

### DIFF
--- a/cobra/core/metabolite.py
+++ b/cobra/core/metabolite.py
@@ -9,6 +9,7 @@ from six import iteritems
 
 from cobra.core.formula import elements_and_molecular_weights
 from cobra.core.species import Species
+from cobra.flux_analysis.summary import metabolite_summary
 
 # Numbers are not required because of the |(?=[A-Z])? block. See the
 # discussion in https://github.com/opencobra/cobrapy/issues/128 for
@@ -137,9 +138,10 @@ class Metabolite(Species):
         except Exception as e:
             if self._model is None:
                 raise Exception("not part of a model")
-            if not hasattr(self._model, "solution") or \
-                    self._model.solution is None or \
-                    self._model.solution.status == "NA":
+            not_solved = (not hasattr(self._model, "solution") or
+                          self._model.solution is None or
+                          self._model.solution.status == "NA")
+            if not_solved:
                 raise Exception("model has not been solved")
             if self._model.solution.status != "optimal":
                 raise Exception("model solution was not optimal")
@@ -157,9 +159,10 @@ class Metabolite(Species):
         except Exception as e:
             if self._model is None:
                 raise Exception("not part of a model")
-            if not hasattr(self._model, "solution") or \
-                    self._model.solution is None or \
-                    self._model.solution.status == "NA":
+            not_solved = (not hasattr(self._model, "solution") or
+                          self._model.solution is None or
+                          self._model.solution.status == "NA")
+            if not_solved:
                 raise Exception("model has not been solved")
             if self._model.solution.status != "optimal":
                 raise Exception("model solution was not optimal")
@@ -196,8 +199,10 @@ class Metabolite(Species):
 
     def summary(self, threshold=0.01, fva=False, floatfmt='.3g', **kwargs):
         """Print a summary of the reactions which produce and consume this
-        metabolite. This method requires the model for which this metabolite is
-        a part to be solved.
+        metabolite.
+
+        This method requires the model for which this metabolite is a part
+        to be solved.
 
         Parameters
         ----------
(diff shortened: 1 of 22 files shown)


cdiener:
Hi, only so I understand it correctly: the `return_frame` argument will be added in 0.6 but removed again in 0.7 (where pandas DFs become defaults)?

hredestig:
Yes a bit dodgy :/ Introduces two breaking changes, have a better suggestion? If we can't return data frames now for compatibility but want the functionality already I guess we need that argument. As we move further, I don't really see the point of returning anything else than a data frame as they are pretty much equivalent.

cdiener:
I agree that the default return type in 0.7 should be DFs. Hmm, maybe just write helper functions that convert the dict outputs to DFs. This way you could use them in 0.6 to convert the output to a DF and in 0.7 those get called in the functions before the return. But I don't know whether the outputs now are so homogeneous that this will be possible with only 1-2 helpers...

pstjohn:
Just return the pandas dataframes in 0.6 😄 .

If you just wrap the FVA return object with a `pd.DataFrame()`, you only have to change the last line `variability.py` (`v.values()` to just `v`), and all the tests pass.

A dataframe really does behave a lot like a nested dict, so I'm not overly concerned about breaking third-party code when all our tests pass unchanged.

cdiener:
Thanks for testing that @pstjohn. This is a strong argument to move sooner.

hredestig:
completely agree on this, let's make pandas hard requirement and data frame default return value where meaningful

cobra/core/metabolite.py
```diff
@@ -217,4 +217,4 @@ def summary(self, threshold=0.01, fva=False, floatfmt='.3g', **kwargs):
             return metabolite_summary(self, threshold=threshold, fva=fva,
                                       floatfmt=floatfmt, **kwargs)
         except ImportError:
-            warn('Summary methods require pandas/tabulate')
+            warn('Summary methods require tabulate')
```
cdiener: I think we could just add tabulate to the dependencies as well. It's a pure Python dep and summsry methods are pretty essential by now.

pstjohn:
How would you guys feel about having `array_type` for `create_stoichiometric_array` give the option to return a dataframe as well?

hredestig:
Sure, don't see a reason not to, added

cdiener:
Still missing sampling :smile: This would make it much clearer which dimensions are samples/reactions.

hredestig:
Good point, made default return dataframe for sampling as well

cobra/test/test_flux_analysis.py
```diff
@@ -78,15 +68,14 @@ def do_pfba(solver):
     def test_pfba(self, model, solver):
         expression = model.objective.expression
         n_constraints = len(model.solver.constraints)
-        optimize_minimal_flux(model, solver=solver)
-        abs_x = [abs(i) for i in model.solution.x]
+        df = optimize_minimal_flux(model, solver=solver)
+        assert numpy.all([df.columns.values == ['flux', 'objective_value']])
```
Midnighter: Is exact equality required here or would `numpy.allclose` be better?

cobra/test/test_flux_analysis.py
```diff
@@ -78,15 +68,14 @@ def do_pfba(solver):
...
         assert model.solution.status == "optimal"
         if isinstance(model.solution, LegacySolution):
             assert abs(model.solution.f - 0.8739) < 0.001
         else:
-            assert abs(
-                model.solution.fluxes['Biomass_Ecoli_core'] - 0.8739) < 0.001
-        assert abs(sum(abs_x) - 518.4221) < 0.001
+            assert abs(df.flux['Biomass_Ecoli_core'] - 0.8739) < 0.001
```
Midnighter: This (and following) is basically a home-grown `numpy.isclose`.

hredestig: True, but also only checking scalars so I think just as good and easy to read like this.. One may consider having whole results and not just checksums. But leave that for another pr.

cobra/test/test_flux_analysis.py
hredestig: Checking the column names here, so should be identical but perhaps there is better way to do it..

Midnighter: Whoops ;)

cobra/test/test_flux_analysis.py
pstjohn: The sum is because we're testing the parsimonious solution, which minimizes the sum of the absolute value. So I think its appropriate. (there might be multiple equally valid parsimonious solutions)

PR #422, opened 2017-02-24
Feat/production envelope [fix #382]

### LINKED ISSUE #382: refactor phenotype phase plane

hredestig:
Introduce a function to obtain a data frame with a production envelope
for one or more variable reactions and a given objective, including
carbon and mass yield conversions.

- Only to be merged after the new Solution class is in place

### DIFF
--- a/cobra/flux_analysis/phenotype_phase_plane.py
+++ b/cobra/flux_analysis/phenotype_phase_plane.py
@@ -442,13 +442,9 @@ def _carbon_flux(reaction):
                  metabolite in reaction.reactants)
 
     try:
-        return flux(reaction) * carbon
+        return reaction.flux * carbon
     except AssertionError:
         return nan
-
-
-def flux(rxn):
-    return rxn.forward_variable.primal - rxn.reverse_variable.primal
 
 
 def _carbon_yield(c_input_output):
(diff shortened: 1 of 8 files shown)


cobra/flux_analysis/phenotype_phase_plane.py
```diff
@@ -2,12 +2,20 @@
...
 
+from warnings import warn
+from collections import defaultdict
+from itertools import product
 from multiprocessing import Pool
+import pandas as pd
 
 from numpy import (
-    abs, arange, dtype, empty, int32, linspace, meshgrid, unravel_index, zeros)
+    nan, abs, arange, dtype, empty, int32, linspace, meshgrid, unravel_index,
```
Midnighter: With this many imports I tend to use `import numpy as np` and then just go with `np.nan` etc.

pstjohn:
There's already a pickled solution for the textbook model in tests/data, I used it in the tests for the summary methods.

Midnighter:
Unfortunately, that pickled solution is generated in the following way:
```Python
solution = cobra.flux_analysis.parsimonious.optimize_minimal_flux(textbook)
with open('textbook_solution.pickle', 'wb') as f:
    dump(solution, f, protocol=2)
```
So it's generated whenever `update_pickles.py` is called and thus not a real gold standard.

PR #425, opened 2017-02-28
fix: array model is deprecated but not defunct - remove model.S

hredestig:
so let's keep the tests there and fix module renaming mistake


hredestig:
model.S was a new feature but as discussed on gitter, suggestion to rename it something easier to understand triggered the conclusion to remove altogether instead, so here it goes, suggesting to drop it again.

PR #430, opened 2017-03-01
feat: add constraint or variable

hredestig:
Whilst updating the notebooks I find the `add_to_solver` utility to be confusing because while we have some (currently unavoidable) convolution between the `Model` and the `Model.solver`, it doesn't feel necessary to explain this to the casual user, and so I would like a `Model.add_constraint_or_variable` (long name though..) as a wrapper to `add_to_solver`.

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -649,7 +649,7 @@ class Model(Object):
             forward_variable = self.problem.Variable(
                 reaction.id, lb=forward_lb, ub=forward_ub)
             reverse_variable = self.problem.Variable(
-                reaction._get_reverse_id(), lb=reverse_lb, ub=reverse_ub)
+                reaction.reverse_id, lb=reverse_lb, ub=reverse_ub)
 
             self.add_cons_vars([forward_variable, reverse_variable])
             self.solver.update()
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -21,7 +21,8 @@ from cobra.solvers import optimize
 from cobra.util.context import HistoryManager, resettable, get_context
 from cobra.util.solver import (
     SolverNotFound, get_solver_name, interface_to_str, set_objective, solvers,
-    add_to_solver, remove_from_solver, choose_solver, check_solver_status)
+    add_cons_vars_to_problem, remove_cons_vars_from_problem, choose_solver,
+    check_solver_status)
 from cobra.util.util import AutoVivification
 
 
@@ -107,7 +108,7 @@ class Model(Object):
         --------
         >>> import cobra.test
         >>> model = cobra.test.create_test_model("textbook")
-        >>> new = model.solver.interface.Constraint(model.objective.expression,
+        >>> new = model.problem.Constraint(model.objective.expression,
         >>> lb=0.99)
         >>> model.solver.add(new)
         """
@@ -133,7 +134,7 @@ class Model(Object):
             raise not_valid_interface
 
         # Do nothing if the solver did not change
-        if self.solver.interface == interface:
+        if self.problem == interface:
             return
 
         for reaction in self.reactions:
@@ -301,7 +302,6 @@ class Model(Object):
         except Exception:  # pragma: no cover
             new._solver = copy(self.solver)  # pragma: no cover
 
-        # No use in copying it, also circular dependencies
         return new
 
     def add_metabolites(self, metabolite_list):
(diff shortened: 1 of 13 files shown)


hredestig:
This actually calls for adding a wrapper to `model.solver.variables`, `model.solver.constraints` as well (`model.variables`, `model.constraints`)

Which would be inline with existence of `model.objective`.

What do you think?

cdiener:
:+1: to model.variables and model.constraints. I would prefer model.problem instead of model.interface since it actually is the LP/MIP/QP problem underlying the model. Model.add_constraint_or_variable is a bit longish for me as well, maybe Model.add_to_problem?

cobra/core/model.py
```diff
@@ -414,37 +415,119 @@ def add_reactions(self, reaction_list):
...
+        upon exit if the model itself is used as context.
+
+        Parameters
+        ----------
+        what : list or tuple of optlang variables or constraints.
+           The variables or constraints to add to the model. Must be of
+           class `optlang.interface.Variable` or
+           `optlang.interface.Constraint`.
+        """
+        add_to_solver(self, what)
```
pstjohn: In #404 I had to add a kwarg to `add_to_solver` when some function needed `sloppy=True`.. Maybe we'll need the same thing here?

cobra/core/model.py
```diff
@@ -414,37 +415,119 @@ def add_reactions(self, reaction_list):
...
                 separate_forward_and_reverse_bounds(*reaction.bounds)
 
-            forward_variable = self.solver.interface.Variable(
+            forward_variable = self.problem.Variable(
                 reaction.id, lb=forward_lb, ub=forward_ub)
-            reverse_variable = self.solver.interface.Variable(
+            reverse_variable = self.problem.Variable(
                 reaction._get_reverse_id(), lb=reverse_lb, ub=reverse_ub)
 
             self.solver.add(forward_variable)
```
pstjohn: So one question I always had was whether or not these would want to get context-managed.
Ideally I'd think everything would go through `self.add_to_problem`, right?

cobra/flux_analysis/loopless.py
```diff
@@ -75,11 +75,11 @@ def add_loopless(model, zero_cutoff=1e-12):
         name = "nullspace_constraint_" + str(i)
         nullspace_constraint = prob.Constraint(S.Zero, lb=0, ub=0, name=name)
         add_to_solver(model, [nullspace_constraint])
```
pstjohn: I think we should standardize how we want to call these functions, `add_to_solver(model, ...)` vs. `model.add_to_problem(...)`

cobra/test/test_solver_model.py
```diff
@@ -674,15 +674,15 @@ def test_objective(self, model):
         assert obj.direction == "max"
 
     def test_change_objective(self, model):
-        expression = 1.0 * model.solver.variables['ENO'] + \
-                     1.0 * model.solver.variables['PFK']
-        model.solver.objective = model.solver.interface.Objective(
+        expression = 1.0 * model.variables['ENO'] + \
+                     1.0 * model.variables['PFK']
+        model.solver.objective = model.problem.Objective(
```
pstjohn: So is there now a difference between assigning to `model.objective` vs. `model.solver.objective`? Would it be possible to replace all calls to the latter with just `model.objective`?

cdiener: model.objective integrates with context manager, model.solver.objective does not. For tests we should use model.objective I guess...

hredestig: Good point, I fixed this in the tests

hredestig:
So the new proposed name `model.add_cons_vars` is admittedly not ideal but `add_to_problem` is  also confusing as doesn't answer what we are adding and to what is also confusing because it looks like add to model... Other suggestions are welcome. 

On making model.solver private, I see your point but then I also don't think we should use it anywhere else than in the model class itself. Fairly big change..

cobra/core/model.py
hredestig: Good point, added

cobra/core/model.py
hredestig: Yes, updated and merged with your pr, now we use model.add_cons_vars where possible

cobra/flux_analysis/loopless.py
hredestig: Yup, done

Issue #428, opened 2017-02-28
weed out the packaged solvers


Midnighter:
I would like to propose to clean up the `cobra.solvers`. So far in the tests I've only seen the legacy "cglpk" and "esolver" show up. There are a couple of other solver interfaces there, though. Are any other ones functional and could be configured to be run by the tests, too? I don't know if there's a good way for us to poll which ones are in use out there. It would be good to remove unnecessary ones, though.

cdiener:
cplex, gurobi and mosek will be used by the tests if installed. See for instance the benchmarks where you will see `[cplex]` and also `[optlang-cplex]`. But I think this will be easier in 0.7 where we will delete the entire folder anyways...

Midnighter:
Another thing about solvers that has been bothering me and is not correctly handled everywhere is the `solver` argument to analysis functions. I vote for removing that argument when we completely rely on optlang and don't support the legacy solvers any longer. I think setting `model.solver` (or whatever we name it) should be the one obvious way to do it and is totally sufficient.

cdiener:
`Model.optimize`, `single_*_deletions` and FVA handle the solver argument correctly and use `Model.solver` when set to None. I agree that this should be removed along with the **solver_kwargs as mentioned in #386.

PR #431, opened 2017-03-01
Refactor solution [fix #411]

### LINKED ISSUE #411: Refactor the solution class (part 2)
**Suggestion:**

* Let solution not be lazy but its contained values are determined by a parameter to `model.optimize` which returns the solution.
* The attribute `model.solution` will be ~~deprecated.~~ removed
* ~~`Reaction.flux` and similar will continue to use the solution attribute but also be deprecated.~~

EDIT (hredestig): don't deprecate reaction.flux (see below), and completely remove `model.solution`

Midnighter:
I'm just opening a pull request for better review of the changes and because right now my brain is too exhausted to deal with the rest.

## Summary

* Remove the `model.solution` attribute.
* Make `model.optimize()` return a solution that is either explicitly `Solution` or `LegacySolution`.
* `Solution` attributes are now `pandas.Series` where that makes sense. This means you can now also access flux values this way: `solution.fluxes.ATPM` (no tab-completion until patching of `pandas.Series`).
* Creating solutions with the new helper function `get_solution` (in `cobra/util/solver.py`) and the `Reaction` and `Metabolite` properties now only issue a warning if the status is non-optimal. I'm not sure that is the best design decision.
* `Model` now has an `is_dirty` flag which tracks modifications and accessing solution values (via `Reaction` or `Metabolite`) will raise a `RuntimeError` if the model is dirty.
* `is_dirty` is part of the context machinery and properly resets the flag of the old context.
* Introduce an `OptimizationError` exception which I'm not using anywhere since I only warn about non-optimal status :) but sounds better than SolveError and could replace that.
* Patch the pickled models.
* Fix all the tests to not use `Model.solution`.
* Make legacy solvers use `LegacySolution`.

## Conclusion

Overall this is a much more extensive pull request than I anticipated.

Besides the merge conflicts there are only three unexpected tests in `test_flux_analysis.py` that fail:

1. `test_summary_methods` I don't know if I inadvertently changed the summary output but some desired entries are not found.
2. In `test_pfba` there are some `optlang` level assertion errors on variable bounds because I probably broke `convert_to_irreversible` and its cousins.
3. I'm really not sure what the expected output is supposed to be in `test_loopless_solution_fluxes`. I added a comment there and in the actual function itself.
4. The old 3D plotting library missing failure that was fixed in another branch.


cdiener:
The loopless tests fail because they expect model.optimize to issue an error if the solution is not optimal. I am not sure what the best behavior for Model.optimize should be but @hredestig had some opinions I think. 

Also depending on the actual value of `model.solver.status` the primal and dual values you get might be wrong. For infeasible solutions it gives you the values of the infeasible solution as intended but I am not sure what happens for other solver status values. In the worst case it would give you the values from a previous optimize call and not the current one.

Sorry, I don't understand what `is_dirty` is supposed to do from looking at your code :cry: Especially the `Model._contexts_dirty` is confusing for me...

Midnighter:
> Sorry, I don't understand what is_dirty is supposed to do from looking at your code :cry: Especially the Model._contexts_dirty is confusing for me...

It is meant to solve the problem that occurs when I optimize a model, then modify something about it, and afterwards try to access, for example, `Reaction.flux`. Which should not be defined until the model was optimized again. Since not all solvers raise an error like cplex does, I thought it would be best to track on the `cobra.Model` itself.

The `_contexts_dirty` is simply a list that tracks the `is_dirty` flag within different contexts. So when a new context is entered the current state of the flag is stored and when the context is left (and its operations undone) the old flag is restored. Although I now realize that even with contexts the underlying state of the solver will have changed so it doesn't really make sense to do that.

hredestig:
I initially liked the idea of the dirty flag but looking at this it seems quite complicated.. In addition, if I directly use model.solver.interface.add the dirty flag won't be set. One can argue I shouldn't do that but that will be one confusing doc string.. Isn't it better then to catch any errors in reaction.flux, convert them to something understandable and let any wrong values simply be wrong, and argue in the docs that this is from the last optimization, whenever that was, and that for safety you're advised to save a solution instead. Perhaps we can get a dirty-flag directly in the optlang interface later on where implementing such a feature may be easier.

Midnighter:
> The loopless tests fail because they expect model.optimize to issue an error if the solution is not optimal.

I actually covered that part but I'm unsure what to do about the last line of the test `assert ll_fluxes is None`. I don't know how to sensibly change loopless itself to return the right thing.

Midnighter:
What you propose @hredestig is certainly a lot easier. And you're correct, the design can be broken anyway. So I agree it would be much better to have that be part of optlang which is supposed to give us a unique interface to solvers already.

cdiener:
Thanks for the explanation. I agree with @hredestig. 

Does Metabolite have a `._model` attribute like Reaction? If yes, we could treat it like there and throw an error if `._model` is None. This way we only have to make sure all reaction/metabolite removal correctly resets this attribute... 

For loopless you can change the code in lines 168-173 to something like:

```Python
model.solver.objective.direction = "min"
sol = model.optimize(objective_sense=None)
if model.solver.status == "optimal":
    fluxes = model.solution.fluxes
else:
    fluxes = None
```

That should work with the current implementation. Is the value of model.solver.status saved in `Solution`?

Midnighter:
> Does Metabolite have a ._model attribute like Reaction? If yes, we could treat it like there and throw an error if ._model is None. This way we only have to make sure all reaction/metabolite removal correctly resets this attribute...

That is already the case. `Metabolite.shadow_price` and `Reaction.flux` are almost identical. Of course, checking for `self._model` being `None` doesn't help when you're adding or removing other things to the model but I'll just stick a big fat warning in the doc strings.

> That should work with the current implementation. Is the value of model.solver.status saved in Solution?

OK, great. I wasn't sure if it should be an individual flux being `None` or the entire distribution. That makes it easy.

The status is indeed stored on the solution. Currently, `Solution` has the following attributes:

    reactions : iterable
        A list of `cobra.Reaction` objects for which the solution is
        retrieved.
    objective_value : float
        The (optimal) value for the objective function.
    status : str
        The solver status related to the solution.
    fluxes : pandas.Series
        Contains the reaction fluxes (primal values of variables).
    reduced_costs : pandas.Series
        Contains reduced costs (dual values of variables).

With the `reduced_costs` in the current creation process being optional. (I might make that always present for consistency and since it's hardly any effort.)

Do you think the solution should also contain metabolites and their shadow prices?

cdiener:
> Do you think the solution should also contain metabolites and their shadow prices?

At least that would be consistent with the implementation of the column primals (fluxes etc.). Might be a speed consideration, given that most of the time you don't need them. But in general, yes, add everything to the solution and recommend that the user gets all info from there.

> I'll just stick a big fat warning in the doc strings.

I think that is sufficient. Also hard to implement. Adding other things to the model or changing bounds might also *not* change the solution for that metabolite. We just have to assume the user knows what he is doing.

Midnighter:
I talked to Niko about tracking model status in optlang (aka the dirty flag) and he was obviously not super enthusiastic about making those changes. (I didn't expect them to happen right away anyway.) His alternative solution suggestion was, however, to just call optimize again in `Reaction.flux` and the like. That should basically be a NOOP if nothing has changed.

I like about his suggestion that it is an easy fix and it guarantees that the retrieved flux value is current. The disadvantages are that even though it should be a NOOP, there's a little overhead which probably mostly concerns you @cdiener since you use `Reaction.flux` to quickly access the values. I also dislike that asking for a property then has the side-effect of potentially optimizing the current model.

For now I would probably stick with warning in the docstring. What do you guys think?

Midnighter:
Just to clarify at the moment I can run the following code for either cplex, gurobi, or glpk without getting an error message:

```Python
import cobra
import cobra.test
model.solver = "optlang-gurobi"
model.optimize()
model.reactions.Biomass_Ecoli_core.flux
model.reactions.PFK.remove_from_model()
model.reactions.EX_glc__D_e.bounds = (-5, 100)
model.reactions.Biomass_Ecoli_core.flux
```

Midnighter:
I've reverted to raising an error in solution creation if status is non-optimal. Otherwise we run into optlang assertion errors on the flux bounds. Getting fluxes directly `Reaction.flux` raises a warning and may cause the optlang assertion error.

I'm not really sure what to do about `test_summary_methods`. Due to the problems with setting solvers that I've also mentioned it in #428, results are really inconsistent and I don't want to go in and change the test until that is fixed.

cobra/core/metabolite.py
```diff
@@ -131,39 +131,29 @@ def y(self):
...
-                raise Exception("not part of a model")
-            if not hasattr(self._model, "solution") or \
-                    self._model.solution is None or \
-                    self._model.solution.status == "NA":
-                raise Exception("model has not been solved")
-            if self._model.solution.status != "optimal":
-                raise Exception("model solution was not optimal")
-            raise e
+            if self._model.solver.status != "optimal":
+                warn("Solver status is not optimal ({}), please treat value"
```
cdiener: If I call that in a loop (which often happens) that might pollute quite a lot. Maybe use the logger?

cobra/core/model.py
```diff
@@ -489,71 +483,58 @@ def to_array_based_model(self, deepcopy_model=False, **kwargs):
...
-            # invalidate their solution if the model is changed afterwards
-            self.solver.optimize()
-            # Not nice, but necessary until next optlang release
-            solution = solution_type(self)
-            # solution = (solution_type(self) if
-            #             self.solver.status == 'optimal' else self.solution)
-            if objective_sense is not None:
-                self.solver.objective.direction = original_direction
-        else:
+        (legacy, solver) = choose_solver(self, solver=kwargs.get("solver"))
```
cdiener: Does not need the brackets around `legacy, solver`.

cobra/core/model.py
```diff
@@ -489,71 +483,58 @@ def to_array_based_model(self, deepcopy_model=False, **kwargs):
...
-            # TODO: make failing optimization raise suitable exception
-            # raise exceptions._OPTLANG_TO_EXCEPTIONS_DICT.get(solution.status,
-            #                                                  SolveError)(
-            #     'Solving model %s did not return an optimal solution. The '
-            #     'returned solution status is "%s"' % (
-            #         self, solution.status))
+        else:
+            original_solver = self.solver
+            original_direction = self.solver.objective.direction
+            self.solver = solver
```
cdiener: `Model.solver` is context managed, so is Model.objective. So maybe use a with block here?

cobra/util/solver.py
```diff
@@ -386,3 +386,59 @@ def fix_objective_as_constraint(model, fraction=1):
         model.objective.expression,
         name=fix_objective_name, ub=ub, lb=lb)
     add_to_solver(model, constraint)
+
+
+def get_solution(model, reactions=None, metabolites=None):
```
cdiener: This looks like it should be part of the solution.py.

cobra/core/metabolite.py
Midnighter: I haven't tested but I was always under the impression that repeated exact same warnings are swallowed whereas log output is always written (unless filtered by level). I could also remove the check completely and make it a further comment in the docstring.

cobra/core/model.py
Midnighter: Yeah, personally I prefer the parentheses but I should adapt the project style.

cobra/core/model.py
Midnighter: Neat, I wasn't aware that the solver is managed! I'll put that in. Although the problem remains that things are handled differently by `Model.optimize`, `fba`, `optimize_minimal_flux`, etc.

cobra/util/solver.py
Midnighter: I figured I'm creating a solution from a solver state so it should go there. I have no preference, though, so I'll stick it in `solution.py`.

cobra/core/metabolite.py
cdiener: Ah ok. Didn't know that. In that case I think whatever you choose is fine with me :)

cobra/core/model.py
cdiener: What is managed differently? I agree that some of my code does not use some of the helper functions that were introduced later like `fix_objective_as_constraint`, etc but I'm still planning to fix that...
I know that FVA does not use the context manager as much since that resetting the contexts is slower than running the entire FVA...

Midnighter: Well, it's awesome that the solver can be context managed but it leads to some really confusing behavior, I believe. Imagine the following scenario: You create a model, call `model.optimize(solver='xy')`, inside that you have the context block using the solver for optimization and returning the solution. Then you go ahead and inspect `Reaction.flux` which could give you the value of a different solver and of a different state. I would just find that terribly confusing.

cobra/core/metabolite.py
Midnighter: I just tested in a loop and that's indeed the case. However, I wonder if that is the right behavior... If there is no warning in the consecutive value retrievals a user might think that everything is fine.

Midnighter:
Alright, one step closer. I ended up not using context management inside of `Model.optimize` because it was breaking our tests all over. It went so far that constraint tests got broken because switching solvers inside context reset all of those, too. One remaining test failure is the summary methods.

Midnighter:
A lot of headache is being caused by `choose_solver` doing the right thing but then I cannot set a legacy solver on the model, i.e., `model.solver = choose_solver()[1]` does not work.

pstjohn:
For fluxes/shadow prices/reduced costs, I would probably vote for None when the model is not solved, and np.nan when the solution is infeasible?

I think that's pretty unambiguous that there's a problem, but it would let you loop over various conditions (some of which may be infeasible) without an explicit try/except.

cdiener:
@pstjohn you mean in the Solution object or for Reaction.flux, etc (or both)? For infeasible states I sometimes like to check the resulting primals to see which constraint causes the infeasibility, but I would be fine with doing that in the optlang interface directly...

pstjohn:
I suppose that's true, I do that as well...

Maybe just flux values then? Or can you have a defined flux value for an infeasible solution?

cdiener:
You can. Will usually be the flux value closest to optimum but might violate bounds or other constraints. The question is how much of that do we want to expose to the user?

Midnighter:
> For infeasible states I sometimes like to check the resulting primals to see which constraint causes the infeasibility, but I would be fine with doing that in the optlang interface directly...

I was thinking of enabling this exact use-case during the development by just warning about non-optimal states rather than raising an error. What I discovered, though, is that `optlang` has an assertion that yells at you when you try to access a primal and it is not within the bounds.

cobra/flux_analysis/loopless.py
```diff
@@ -166,11 +166,10 @@ def loopless_solution(model, fluxes=None):
                     {rxn.forward_variable: -1, rxn.reverse_variable: 1})
 
         model.solver.objective.direction = "min"
-        try:
-            model.optimize(objective_sense=None)
-            fluxes = model.solution.fluxes
-        except SolveError:
-            fluxes = None
+
```
hredestig: remember to put the try/except back...

cdiener:
@Midnighter I discussed the bounds check for infeasible thing with the optlang guys. They agreed that we should be able to get the primals for infeasible solutions without the bounds check and that change has been merged. IDK if it's already released though.

Midnighter:
OK, that's cool. At the moment `Reaction.flux` checks only for `"optimal"`, should I change that to include any other status than "infeasible"?

cdiener:
Those two are fine with me.

Midnighter:
Alright, next and hopefully final version of this PR. I worked on the summary functions a bit and split up the tests for it. Also slightly changed the behavior of `Reaction.flux` again. Completely fixing the summary and FVA therein is a separate PR IMO. I did my best and only skip a few of the tests now.

cobra/test/test_flux_analysis.py
```diff
@@ -387,25 +387,56 @@ def test_phenotype_phase_plane(self, model):
...
+            'Biomass_Ecol...  0.874',
+        ]
+        # Need to use a different method here because
+        # there are multiple entries per line.
+        model.optimize()
+        with captured_output() as (out, err):
+            model.summary()
+        self.check_in_line(out.getvalue(), expected_entries)
+
+    @pytest.mark.skipif((pandas is None) or (tabulate is None),
```
hredestig: pandas is a hard requirement now, as is tabulate

cobra/core/reaction.py
```diff
@@ -263,21 +266,111 @@ def bounds(self, value):
...
+        >>> import cobra
+        >>> import cobra.test
+        >>> model = cobra.test.create_test_model("textbook")
+        >>> solution = model.optimize()
+        >>> model.reactions.PFK.flux
+        7.477381962160283
+        >>> solution.fluxes.PFK
+        7.4773819621602833
+        """
+        try:
```
cdiener: Those are a lot of error checks and raises. Can't we just return None if the first try fails?

Midnighter: I agree they are a lot of checks and I would like to reduce them. However, the detailed checks are there in order to provide as helpful feedback as possible. If I just get `None` as a user I will have to hunt down the source myself rather than being pointed at the likely reason right away. Hopefully the first try will simply go through in 99% of the cases.

cobra/test/test_flux_analysis.py
Midnighter: True, I just copied the summary test which still had the statements. I guess the test files are not cleaned up in this regard.

PR #433, opened 2017-03-05
Add compartments and exchanges properties to `Model`

Midnighter:
These are minor additions that I created for memote.

`Model.exchanges` is behavior that is already found on the cameo model and I've seen it needed in the codebase.

`Model.comartments` seems better to me than `Model.get_compartments()` and also returns a set rather than a list.


cobra/core/model.py
```diff
@@ -543,6 +543,14 @@ def remove_reactions(self, reactions, delete=True,
                 reaction._metabolites = {}
                 reaction._genes = set()
 
+    @property
+    def exchanges(self):
+        """Exchange reactions in model.
+
+        Reactions that either don't have products or substrates.
+        """
+        return [rxn for rxn in self.reactions if rxn.boundary]
```
hredestig: can use `self.reactions.query(lambda rxn: rxn.boundary)` as well

PR #438, opened 2017-03-06
fix: return solution even when reduced costs undefined

hredestig:
for integer problems reduced costs are not well defined so some toy
models don't get any.

### DIFF
--- a/cobra/core/solution.py
+++ b/cobra/core/solution.py
@@ -8,7 +8,7 @@ import logging
 from builtins import object, super
 from warnings import warn
 
-from numpy import zeros, asarray
+from numpy import zeros, asarray, nan
 from pandas import Series
 
 from cobra.util.solver import check_solver_status
@@ -300,7 +300,7 @@ def get_solution(model, reactions=None, metabolites=None):
             reduced[i] = var_duals[forward] - var_duals[reverse]
         except TypeError:
             # reduced costs are not always defined, e.g. for integer problems
-            pass
+            reduced[i] = nan
     met_index = [met.id for met in metabolites]
     constr_duals = model.solver.shadow_prices
     shadow = asarray([constr_duals[met.id] for met in metabolites])
--- a/cobra/core/solution.py
+++ b/cobra/core/solution.py
@@ -298,11 +298,11 @@ def get_solution(model, reactions=None, metabolites=None):
         for (i, rxn) in enumerate(reactions):
             fluxes[i] = var_primals[rxn.id] - var_primals[rxn.reverse_id]
     else:
-    for (i, rxn) in enumerate(reactions):
-        forward = rxn.id
-        reverse = rxn.reverse_id
-        fluxes[i] = var_primals[forward] - var_primals[reverse]
-        reduced[i] = var_duals[forward] - var_duals[reverse]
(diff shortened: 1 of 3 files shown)


cobra/core/solution.py
```diff
@@ -296,7 +296,11 @@ def get_solution(model, reactions=None, metabolites=None):
         forward = rxn.forward_variable.name
         reverse = rxn.reverse_variable.name
         fluxes[i] = var_primals[forward] - var_primals[reverse]
-        reduced[i] = var_duals[forward] - var_duals[reverse]
+        try:
+            reduced[i] = var_duals[forward] - var_duals[reverse]
+        except TypeError:
+            # reduced costs are not always defined, e.g. for integer problems
+            reduced[i] = nan
```
cdiener: Since the reduced cost is None in that case I think just testing with `if var_duals[forward]:` would be more explicit. Or test with `if var_duals[forward] and var_duals[reverse]:` but that is not necessary...

hredestig: the try/catch is some nanosecond faster if the var_duals are defined :)

cdiener: So better ask for forgiveness than permission? :laughing: I'm fine with it than.

Midnighter:
Now perform the check for `None` outside of the loop and react accordingly. I prefer this way to both other suggestions so far. Also making use of the shiny new `Reaction.reverse_id` ;)

cobra/core/solution.py
```diff
@@ -292,11 +292,17 @@ def get_solution(model, reactions=None, metabolites=None):
...
-        fluxes[i] = var_primals[forward] - var_primals[reverse]
-        reduced[i] = var_duals[forward] - var_duals[reverse]
+    # reduced costs are not always defined, e.g. for integer problems
+    if var_duals[0] is None:
+        reduced.fill(nan)
+        for (i, rxn) in enumerate(reactions):
+            fluxes[i] = var_primals[rxn.id] - var_primals[rxn.reverse_id]
+    else:
+        for (i, rxn) in enumerate(reactions):
+            forward = rxn.id
```
cdiener: That actually gives me a small performance boost across the board :grinning: :+1:

cobra/core/solution.py
```diff
@@ -292,11 +292,17 @@ def get_solution(model, reactions=None, metabolites=None):
     var_primals = model.solver.primal_values
     reduced = zeros(len(reactions))
     var_duals = model.solver.reduced_costs
-    for (i, rxn) in enumerate(reactions):
-        forward = rxn.forward_variable.name
-        reverse = rxn.reverse_variable.name
-        fluxes[i] = var_primals[forward] - var_primals[reverse]
-        reduced[i] = var_duals[forward] - var_duals[reverse]
+    # reduced costs are not always defined, e.g. for integer problems
+    if var_duals[0] is None:
```
cdiener: I don't think that will work since you need to index with a reaction id. There is `six.next(six.itervalues(var_duals))` but it's a bit ugly :thinking:

Midnighter: Doh, fair enough. Will change to `var_duals[reactions[0].id]`.

Issue #442, opened 2017-03-07
"create_stoichiometric_array" is not very intuitive


joaocardoso:
Hi!

I was made aware of `cobra.util.array.create_stoichiometric_array`.
I don't think it is the best naming. The stoichiometric matrix (S), is part of the conceptual representation of the constraint-based model. 

Anyone who learns about these models, learns first about how to build a S matrix and then how to convert it into a linear problem. As a user, I will be looking for the stoichiometric matrix in the documentation. I suggest renaming the function to `create_stoichiometric_matrix`.

Cheers,

João

pstjohn:
We talked about having this function be called by a model property, `model.S`, but this was a controversial issue 😄 . It got removed in #425 i believe. 

Is your problem with the function the fact that it's buried in the cobrapy util namespace? Or that it's called ...`_array` rather than ...`_matrix`?

*Note:* One of the reasons for `array` is that we return a numpy array, (rather than a numpy matrix, which caused some headaches in previous versions). Now that pandas is a hard dependency, it might make sense to return a dataframe, with rows/columns labelled with metabolite id's.

joaocardoso:
Well, it is the name. It actually returns so many stuff (like DataFrame,
some stuff from `sympy`). So, my argument is that the stoichiometric matrix
is a concept, regardless of `numpy` name for the data structure.

hredestig:
closed with #443

PR #443, opened 2017-03-07
Some changes to stoichiometric matrix

pstjohn:
Renames `create_stochiometric_array` to `create_stoichiometric_matrix`, as suggested in #442.

Also, I wasn't paying too close attention to #417, was this really the behavior we want for pandas DataFrames? I would assume that the dataframe returned should be consistent with the other returned values, and have metabolites and rows and reactions as columns.

This PR also implements that change, such that the dataframe return will look like

```
         ACALD  ACALDt  ACKr  ACONTa  ACONTb  ACt2r  ADK1  AKGDH  AKGt2r  \
13dpg_c    0.0     0.0   0.0     0.0     0.0    0.0   0.0    0.0     0.0
2pg_c      0.0     0.0   0.0     0.0     0.0    0.0   0.0    0.0     0.0
3pg_c      0.0     0.0   0.0     0.0     0.0    0.0   0.0    0.0     0.0
6pgc_c     0.0     0.0   0.0     0.0     0.0    0.0   0.0    0.0     0.0
6pgl_c     0.0     0.0   0.0     0.0     0.0    0.0   0.0    0.0     0.0

         ALCD2x ...   RPI  SUCCt2_2  SUCCt3  SUCDi  SUCOAS  TALA  THD2  TKT1  \
13dpg_c     0.0 ...   0.0       0.0     0.0    0.0     0.0   0.0   0.0   0.0
2pg_c       0.0 ...   0.0       0.0     0.0    0.0     0.0   0.0   0.0   0.0
3pg_c       0.0 ...   0.0       0.0     0.0    0.0     0.0   0.0   0.0   0.0
6pgc_c      0.0 ...   0.0       0.0     0.0    0.0     0.0   0.0   0.0   0.0
6pgl_c      0.0 ...   0.0       0.0     0.0    0.0     0.0   0.0   0.0   0.0

         TKT2  TPI
13dpg_c   0.0  0.0
2pg_c     0.0  0.0
3pg_c     0.0  0.0
6pgc_c    0.0  0.0
6pgl_c    0.0  0.0

[5 rows x 95 columns]
```

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -687,7 +687,7 @@ class Model(Object):
         which may be used to perform linear algebra operations with the
         stoichiometric matrix.
 
-        Deprecated (0.6). Use `cobra.util.array.create_stoichiometric_array`
+        Deprecated (0.6). Use `cobra.util.array.create_stoichiometric_matrix`
         instead.
 
         Parameters
@@ -697,7 +697,7 @@ class Model(Object):
 
         """
         warn("to_array_based_model is deprecated. "
-             "use cobra.util.array.create_stoichiometric_array instead",
+             "use cobra.util.array.create_stoichiometric_matrix instead",
              DeprecationWarning)
         from cobra.core.arraybasedmodel import ArrayBasedModel
         return ArrayBasedModel(self, deepcopy_model=deepcopy_model, **kwargs)
--- a/cobra/flux_analysis/loopless.py
+++ b/cobra/flux_analysis/loopless.py
@@ -10,7 +10,7 @@ from sympy.core.singleton import S
 
 from cobra.core import Metabolite, Reaction, get_solution
 from cobra.util import (linear_reaction_coefficients,
-                        create_stoichiometric_array)
+                        create_stoichiometric_matrix)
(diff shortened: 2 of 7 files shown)


hredestig:
My reason for multiindex was that for plotting, merging etc with other data frames it is more convenient to have the data in 'long' form and I thought that would be the main use case of having a data frame rather than an array, for which the 2d makes more sense for mathematical operations. 2d data frames seems a bit redundant,  but, it's easy enough to `pandas.melt` so if you think that format is better then fine with me.

pstjohn:
That's fair, I could be convinced either way. Other opinions?

I assumed the dataframe's main use would be to keep track of the physical meaning of the rows and columns of the matrix, especially after slicing/reordering.

pstjohn:
I was thinking mainly for [conservation analysis](dx.doi.org/10.1016/j.bpc.2003.08.009), since you end up reshuffling and dropping some rows of the matrix. But for most applications I'd probably use the array as well.

Good call on the release notes, not sure why my `git grep` didn't pick up on that... must have run it from inside the cobra folder.

PR #446, opened 2017-03-08
Fix loopless FVA

cdiener:
Fixes a bug with loopless FVA that made it impossible to remove loops. Now correctly detects the FRD7- SUCDi loop in the textbook model (test changed as well). Also refactored to make the function simpler. Seems to be about 10-20% faster now as well.


cobra/flux_analysis/loopless.py
```diff
@@ -82,6 +81,26 @@ def add_loopless(model, zero_cutoff=1e-12):
         model.constraints[name].set_linear_coefficients(coefs)
 
 
+def __add_cycle_free(model, fluxes):
```
hredestig: a reason for the double under instead of just single?

cobra/flux_analysis/loopless.py
```diff
@@ -200,57 +201,52 @@ def loopless_fva_iter(model, reaction, all_fluxes=False, zero_cutoff=1e-9):
...
-    else:
-        best = reaction.flux
+            rid = rxn.id
+            if ((abs(ll_sol[rid]) < zero_cutoff) and
+                    (abs(almost_ll_sol[rid]) > zero_cutoff)):
+                rxn.bounds = max(0, rxn.lower_bound), min(0, rxn.upper_bound)
+
+        if solution:
+            solution = model.optimize(objective_sense=None)
+            best = solution.fluxes
```
hredestig: when I try, I never reach this statement but seems if I did I wouldn't get the requested whole solution?

cobra/test/test_flux_analysis.py
```diff
@@ -259,14 +259,15 @@ def test_flux_variability(self, model, fva_results, solver):
...
-            for k, v in iteritems(result):
-                assert abs(fva_results[k][name] - v) < 0.00001
+        loop_reactions = [model.reactions.get_by_id(rid)
+                          for rid in ("FRD7", "SUCDi")]
+        fva_normal = flux_variability_analysis(
+            model, solver=solver, reaction_list=loop_reactions)
+        fva_loopless = flux_variability_analysis(
+            model, solver=solver, reaction_list=loop_reactions, loopless=True)
+
+        assert not numpy.allclose(fva_normal["maximum"], fva_normal["minimum"])
```
hredestig: nice, much better test, let's add this as example in the loopless.ipynb as well

hredestig: I just updated the loopless notebook with this, pardon the cluttered commit, took the opportunity to just re-run the other notebooks as well.

cdiener: Sure! I think we should not advertise add_loopless to much. For almost all use cases it is the wrong choice. Will try to rewrite a bit...

cobra/flux_analysis/loopless.py
cdiener: You are right. Will fix.

cobra/flux_analysis/loopless.py
cdiener: Not really. Hoped that this would obfuscate the name as for class members so it wont be used accidentally by the user. will change it to one.

cdiener:
Hi @hredestig. Thanks for the review. The PR was a bit hurried so it is appreciated. Also saw that I introduced a change to `fix_objective_as_constraint` (name is now an arg) because I thought I needed it but could not use it in the end. If you want to revert that just say the word.

PR #455, opened 2017-03-13
fix: missing link in docs / libsbml

hredestig:
- missing link to notebook
- use own python-libsbml index to get mac wheels


hredestig:
yay ci for mac down from ~20 min to 10

PR #457, opened 2017-03-14
bump version

matthiaskoenig:
This version should be in sync with version in `__init__.py` and `setup.py`. 
Is this file needed? For what is it used?


hredestig:
It's actually just lingering lint, not used for anything anymore. If you delete it I'm happy to merge that.

PR #460, opened 2017-03-15
Fix/blocked reactions

hredestig:


cobra/flux_analysis/variability.py
```diff
@@ -197,33 +197,56 @@ def _fva_optlang(model, reaction_list, fraction, loopless):
...
+        # limit to reactions which are already 0. If the reactions already
+        # carry flux in this solution, then they can not be blocked.
+        if legacy:
+            solution = solver_interface.solve(model, **solver_args)
+            reaction_list = [i for i in reaction_list
+                             if abs(solution.x_dict[i.id]) < zero_cutoff]
+        else:
+            model.solver = solver
+            model.solver.optimize()
+            reaction_list = [rxn for rxn in reaction_list if
```
cdiener: Since you're iterating over all reactions this might actually be faster with Solution. But I guess you tried it already :)

cobra/flux_analysis/variability.py
```diff
@@ -197,33 +197,56 @@ def _fva_optlang(model, reaction_list, fraction, loopless):
...
+                             abs(rxn.flux) < zero_cutoff]
+        # run fva to find reactions where both max and min are 0
+        flux_span = flux_variability_analysis(
+            model, fraction_of_optimum=0., reaction_list=reaction_list,
+            solver=solver, **solver_args)
+        return [k for k, v in iteritems(flux_span.T) if
```
cdiener: pandas.DataFrame has an iterrows function so you dont need to transpose

cobra/flux_analysis/variability.py
hredestig: You're right, using the solution is cleaner, I was being lazy mea culpa :) Thanks!

cobra/flux_analysis/variability.py
hredestig: Yup, fixed, also my laziness

PR #461, opened 2017-03-21
fix: enable pickling an model with context

hredestig:
contexts cannot (and arguably shouldn't) be saved, but one may want to
save a model object that has a context.

(currently trying to pickle a model with an active context will result in uncaught exception)

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -54,8 +54,11 @@ class Model(Object):
     """
 
     def __setstate__(self, state):
-        """Make sure all cobra.Objects in the model point to the model"""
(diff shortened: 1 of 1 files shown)


hredestig:
Good intentions but not the right solution, reading a bit more realise this needs to be done differently. I'll be back...

pstjohn:
Interesting, I guess I'm not positive that you *wouldn't* want the temporary changes to be saved. Why else would you be pickling inside a with block? Loading the model, there's no way to signal an 'exit' to the context, so I don't think you would want the context history saved even if we could save it. Probably a niche application in either direction.

PR #466, opened 2017-03-24
fix: add metabolites in nested contexts

hredestig:
With model as context, `reaction.add_metabolites` would make an
addition of a partial call to `reaction.subtract_metabolites`, which in
turn called `reaction.add_metabolites` thus adding the revert itself back
to the history stack. Avoid this by adding a flag to avoid this addition
from happening. Shame on the ugly extra parameter but saw no good alternatives as the `reaction.add_metabolites` function is quite complex.

Also add small performance boost to make
`model.add_metabolites` on empty list a noop (can happen fairly often when
manipulating many reactions).

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -331,6 +331,9 @@ class Model(Object):
         """
         if not hasattr(metabolite_list, '__iter__'):
             metabolite_list = [metabolite_list]
+        if len(metabolite_list) == 0:
+            return None
+
         # First check whether the metabolites exist in the model
         metabolite_list = [x for x in metabolite_list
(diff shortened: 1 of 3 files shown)


pstjohn:
Yikes, thanks for catching that. Sorry for the bugs :)
Those add/remove metabolite and reaction methods are a definitely pain point.

Whether or not its possible to make the connections between reactions, metabolites, and models simpler is definitely something to keep in mind for future refactorings

hredestig:
No problem at all, yes definitely, it would be great to come up with a structure that wasn't so densely interconnected..

PR #471, opened 2017-03-27
feat: expose _{to,from}_dict

hredestig:
Useful functions outside cobrapy so give public names.


pstjohn:
I use these in d3flux as well, thumbs up!

Midnighter:
I used the dicts for making a quick YAML export as well. Good idea to expose them.

PR #474, opened 2017-03-29
feat: assert solver status optimal

hredestig:
The pattern of checking if solver.status != 'optimal' is not and then
raise appropriate exception started to appear in a number of places -
let's have a function for that.

Also tidy up the use of optlang string constants.

### DIFF
--- a/cobra/util/solver.py
+++ b/cobra/util/solver.py
@@ -20,7 +20,7 @@ from warnings import warn
 import optlang
 import sympy
 
-from cobra.exceptions import OptimizationError
+from cobra.exceptions import OptimizationError, OPTLANG_TO_EXCEPTIONS_DICT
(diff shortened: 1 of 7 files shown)


cobra/util/solver.py
```diff
@@ -20,7 +20,7 @@
 import optlang
 import sympy
 
-from cobra.exceptions import OptimizationError
+from cobra.exceptions import OptimizationError, _OPTLANG_TO_EXCEPTIONS_DICT
```
Midnighter: I'd feel better if `_OPTLANG_TO_EXCEPTIONS_DICT` was not marked as private or exposed to public in some other way.

hredestig: good one, copy-paste mistake, fixed

PR #479, opened 2017-04-06
feat: add_exchange reaction

hredestig:
Add convenient function for rapidly creating an exchange reaction for a
given metabolite

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -15,7 +15,7 @@ from sympy import S
 
 from cobra.core.dictlist import DictList
 from cobra.core.object import Object
-from cobra.core.reaction import separate_forward_and_reverse_bounds
+from cobra.core.reaction import separate_forward_and_reverse_bounds, Reaction
 from cobra.core.solution import get_solution
 from cobra.solvers import optimize
 from cobra.util.context import HistoryManager, resettable, get_context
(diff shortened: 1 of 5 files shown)


matthiaskoenig:
Just some comments:

[1] The arbitrary bound 1000 is used in many positions. Probably a good idea to define a constant and reuse it so it is consistant throughout the code base. And also to allow people to use their own arbitrary bound. I.e. define once DEFAULT_BOUND=1000 and reuse it (also in model importers and reaction constructor for instance).
```
bound=1000.0
```

[2] Their is not only demand and sink. Often also exchange, i.e. (-bound, bound) with flux in both direction.

[3] Why not use the `EX_` prefix which is used throughout all BiGG models and somehow de facto standard to name exchange reactions.

hredestig:
Cheers for the comments! 
Your first point is valid although not really related to this PR, feel free to raise an issue on that topic if you like. 

The second point as well of course however, this is convenience function (coming from cameo) and making it more flexible somehow refutes the purpose a bit I feel as you can just create your own special exchange reaction directly.. 

The third point. Taken, fixed!

Midnighter:
So [1] talks of the following artificial reactions: exchange, sink, and demand reactions.

> **Exchange reactions:** These reactions are unbalanced, extra-organism reactions that represent the supply to or removal of metabolites from the extra-organism “space”.

> **Sink reaction:** When the synthesis reaction(s) of a metabolite is not known or outside the scope of the reconstruction, its discharge can be represented by this unbalanced, intracellular reaction (e.g., 1 A < -->).

> **Demand reaction:** When the consumption reaction(s) of a metabolite is not known or outside the scope of the reconstruction it can be represented by this unbalanced, intracellular reaction (e.g., 1 A -->).

So exchanges are extracellular and reversible, sinks are intracellular and reversible, and demands are intracellular and irreversible.

If we look at http://bigg.ucsd.edu/ the typical identifiers are: EX_<metabolite ID>, SK_<metabolite ID>, and DM_<metabolite ID>. I think we should stick to their definitions for the sake of standardization.

[1] Thiele, Ines, and Bernhard Ø Palsson. “A Protocol for Generating a High-Quality Genome-Scale Metabolic Reconstruction.” Nature Protocols 5, no. 1 (January 2010): 93–121. doi:10.1038/nprot.2009.203.

cobra/util/util.py
```diff
@@ -57,3 +59,38 @@ def __getitem__(self, item):
         except KeyError:
             value = self[item] = type(self)()
             return value
+
+
+def add_exchange(model, metabolite, demand=True, prefix='EX_', bound=1000.0):
```
cdiener: As mentioned by @midnighter exchange reaction can also export metabolites or both (import and export at the same time). I don't see how the function makes that possible. 

I also agree with the standard prefixes of EX_, DM_, etc. I actually have packages that break if you don't use them...

pstjohn:
Just a minor comment, is `util/util.py` the best place for this? I'd think it would want to be in something like `manipulation`?
Currently the stuff in util/util are random helper functions and classes, certainly nothing a user would want to call outside the package.

cobra/util/util.py
hredestig: Alright, good point then, adjusted

hredestig:
@pstjohn users only interact with `cobra.util` which imports it all but I wasn't really sure where to put it as didn't want to make even more modules in `manipulation` (none there seemed to fit..) but perhaps it's better -> moved

cobra/manipulation/add.py
```diff
@@ -0,0 +1,48 @@
...
+        'exchange', 'DM_' for demand and 'SK_' for sink reactions.
+
+    Returns
+    -------
+    cobra.Reaction
+        The created exchange reaction.
+
+    """
+    exchange_types = {'demand': ('DM_', (0, bound)),
+                      'sink': ('SK_', (-bound, 0)),
```
cdiener: A sink is not the same as an export. Sink means the metabolite "disappears" inside of the compartment (for instance if there are unknown consuming reactions). An export means the metabolite is transported to the outside across the compartment boundary. Both look the same in stoichiometry but only the export would be considered in some algorithms (for instance modeling exchanges between several species), whereas the sink would not. So we should give a way to distinguish export and sink reactions by the prefix ("SK_" or "EX_"). 

My proposal: add "import" and "export" specifiers to the `exchange_types` and handle their prefix accordingly.

Thanks!

cobra/core/model.py
```diff
@@ -418,6 +418,72 @@ def add_reaction(self, reaction):
...
+        >>> demand.id
+        'DM_atp_c'
+        >>> demand.name
+        'ATP demand'
+        >>> demand.bounds
+        (0, 1000.0)
+        >>> demand.build_reaction_string()
+        'atp_c --> '
+        """
+        types = dict(exchange=("EX", -ub, ub), demand=("DM", 0, ub),
```
hredestig: should it be -lb, ub for 'EX'?

cobra/core/model.py
```diff
@@ -418,6 +418,72 @@ def add_reaction(self, reaction):
...
+        exchange, demand, and sink reactions.
+        An exchange reaction is a reversible, imbalanced reaction that adds
+        to or removes an extracellular metabolite from the extracellular
+        compartment.
+        A demand reaction is an irreversible reaction that consumes an
+        intracellular metabolite.
+        A sink is similar to an exchange but specifically for intracellular
+        metabolites.
+
+        If you specify a type other than one of these three, you can
```
hredestig: 'If you set the reaction `type` to something else, you must specify the desired identifier of the created reaction along with its upper and lower bound.'

cobra/core/model.py
```diff
@@ -418,6 +418,72 @@ def add_reaction(self, reaction):
...
+
+        Parameters
+        ----------
+        metabolite : cobra.Metabolite
+            Any given metabolite. The compartment is not checked but you are
+            encouraged to stick to the definition of exchanges and sinks.
+        type : str, {"exchange", "demand", "sink"}
+            Using one of the pre-defined reaction types is easiest. If you
+            want to create your own kind of boundary reaction choose
+            something else.
```
hredestig: ...choose any other string, e.g. 'my-boundary'.

Midnighter:
@cdiener Henning and I are both happy with the current state. Can you give the final approval, please?

PR #481, opened 2017-04-06
Clean up the `get_solution` function using new optlang features

Midnighter:
`optlang.Model` now has an `is_integer` attribute which we can use to greatly improve the `get_solution`.

### DIFF
--- a/cobra/core/solution.py
+++ b/cobra/core/solution.py
@@ -8,7 +8,7 @@ import logging
 from builtins import object, super
 from warnings import warn
 
-from numpy import zeros, asarray, nan
+from numpy import empty, nan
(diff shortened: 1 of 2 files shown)


Midnighter:
Yeah, I commented on that in the commit. Forgot to replicate that here. This avoids looping through the reactions/metabolites multiple times. Still, although I'm not opposed to speed-ups it's pretty clear to me that `get_solution` is a convenience function and there's only so much to be done about speed.

cdiener:
Looks good. I agree. For speed I would favor adding helper functions to `util/solver`. For instance a functions just getting the fluxes in the fastest way possible for users who need that.

PR #484, opened 2017-04-07
port sampling to optlang

### LINKED ISSUE #415: Port flux sampling to optlang
So that one is a bit more complex. Basically need to implement the following:

- [x] port eveything to the iireversible form used in model.solver (performance?)
- [x] Port the warmup points generation to FVA
- [ ] Add an option for loopless warmup points to sample only from the loopless sub-space
- [x] Respect additional equality and inequality constraints declared in optlang interface (hardest one)

Things that I may *not* want to implement yet and why (all of them open for discussion of course):

- respect integer variables (can not be done with HR samplers)
- include non-reaction variables in returned flux samples (with pure numpy hard to see which value  corresponds to what)

This will require construction of a complete equality constraint matrix [A -b] for all optlang variables v so that Av - b = 0 and inequality constraints matrix B with lb >= Bv >= ub. I though about putting functions for those in `cobra.util.array` to not fill up `cobra.flux_analysis.sampling` too much.

cdiener:
Sorry this took so long. I encountered some difficulties getting the sampling to respect variable/constraint equalities/inequalities (particularly equalities are not trivial). The sampler now internally uses the irreversible formulation as in Model.solver. This makes the sampling a little bit slower however I could rescue most of that by upping the reprojection frequency. As a consequence optGP is more or less as fast as before for smaller models and faster for large models. ARCH is a little bit slower for small models, but faster for larger ones. Using non-zero equality constraints requires the solve to reproject in the feasible plane in each iteration which increases the running time by approximately 10x.

Initialization time has increased by a factor of 2-3 which is due to the fact that obtaining primal values for a solved model is slower in optlang than with the old cobrapy solvers. Recent fixes in optlang might improve on that.

`sample` and `batch` can now return the samples by fluxes or variables. `validate` can detect which one was used and adapt accordingly.

[Convergence estimates and benchmarks](https://github.com/cdiener/cobra-docker/blob/master/sampling_benchmark.ipynb)

## To consider
- fine tune the reprojection frequency in `NPROJ` a bit more since it gives significant speed-up when higher, basically try with many model and lost of iterations if for any given `NPROJ` there will be invalid samples generated. Can be left for future releases.

The PR fixes #415.


hredestig:
This looks really great to me, thanks for all your work with it! Of course needs a proper review, as not yet well versed with flux sampling, I'm taking the opportunity to read the literature but will take a few more days. In the mean time, if anyone else feels qualified, please review this.

cdiener:
Ah sure, no worries. I think the parts that require review the most are not even that related to sampling, for instance:

- the `constraint_matrices` function and its return type
- the attributes in HRSampler.__init__ -> lot's of public ones, but sometimes it's nice to have easy access to those for debugging purposes
- and in general just running it with some of you fav models and see whether the produced samples remain valid with the `sampler.validate` function

cobra/flux_analysis/sampling.py
```diff
@@ -84,14 +98,37 @@ class HRSampler(object):
...
+        # required to get deterministic warmup point generation
+        # (in turn required for a working `seed` arg)
+        if model.solver.is_integer:
+            raise TypeError("sampling does not work with integer problems :(")
+        self.model = model.copy()
         self.thinning = thinning
         self.n_samples = 0
-        self.S = create_stoichiometric_matrix(model, array_type='dense')
+        prob = constraint_matrices(model)
+        self.S = prob.equalities
```
hredestig: I think it would help readability to avoid the single letter attributes (S, NS, M, b etc, also BTOL, FTOL) in favor of complete names (and variables lower-case)

cobra/flux_analysis/sampling.py
```diff
@@ -113,33 +150,29 @@ def generate_fva_warmup(self, solver=None, **solver_args):
...
+        self.model.objective = S.Zero
+        self.model.objective.direction = "max"
+        variables = self.model.variables
+        for i in idx:
+            # Omit fixed reactions
+            if self.fixed[i]:
+                pass
+            self.model.objective.set_linear_coefficients({variables[i]: 1})
+            self.model.solver.optimize()
+            if self.model.solver.status != "optimal":
```
hredestig: could use the `assert_optimal` function and avoid the string literal

cobra/flux_analysis/sampling.py
```diff
@@ -113,33 +150,29 @@ def generate_fva_warmup(self, solver=None, **solver_args):
...
+                pass
+            primals = self.model.solver.primal_values
```
hredestig: this throws AssertionError for me for iMM904 and yeast7, would help to catch those (although I guess need to throw away the whole sample then :/ )..

cobra/flux_analysis/sampling.py
```diff
@@ -496,8 +592,7 @@ def __getstate__(self):
         return d
 
 
-def sample(model, n, method="optgp", processes=1, seed=None,
-           solver=None, **solver_kwargs):
+def sample(model, n, method="optgp", processes=1, seed=None):
```
hredestig: for larger models, sampling obviously takes quite a while, I think a progress bar would help.. tqdm is such a nice package perhaps it can be allowed to be a hard requirement even, what do you think? Can also be left for a different pr of course.

cobra/test/test_flux_analysis.py
```diff
@@ -542,22 +542,40 @@ def test_wrong_method(self, model):
...
-        assert numpy.allclose(s.TPI[0], [8.38570846])
+        assert numpy.allclose(s.TPI[0], 5.6918488269)
+
+    def test_equality_constraint(self, model):
+        model.reactions.ACALD.bounds = (-1.5, -1.5)
+        s = sample(model, 10)
+        assert numpy.allclose(s.ACALD, -1.5, atol=1e-6, rtol=0)
+
+    def test_inequality_constraint(self, model):
+        co = model.problem.Constraint(
```
hredestig: very nice this

cobra/util/array.py
```diff
@@ -109,3 +110,116 @@ def nullspace(A, atol=1e-13, rtol=0):
...
+    -------
+    collections.namedtuple
+        A named tuple consisting of 6 matrices and 2 vectors:
+        - "equalities" is a matrix S such that S*vars = b. It includes a row
+          for each constraint and one column for each variable.
+        - "b" the right side of the equality equation such that S*vars = b.
+        - "inequalities" is a matrix M such that lb <= M*vars <= ub.
+          It contains a row for each inequality and as many columns as
+          variables.
+        - "bounds" is a compound matrix [lb ub] constaining the lower and upper
```
hredestig: some spelling typos 'constaining'

cobra/flux_analysis/sampling.py
cdiener: Uhh, that is a bug. Should be continue instead of pass.

cobra/flux_analysis/sampling.py
cdiener: I agree. Do you think I should encapsulate all the matrices into a struct like object (namedtuple for instance)? So you would have something like sampler.problem.stoichiometry, sampler.problem.constraints, etc... Or rather leave it as sampler.stoichiometry?

cobra/flux_analysis/sampling.py
cdiener: Will do.

cobra/flux_analysis/sampling.py
cdiener: Yeah, major run time actually comes from getting the initial search directions. Thought about alternatives but it's difficult since the initial vectors need to span all dimensions. It comes back to the known conundrum of getting the fluxes fast which is the major bottleneck here. tqdm could be nice but in that case we would also want it for other functions such as double deletions, flux variability etc...

hredestig: Yes let's handle it in a different pr

cobra/flux_analysis/sampling.py
hredestig: that sounds good, would gather all those related attributes to a single unit. But no strong opinion there.

cdiener:
Ok, have a look. The error you see with iMM904 and yeast7 is a bug in optlang. Before returning the primal values optlang rounds them to the bounds and does a feasibility check. Unfortunately, optlang uses a fixed chosen tolerance for that check that does not coincide with the one the solver used. In the those two models some upper bounds are very large (~1e6) and the glpk uses a relative tolerance thus allowing pretty large absolute deviations in that case (because rtol=1e-9 and 1e-9 * 1e6 = 1e-3). optlang does not respect that and applies its own absolute tolerance of 1e-5 which is violated.

cobra/flux_analysis/sampling.py
```diff
@@ -14,27 +14,47 @@
...
+from sympy.core.singleton import S
+from cobra.util import (create_stoichiometric_matrix, constraint_matrices,
+                        nullspace, assert_optimal)
 
-BTOL = np.finfo(np.float32).eps
+bounds_tol = np.finfo(np.float32).eps
 """The tolerance used for checking bounds feasibility."""
 
-FTOL = BTOL
+feasibility_tol = bounds_tol
```
hredestig: My apologies for being unclear, with lowercase vars I meant local variables not the globals which are fine as all caps.

cobra/flux_analysis/sampling.py
```diff
@@ -14,27 +14,47 @@
...
 
+nproj = 100000
+"""Reproject the solution into the feasibility space every NPROJ iterations."""
+
 bad_x = None
 bad_delta = None
 bad_alpha = None
 
 
+class Problem:
```
cdiener: Yeah, actually I just wanted the namedtuple from `constraint_matrices` but with one additional attribute. Initially used SimpleNamespace but that is not supported by Python 2. Will come up with a better solution...

cobra/flux_analysis/sampling.py
```diff
@@ -113,55 +186,60 @@ def generate_fva_warmup(self, solver=None, **solver_args):
...
-                solver.change_variable_objective(lp, i, 0.)
+        idx = np.hstack([self.fwd_idx, self.rev_idx])
+        self.warmup = np.zeros((len(idx), len(self.model.variables)))
+        self.model.objective = S.Zero
+        self.model.objective.direction = "max"
+        variables = self.model.variables
+        for i in idx:
+            # Omit fixed reactions
+            if self.problem.fixed[i]:
+                pass
```
cdiener: Yes, definitely continue :sweat_smile:

cobra/flux_analysis/sampling.py
```diff
@@ -113,55 +186,60 @@ def generate_fva_warmup(self, solver=None, **solver_args):
...
+            self.model.objective.set_linear_coefficients({variables[i]: 1})
+            self.model.solver.optimize()
+            try:
+                assert_optimal(self.model)
+            except Exception:
```
cdiener: Ok, wasn't sure what assert_optimal could raise. Maybe we should add that to the docstring.

cobra/flux_analysis/sampling.py
```diff
@@ -113,55 +186,60 @@ def generate_fva_warmup(self, solver=None, **solver_args):
...
+        if prob.bounds.shape[0] > 0:
+            const = prob.inequalities.dot(p)
+            const_lb_dist = (const - prob.bounds[0, ]).min()
+            const_ub_dist = (prob.bounds[1, ] - const).min()
+            lb_dist = min(lb_dist, const_lb_dist)
+            ub_dist = min(ub_dist, const_ub_dist)
         return np.array([lb_dist, ub_dist])
 
-    def sample(self, n):
+    def sample(self, n, fluxes=True):
```
cdiener: In theory yes, only problem is that if you change it continuously you might not know anymore with which thinning earlier samplers were generated. I right now it is also possible to do that, just reassign sampler.thinning (it is not protected).

cobra/flux_analysis/sampling.py
```diff
@@ -487,18 +654,26 @@ def sample(self, n):
...
+        """Return the object for serialization."""
         d = dict(self.__dict__)
         del d['model']
         return d
 
 
-def sample(model, n, method="optgp", processes=1, seed=None,
-           solver=None, **solver_kwargs):
-    """Samples valid flux distribution from a cobra model.
+def sample(model, n, method="optgp", processes=1, seed=None):
```
cdiener: That sounds like a good idea.

cdiener:
Ok will check what is going on with iMM904. Looking at your analysis it seems that a lot of fluxes are fixed/near constant so that might be the problem. Will investigate further.

cobra/flux_analysis/sampling.py
hredestig: Good point, sorry for the bad doc there 😅

cobra/flux_analysis/sampling.py
hredestig: Alright, good point

cdiener:
Okay, added several changes. for `iMM904` it was indeed a fixed reaction (ATPM flux is set to be = 1), so sampling does not make too much sense for that model since the fluxes are essentially all fixed. However, those non-homogeneous cases now run much faster and can be sampled pretty much with the same speed as the homogeneous cases (maybe 10% slower).

`sampler.problem` now is a named tuple. Also refactored problem creation into its own private method.

`sample` now takes a `thinning` argument.

ARCH is now abbreviated as ACHR which is the common denomination in the literature even though it's impossible to pronounce.

cobra/flux_analysis/sampling.py
```diff
@@ -487,29 +651,37 @@ def sample(self, n):
...
+        """Return the object for serialization."""
         d = dict(self.__dict__)
         del d['model']
         return d
 
 
-def sample(model, n, method="optgp", processes=1, seed=None,
-           solver=None, **solver_kwargs):
-    """Samples valid flux distribution from a cobra model.
+def sample(model, n, method="optgp", thinning=100, processes=1, seed=None):
```
hredestig: Guess you mean to also pass this parameter to the constructors?

hredestig:
see you're still putting in some changes, the notebook should then also be updated with the new ACHR name

cobra/flux_analysis/sampling.py
cdiener: Ah yeah good catch.

cdiener:
Ok, the requested changes.

hredestig:
Looks great and works well in my small tests, cheers for all the hard work!

cdiener:
Good to hear that :)

PR #493, opened 2017-04-25
feat: add to_frame method to solution

hredestig:


cobra/core/solution.py
```diff
@@ -169,6 +169,11 @@ def y(self):
         warn("use solution.reduced_costs.values() instead", DeprecationWarning)
         return self.reduced_costs.values
 
+    def to_frame(self):
+        """Return the fluxes and reduced costs as a data frame"""
+        return DataFrame({'fluxes': self.fluxes,
```
Midnighter: Is the index of the `DataFrame` automatically taken to be the index of the fluxes and reduced costs, i.e., the reaction IDs? If not I would add that explicitly.

hredestig: It is, since that's the index of those two series

PR #495, opened 2017-04-26
make test_all work with Python 2.7

Midnighter:
Previously, the `small_model` fixture was not found.

### DIFF
--- a/cobra/test/__init__.py
+++ b/cobra/test/__init__.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+
 from os.path import abspath, dirname, join
 
 from cobra.io import read_sbml_model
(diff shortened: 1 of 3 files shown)


cobra/test/__init__.py
```diff
@@ -50,7 +53,7 @@ def test_all(args=None):
 
         return pytest.main(
             ['--pyargs', 'cobra', '--benchmark-skip', '-v', '-rs'] + args
-        ) != 0
+        )
```
hredestig: what is now the return value? do the [install instructions](https://github.com/opencobra/cobrapy/blob/devel/INSTALL.rst#testing-your-installation) need to be updated?

Midnighter: Ah, probably easiest to revert that. It would return the code that `pytest.main` gives. So `0` if everything passes as expected and `1` otherwise. Would `test_all` evaluating to `True` be more logical when all tests pass?

PR #496, opened 2017-04-28
Add YAML round tripping support

Midnighter:
In this PR:
* Conversion of a model to/from a `dict` is given its own `cobra.io` module and the order of elements is maintained using `OrderedDict`s.
* JSON uses the new dictionaries without problems.
* Models can be round tripped to/from YAML using ruamel.yaml.


cobra/io/dict.py
```diff
@@ -0,0 +1,224 @@
...
+    obj["reactions"] = [reaction_to_dict(reaction)
+                        for reaction in model.reactions]
+    obj["metabolites"] = [metabolite_to_dict(metabolite)
+                          for metabolite in model.metabolites]
+    obj["genes"] = [gene_to_dict(gene)
+                    for gene in model.genes]
+    obj["id"] = model.id
+    _update_optional(model, obj, _OPTIONAL_MODEL_ATTRIBUTES,
+                     _ORDERED_OPTIONAL_MODEL_KEYS)
+    # add in the YAML version
```
hredestig: what's this, should it be added in `to_yaml` instead?

PR #497, opened 2017-05-01
feat: gapfilling with optlang (fix #383)

### LINKED ISSUE #383: refactor gapfilling / update  hints

hredestig:
Suggested re-factor of the gapfilling algorithms. Is the api okay? The reactions are no longer split in forward and reverse, does that make it less useful?

~~Also, sadly this implementation still has a serious bug that I searched too long for and would be happy for hints: the old tests pass but the gapfilling.ipynb doesn't since for some reason the solver ignores the switch constraints allowing flux through reactions that aren't switched on.~~

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -316,6 +316,10 @@ class Model(Object):
         except Exception:  # pragma: no cover
             new._solver = copy(self.solver)  # pragma: no cover
 
+        # it doesn't make sense to retain the context of a copied model so
+        # assign a new empty context
+        new._contexts = list()
+
         return new
 
     def add_metabolites(self, metabolite_list):
--- a/cobra/flux_analysis/gapfilling.py
+++ b/cobra/flux_analysis/gapfilling.py
@@ -2,8 +2,338 @@
 
 from __future__ import absolute_import
 
-from cobra.exceptions import DefunctError
+from sympy import Add
+from warnings import warn
+
+from optlang.interface import OPTIMAL
+from cobra.core import Model
+from cobra.util import assert_optimal, fix_objective_as_constraint
 
 
-def growMatch():
-    raise DefunctError('growMatch', 'modelseed', 'http://modelseed.org/')
+class GapFiller(object):
+    """Class for performing gap filling.
+
+    This class implements gap filling based on a mixed-integer approach,
+    very similar to that described in [1]_ and the 'no-growth but growth'
+    part of [2]_ but with minor adjustments. In short, we add indicator
+    variables for using the reactions in the universal model, z_i and then
+    solve problem
+
+    minimize \sum_i c_i * z_i
+    s.t. Sv = 0
+         v_o >= t
+         lb_i <= v_i <= ub_i
+         v_i = 0 if z_i = 0
+
+    where lb, ub are the upper, lower flux bounds for reaction i, c_i is a
+    cost parameter and the objective v_o is greater than the lower bound t.
+    The default costs are 1 for reactions from the universal model, 100 for
+    exchange (uptake) reactions added and 1 for added demand reactions.
+
+    Note that this is a mixed-integer linear program and as such will
+    expensive to solve for large models. Consider using alternatives [3]_
+    such as CORDA instead [4,5]_.
+
+    Parameters
+    ----------
+    model : cobra.Model
+        The model to perform gap filling on.
+    universal : cobra.Model
+        A universal model with reactions that can be used to complete the
+        model.
+    lower_bound : float
+        The minimally accepted flux for the objective in the filled model.
+    penalties : dict, None
+        A dictionary with keys being 'universal' (all reactions included in
+        the universal model), 'exchange' and 'demand' (all additionally
+        added exchange and demand reactions) for the three reaction types.
+        Can also have reaction identifiers for reaction specific costs.
+        Defaults are 1, 100 and 1 respectively.
+    integer_threshold : float
+        The threshold at which a value is considered non-zero (aka
+        integrality threshold). If gapfilled models fail to validate,
+        you may want to lower this value.
+    exchange_reactions : bool
+        Consider adding exchange (uptake) reactions for all metabolites
+        in the model.
+    demand_reactions : bool
+        Consider adding demand reactions for all metabolites.
+
+    References
+    ----------
+    .. [1] Reed, Jennifer L., Trina R. Patel, Keri H. Chen, Andrew R. Joyce,
+       Margaret K. Applebee, Christopher D. Herring, Olivia T. Bui, Eric M.
+       Knight, Stephen S. Fong, and Bernhard O. Palsson. “Systems Approach
+       to Refining Genome Annotation.” Proceedings of the National Academy
+       of Sciences 103, no. 46 (2006): 17480–17484.
+
+       [2] Kumar, Vinay Satish, and Costas D. Maranas. “GrowMatch: An
+       Automated Method for Reconciling In Silico/In Vivo Growth
+       Predictions.” Edited by Christos A. Ouzounis. PLoS Computational
+       Biology 5, no. 3 (March 13, 2009): e1000308.
+       doi:10.1371/journal.pcbi.1000308.
+
+       [3] http://opencobra.github.io/cobrapy/tags/gapfilling/
+
+       [4] Schultz, André, and Amina A. Qutub. “Reconstruction of
+       Tissue-Specific Metabolic Networks Using CORDA.” Edited by Costas D.
+       Maranas. PLOS Computational Biology 12, no. 3 (March 4, 2016):
+       e1004808. doi:10.1371/journal.pcbi.1004808.
+
+       [5] Diener, Christian https://github.com/cdiener/corda
+     """
+
+    def __init__(self, model, universal=None, lower_bound=0.05,
+                 penalties=None, exchange_reactions=False,
+                 demand_reactions=True, integer_threshold=1e-6):
+        self.original_model = model
+        self.lower_bound = lower_bound
+        self.model = model.copy()
+        # TODO: adjust once optlang supports integrality constraint settings
+        try:
+            self.model.solver.configuration._iocp.tol_int = integer_threshold
+        except AttributeError:
+            try:
+                self.model.solver.problem.parameters.mip.tolerances. \
+                    integrality.set(integer_threshold)
+            except AttributeError:
+                warn("tried to set integrality constraint, but don't know "
+                     "how to do that for "
+                     "solver {}".format(self.model.problem.__name__))
+        self.universal = universal.copy() if universal else Model('universal')
+        self.penalties = dict(universal=1, exchange=100, demand=1)
+        if penalties is not None:
+            self.penalties.update(penalties)
+        self.integer_threshold = integer_threshold
+        self.indicators = list()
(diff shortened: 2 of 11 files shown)


gregmedlock:
Thanks for pushing this forward! I think it makes sense to leave reaction-splitting up to the user if they need it.

I _think_ I ran into issues similar to this with the old gapfilling functions; if the issue is the same, then the indicator constraints that have non-integer values (like constraint_ub_GF6PTA in your notebook) are reactions required for an optimal solution and should be getting set to 1.0.

In the 0.5.X gapfilling model, reactions were added to the used_reactions list if flux through the indicator reaction exceeded a numerical cutoff, rather than using ind.primal == 1.0. After gapfilling some of my own models for biomass production, only ~1/3 of the reaction sets for each model actually allowed flux through biomass. I manually inspected the SUXModelMILP object after gapfilling and found that several reactions had non-zero fluxes below the numerical cutoff, and lowering the cutoff resolved the issue (i.e. my base model + the returned reactions had a non-zero solution).

I'm not familiar enough with the old or new solver interfaces to interpret this, but I hope it offers hints. it seems like the solver parameters would require some tuning for ind.primal == 1.0 to work.

As an aside, it may be nice to include gapfilling tests that are non-trivial (e.g. more complex than removing essential reactions from a model and then adding them back), but runtime will obviously be an issue.

cobra/core/model.py
```diff
@@ -316,6 +316,8 @@ def copy(self):
         except Exception:  # pragma: no cover
             new._solver = copy(self.solver)  # pragma: no cover
 
+        new._contexts = list()
```
mmundy42: What is _contexts being used for? I don't see it being used anywhere else in the pull request.

mmundy42:
Overall, the API looks good. Here are my general comments:

1. Are there any practical limits to the size of the universal model? There are thousands of reactions in some of the universal models I've worked with.
2. I would like to see a way to gap fill to a specific medium on the gapfill() function. It looks like the only option is to gap fill to a "complete" medium by using exchange reactions for all metabolites in the model. Not sure of the best way to implement that, maybe another parameter with the list of exchange reaction IDs? 
3. I would like to see a way to set a cost on a per reaction basis. It looks like that could be accomplished if the user could set "gapfilling_type" on the reactions in the universal model and then include the corresponding data in the penalties dictionary. That would require extend_model() to only set "gapfilling_type" if it is not already set.
4. How should a user add the reactions returned by the gapfill() function to the model? Would it make sense to have a flag on gapfill() to update the model with the reactions?
5. Would be nice to have docstrings for all of the methods in the Gapfiller class.

cobra/flux_analysis/gapfilling.py
```diff
@@ -1,9 +1,176 @@
...
+        self.model.add_cons_vars([constraint], sloppy=True)
+
+    def extend_model(self, exchange_reactions=False, demand_reactions=True):
+        for rxn in self.universal.reactions:
+            rxn.gapfilling_type = 'universal'
+        for met in self.model.metabolites:
+            if exchange_reactions:
+                rxn = self.universal.add_boundary(
+                    met, type='exchange_smiley', lb=-1000, ub=0,
+                    reaction_id='SMILEY_EX_{}'.format(met.id))
```
mmundy42: Since this is a generic gap fill algorithm, does it make sense for exchange and demand reactions to be prefixed with "SMILEY"?

cobra/core/model.py
hredestig: It's not, just realized, and got bitten by, the fact that when you copy a model, it's history manager context is also copied but that doesn't make a lot of sense to me so adding this fix here.

hredestig:
@gregmedlock thanks for having a look! I already tried all combinations I could think of with the threshold for the indicator but to no avail, the primal really is zero, causing violated constraints that the solver somehow is happy with 😕

cobra/flux_analysis/gapfilling.py
hredestig: No indeed, just copy-paste to be cleaned up

hredestig:
@mmundy42 thanks for all the comments! The purpose of this pr was to re-implement the 0.5.11 version to a point where it did pretty much the same thing with some smaller adjustments. After that we can always improve further.

1. I'm sure theres is, it's MILP so with big models this will undoubtedly be slow. I think @cdiener has a good alternative for more realistic scenarios with the CORDA method
2. To do that, you could just set `exchange_reactions=False` and then use a universal model that included the exchange reactions for that chosen medium no?
3. That could easily be included, so that the penalties dict is allowed to have reaction ids as well as the gapfilling type
4. The intention is that you just `model.add_reactions(list_of_reactions)`
5. Oh yes, more docs to come. Just want to get it to work first, currently it works sometimes with cplex and never with glpk....

hredestig:
@gregmedlock found the issue, in the end, it was exactly what you mentioned just that the new solver interface also implemented rounding so the reported values were == 0. Fixing

hredestig:
## Update
Introduce new class `cobra.flux_analysis.gapfilling.GapFiller` and helper
function `gapfill` to replace the `growMatch` and `SMILEY`
implementations. The functionality is near identical with the
differences

- model is not made irreversible so the universal model has be
  constructed accordingly. This has the benefit that the proposed
  reactions can be added directly the the original model without
  creating any funny irreversible reactions where such are not needed.
- penalties can be set on a per-reaction basis.
- the proposed solution is validated and exception thrown if the
  algorithm fails.

hredestig:
@mmundy42 please have look again, see what you think

mmundy42:
> To do that, you could just set exchange_reactions=False and then use a universal model that included the exchange reactions for that chosen medium no?

Yes, that would work just fine.

Reviewing the updates now ...

cobra/core/model.py
mmundy42: Would be nice to have a comment to that effect so that we remember why this line is here.

hredestig:
The failing build on linux is unrelated to this, just another "No LP  solvers found" plague.

cobra/flux_analysis/gapfilling.py
```diff
@@ -2,8 +2,350 @@
...
+            used to gapfill the model.
+
+        Raises
+        ------
+        RuntimeError
+            If the model fails to be validated (i.e. the original model with
+            the proposed reactions added, still cannot get the required flux
+            through the objective).
+        """
+        used_reactions = list()
```
mmundy42: Just a suggestion (that you can ignore), a better variable name might be "found_reactions" as in "reactions from the universal model that were found to gap fill the model".

mmundy42:
@hredestig, the latest commit looks good to me. Thanks for adding the support to set costs per reaction and the additional documentation.

cobra/core/model.py
hredestig: good point, fixed, and added test for the new behavior

cobra/flux_analysis/gapfilling.py
hredestig: Thanks for the suggestion, since I either use or don't use a reaction in the model however, I think I'll stick to current name.

cobra/flux_analysis/gapfilling.py
```diff
@@ -2,8 +2,350 @@
...
+    exchange (uptake) reactions added and 1 for added demand reactions.
+
+    Note that this is a mixed-integer linear program and as such will
+    expensive to solve for large models. Consider using CORDA instead _[3,4].
+
+    Parameters
+    ----------
+    model : cobra.Model
+        The model to perform gap filling on.
+    objective : model.interface.Objective
```
cdiener: Maybe we can alternatively simply use the previously set objective and have the user assign the objective with the usual `model.objective = x`. Would be consistent with many other algorithms from `flux_analysis` that do not allow the objective to be set in the function call (variability, single_deletions, etc.).

cobra/flux_analysis/gapfilling.py
```diff
@@ -2,8 +2,350 @@
...
+
+        Raises
+        ------
+        RuntimeError
+            If the model fails to be validated (i.e. the original model with
+            the proposed reactions added, still cannot get the required flux
+            through the objective).
+        """
+        used_reactions = list()
+        for i in range(iterations):
```
cdiener: So I haven't really read the original papers am am not completely sure what the loop does. In case it is used to find redundant activation pathways it might benefit from a break statement in case that solution is empty (no additional reactions found). Otherwise it just reruns the same optimization several times without doing anything.

documentation_builder/gapfilling.ipynb
```diff
@@ -0,0 +1,255 @@
...
+    "solution = gapfill(model, \"f6p_c\", universal)\n",
+    "for reaction in solution[0]:\n",
+    "    print(reaction.id)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Finally, note that using mixed-integer linear programming is computationally quite expensive and for larger models you may want to consider using [CORDA](https://github.com/cdiener/corda)."
```
cdiener: In case there are others we might mention them as well. @mmundy42 I think mackinac is a good fit, right? 

There is also gimme and gim3e but the last time I checked they looked abandoned and many others (tINIT, mCADRE, etc.) are only available in Matlab. 

An alternative would be to link to the reconstruction or gapfilling tag on our website:
http://opencobra.github.io/cobrapy/tags/reconstruction/
http://opencobra.github.io/cobrapy/tags/gapfilling/
Those are probably kept more up-to-date.

cobra/flux_analysis/gapfilling.py
hredestig: I was also hesitant about this.. added it to mimic the SMILEY function the then also required the pretty ugly function `get_objective_for`. Happy to simplify by removing this.

documentation_builder/gapfilling.ipynb
hredestig: Good idea will add those tag links

mmundy42: Yes, mackinac includes a create_universal_model() function that creates a universal model from a ModelSEED template model that you could use with the gapfill() function. And there is a gapfill_modelseed_model() function that runs the gapfill function on the ModelSEED server.

I should also make a version of cobrababel available as it has functions for creating universal models from other databases.

And it looks like mCADRE has been abandoned too according to the README on github.

cobra/flux_analysis/gapfilling.py
cdiener: Ok, so it's not really redundancy. So if I set iterations to 10 it will eventually also include alternative pathways of length 9 even if the shortest one is 1? So iterations should be kept small right? Either way we might mention the consequences of this parameter in the docstring.

cobra/flux_analysis/gapfilling.py
cdiener: Ah ok. Didn't see that we need it for backwards compatibility. In that case better remove it after growMatch and SMILEY are deprecated in favour of the new function...

hredestig: Now SMILEY and growMatch still do what they should but the new implementation has less mess

cobra/flux_analysis/gapfilling.py
hredestig: all reactions are always included, just those used in previous solutions get increasingly expensive. Docstring updated.

PR #503, opened 2017-05-04
Fixing and adding test for combine kwarg in reaction.add_metabolites

pstjohn:
Previously using `reaction.add_metabolites(*args, combine=False)` in a model context would result in the wrong reaction being returned.


Midnighter:
I know this is done and merged. Just wondering why the use of `iterkeys`? Isn't `for key in dict: pass` perfectly fine in all versions of Python?

pstjohn:
Yeah I guess you're right, no real reason behind iterkeys

Issue #506, opened 2017-05-10
Errors with reaction.remove_from_model and forward_variable


pstjohn:
This is failing on devel:

```python
import cobra
import cobra.test

model = cobra.test.create_test_model('textbook')

with model:
    model.reactions.PFK.remove_from_model()

model.reactions.PFK.forward_variable
```

Gives the error message:

```
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    model.reactions.PFK.remove_from_model()
  File "/Users/pstjohn/Packages/cobrapy/cobra/core/reaction.py", line 545, in remove_from_model
    self._model.remove_reactions([self], remove_orphans=remove_orphans)
  File "/Users/pstjohn/Packages/cobrapy/cobra/core/model.py", line 604, in remove_reactions
    forward = reaction.forward_variable
  File "/Users/pstjohn/Packages/cobrapy/cobra/core/reaction.py", line 143, in forward_variable
    assert self._forward_variable.problem is self.model.solver
AssertionError
```

Apparently the forward_variable.problem isn't being reset to the model's problem. I'll try to look into whats going on. Anyone have a good guess as to why that might be happening?

pstjohn:
Looks like we need to reset the reaction variable cache. PR incoming

pstjohn:
Apparently a larger issue, this is still failing on my PR:

```python
import cobra
import cobra.test

model = cobra.test.create_test_model('textbook')

with model:
    model.reactions.ATPM.remove_from_model()
    model.metabolites.atp_c.remove_from_model()


with model:
    model.reactions.ATPM.remove_from_model()
    model.metabolites.atp_c.remove_from_model()
```

pstjohn:
Reopening just since we're still not out of the woods with that last problem. Looking into it now.

hredestig:
Closing again with #508

PR #508, opened 2017-05-11
deprecating reaction.delete, fixing remove_from_model

pstjohn:
This should close the remaining issue in #506.

Ultimately I don't think the old behavior in reaction.remove_from_model / reaction.delete really make sense if we're trying to get the new context systems to work. They seem to be shortcut methods to avoid copying a reaction object, but if a reaction is inside a context, we'll need to keep a copy around in order to restore it quickly later.

Is anyone opposed to deprecating `reaction.delete` and keeping `remove_from_model`? That way its parallel to the metabolite case, which only implements `remove_from_model`.

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -569,8 +569,7 @@ class Model(Object):
         # from cameo ...
         self._populate_solver(reaction_list)
 
-    def remove_reactions(self, reactions, delete=True,
-                         remove_orphans=False):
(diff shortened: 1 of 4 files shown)


hredestig:
I agree, don't see the point in that one. Just made small fix to remove unused parameter, hope you didn't mind the push @pstjohn

pstjohn:
Not at all, thanks for the catch!

Issue #511, opened 2017-05-18
consistent behavior between `add_metabolites` and `add_reactions`


pstjohn:
Currently, `model.add_reactions` will throw a `ValueError` if any of the reaction IDs are already in the model, while `model.add_metabolites` will just not add them.

Is there any motivation to this? If we were to standardize them, which would we prefer?

cdiener:
Good one. Considering the `+` operators for models it might be easier to implement if it would just not add them. This way adding reactions and metabolites from one model to another only adds new ones. In this case we could add warnings to the logger so it isn't done completely silent...

hredestig:
I'd like the behavior to not just add them, what the user is requesting is then already achieved so no reason to error. There is very little logging going  on in cobrapy currently but would be nice to add that a bit all over the package..

Issue #505, opened 2017-05-09
Model objective after adding two models different between 0.5.11 and 0.6


mmundy42:
I ran across this while getting mminte running on cobra 0.6. With cobra 0.5.11, when two models are added with the "+=" operator, the objective is updated too because of the way reaction objective coefficients are handled. For example, using the BT and FP models from Thiele, I can add a prefix to the reaction IDs, add the models together and both objective reactions are set in the combined model:

    Original BT objective: {<Reaction Biomass_BT_v2 at 0x110953fd0>: 1.0}
    Original BT objective value: 0.440738418826
    Original FP objective: {<Reaction Biomass_FP at 0x110967c10>: 1.0}
    Original FP objective value: 0.169337957302
    Combined objective: {<Reaction FP_Biomass_FP at 0x1110f8310>: 1.0, <Reaction BT_Biomass_BT_v2 at 0x110953fd0>: 1.0}
    Combined objective value: 1.10532735493

With cobra 0.6, when two models are added with the "+=" operator, the objective in the combined model is unchanged. For example:

    Original BT objective: {<Reaction Biomass_BT_v2 at 0x110f2d510>: 1.0}
    Original BT objective value: 0.440738418827
    Original FP objective: {<Reaction Biomass_FP at 0x111d6c7d0>: 1.0}
    Original FP objective value: 0.169337957301
    Combined objective: {<Reaction BT_Biomass_BT_v2 at 0x110f2d510>: 1.0}
    Combined objective value: 1.07958792212

Would it make sense for the "+" and "+=" operators to explicitly set the objective in cobra 0.6?

hredestig:
I must say I find `__add__` and `__iadd__` a bit confusing altogether.. They also lack unit-tests I notice. I guess you preferred having the objective become the sum of the two objectives? I find that somewhat surprising, or at least not obvious behavior.. Also no direct feeling what the obvious behavior would be..

I think I'd favor removing them altogether and replace with a utility function `combine_models` which would allow arguments controlling what should happen with reactions that are already present in the other model and what should happen with the objective. What do you think? Is `modela + modelb`  very useful? Is `modela += modelb` also important?

pstjohn:
We should figure out what we want to do with the operator overloading methods. We talked about them in #311 and #400 as well.

For reactions, `__imul__` and `__add__` is sometimes useful, i.e. if I'm writing a reaction thats comprised of many individual parts. Like a biomass function, you can do something like
```python
[...]

PR #513, opened 2017-05-19
Fix/model add etc

hredestig:
- chore in setup.cfg for parsing versions
- make adding existing reactions just ignore them instead of ValueError #511 
- make objective the sum of the two models objectives when adding and introduce new `model.merge` #505. Should all reactions be prefixed here instead of only the pre-existing?

I go on vacation a couple of weeks now, welcome to finish the PR 😛  😎  🌞

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -228,35 +228,25 @@ class Model(Object):
             set_active_bound(rxn, 0)
 
     def __add__(self, other_model):
-        """Adds two models. +
+        """Add the content of another model to this model (+).
 
-        The issue of reactions being able to exists in multiple Models now
-        arises, the same for metabolites and such.  This might be a little
-        difficult as a reaction with the same name / id in two models might
-        have different coefficients for their metabolites due to pH and whatnot
-        making them different reactions.
-
+        The model is copied as a new object, with a new model identifier,
+        and copies of all the reactions in the other model are added to this
+        model. The objective is the sum of the objective expressions for the
+        two models.
         """
-        new_model = self.copy()
-        new_reactions = deepcopy(other_model.reactions)
-        new_model.add_reactions(new_reactions)
-        new_model.id = self.id + '_' + other_model.id
-        return new_model
+        warn('use model.merge instead', DeprecationWarning)
+        return self.merge(other_model, objective='sum', inplace=False)
 
     def __iadd__(self, other_model):
-        """Adds a Model to this model +=
+        """Incrementally add the content of another model to this model (+=).
 
-        The issue of reactions being able to exists in multiple Models now
-        arises, the same for metabolites and such.  This might be a little
-        difficult as a reaction with the same name / id in two models might
-        have different coefficients for their metabolites due to pH and whatnot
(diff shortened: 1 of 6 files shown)


cobra/test/test_model.py
```diff
@@ -691,6 +691,33 @@ def test_add_reaction_orphans(self, model):
         # 'check not dangling metabolites when running Model.add_reactions
         assert len(orphan_metabolites) == 0
 
+    def test_merge_models(self, model, tiny_toy_model):
+        with model, tiny_toy_model:
```
pstjohn: Is the idea that model merges will get reversed after the context exits? That's awesome btw 😄 . Could we add a couple tests to make sure that happens?

cobra/core/model.py
```diff
@@ -938,3 +932,52 @@ def __exit__(self, type, value, traceback):
...
+        Custom constraints and variables from right models are also copied
+        to left model, however note that, constraints and variables are
+        assumed to be the same if they have the same name.
+
+        right : cobra.Model
+            The model to add reactions from
+        prefix_existing : string
+            Prefix the reaction identifier in the right that already exist
+            in the left model with this string.
+        inplace : bool
```
pstjohn: If we are trying to make this context-aware, maybe mention that in the `inplace` argument?

cobra/core/model.py
```diff
@@ -938,3 +932,52 @@ def __exit__(self, type, value, traceback):
...
+            Add reactions from right directly to left model object. Otherwise,
+            create a new model leaving the left model untouched.
+        objective : string
```
cdiener: Good idea :+1:

cobra/core/model.py
```diff
@@ -491,32 +487,30 @@ def add_boundary(self, metabolite, type="exchange", reaction_id=None,
...
-                              in self.reactions]
-
-        if len(reactions_in_model) > 0:
-            raise ValueError(
-                "The following reactions are already in the model: {}"
-                "".format(", ".join(reactions_in_model)))
+        # First check whether the metabolites exist in the model
+        existing = [rxn for rxn in reaction_list if rxn.id in self.reactions]
+        for rxn in existing:
+            LOGGER.info('skip adding reaction %s as already existing' % rxn.id)
```
Midnighter: Just nit-picking. It's better to write it with comma instead of the format operator. The way the current line is written, the string is always evaluated thus adding a slight cost even if the logging level ignores info. If you write `.info("... %s ...", rxn.id)` instead, the string will only be formatted if info level is actually desired.

hredestig: fixed, cheers

cobra/core/model.py
hredestig: Cheers, doc updated

cobra/test/test_model.py
hredestig: That all came for free as using `add_cons_vars` etc and the context in the unit-tests was mainly so that we don't have re-load the model for the next test, but true, a couple of extra checks doesn't hurt; added

PR #521, opened 2017-06-09
[sampling] use shared memory for all large arrays

### LINKED ISSUE #516: Metabolic sampling
I tried both with OptGPSampler and sampling, in both cases show something like this

  File "/home/[usr]/.local/lib/python3.5/site-packages/cobra/flux_analysis/sampling.py", line 644, in sample
    chunksize=1)
  File "/usr/lib/python3.5/multiprocessing/pool.py", line 260, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib/python3.5/multiprocessing/pool.py", line 608, in get
    raise self._value
  File "/usr/lib/python3.5/multiprocessing/pool.py", line 385, in _handle_tasks
    put(task)
  File "/usr/lib/python3.5/multiprocessing/connection.py", line 206, in send
    self._send_bytes(ForkingPickler.dumps(obj))
  File "/usr/lib/python3.5/multiprocessing/connection.py", line 393, in _send_bytes
    header = struct.pack("!i", n)
struct.error: 'i' format requires -2147483648 <= number <= 2147483647

cdiener:
Sorry for the delay. Took me quite a while to figure out how to manage class attributes in shared memory with `multiprocessing.Pool`. This PR moves all of those to shared memory so additional process in the OptGPSampler do not duplicate those anymore. This has the following effects on `OptGPSampler`:

1. constant memory use independent of the number of processes
2. a bit faster `sample` since the initial memory copy is gone
3. now allows sampling of models with arbitrary size in memory (as long as you have enough RAM for the built matrices), no more 4GB limit due to `pickle` constraints
4. runs much smoother in jupyter now, no idea why though

Also added a section to the docstrings giving the rough order of memory usage.

[Benchmark and convergence analysis] (https://github.com/cdiener/cobra-docker/blob/master/sampling_benchmark.ipynb)

Fixes #516.

### DIFF
--- a/cobra/flux_analysis/sampling.py
+++ b/cobra/flux_analysis/sampling.py
@@ -56,6 +56,34 @@ projection : numpy.matrix
 """
 
 
+def shared_np_array(shape, data=None):
+    """Create a new numpy array that resides in shared memory.
+
+    Parameters:
+    ----------
+    shape : tuple of ints
+        The shape of the new array.
+    data : numpy.array
+        Data to copy to the new array. Has to have the same shape.
+    """
+    N = np.prod(shape)
+    array = Array(ctypes.c_double, int(N))
+    np_array = np.frombuffer(array.get_obj())
+    np_array.reshape(shape)
+
+    if data is not None:
+        if len(shape) != len(data.shape):
+            raise ValueError("`data` must have the same dimensions"
+                             "as the created array.")
+        same = all(x == y for x, y in zip(shape, data.shape))
+        if not same:
+            raise ValueError("`data` must have the same shape"
+                             "as the created array.")
+        np_array = data[:]
+
+    return np_array
+
+
 # Has to be declared outside of class to be used for multiprocessing :(
 def _step(sampler, x, delta, fraction=None):
     """Sample a new feasible point from the point `x` in direction `delta`."""
@@ -159,6 +187,8 @@ class HRSampler(object):
         # check if there any non-zero equality constraints
         equalities = prob.equalities
         b = prob.b
+        bounds = np.atleast_2d(prob.bounds).T
+        var_bounds = np.atleast_2d(prob.variable_bounds).T
         homogeneous = all(np.abs(b) < feasibility_tol)
         fixed_non_zero = np.abs(prob.variable_bounds[:, 1]) > feasibility_tol
         fixed_non_zero &= prob.variable_fixed
@@ -177,14 +207,17 @@ class HRSampler(object):
         projection = nulls.dot(nulls.T)
         # convert bounds to a matrix and add variable bounds as well
         return Problem(
-            equalities=equalities,
-            b=b,
-            inequalities=prob.inequalities,
-            bounds=np.atleast_2d(prob.bounds).T,
-            variable_fixed=prob.variable_fixed,
-            variable_bounds=np.atleast_2d(prob.variable_bounds).T,
-            projection=projection,
-            homogeneous=homogeneous)
+            equalities=shared_np_array(equalities.shape, equalities),
+            b=shared_np_array(b.shape, b),
+            inequalities=shared_np_array(prob.inequalities.shape,
+                                         prob.inequalities),
+            bounds=shared_np_array(bounds.shape, bounds),
+            variable_fixed=shared_np_array(prob.variable_fixed.shape,
+                                           prob.variable_fixed),
+            variable_bounds=shared_np_array(var_bounds.shape, var_bounds),
+            projection=shared_np_array(projection.shape, projection),
+            homogeneous=homogeneous
+        )
 
     def generate_fva_warmup(self):
         """Generate the warmup points for the sampler.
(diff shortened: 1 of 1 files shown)


cobra/flux_analysis/sampling.py
```diff
@@ -56,6 +56,40 @@
...
+
+    Parameters:
+    ----------
+    shape : tuple of ints
+        The shape of the new array.
+    data : numpy.array
+        Data to copy to the new array. Has to have the same shape.
+    """
+    N = np.prod(shape)
+    array = Array(ctypes.c_double, int(N))
```
hredestig: neat, learned something :)

cobra/flux_analysis/sampling.py
```diff
@@ -56,6 +56,40 @@
...
+    """Create a new numpy array that resides in shared memory.
+
+    Parameters:
+    ----------
+    shape : tuple of ints
+        The shape of the new array.
+    data : numpy.array
+        Data to copy to the new array. Has to have the same shape.
+    """
+    N = np.prod(shape)
```
cdiener: Yes definitely!

cobra/flux_analysis/sampling.py
```diff
@@ -177,14 +213,17 @@ def __build_problem(self):
...
+            bounds=shared_np_array(bounds.shape, bounds),
+            variable_fixed=shared_np_array(prob.variable_fixed.shape,
+                                           prob.variable_fixed),
+            variable_bounds=shared_np_array(var_bounds.shape, var_bounds),
+            projection=shared_np_array(projection.shape, projection),
+            homogeneous=homogeneous
+        )
 
     def generate_fva_warmup(self):
         """Generate the warmup points for the sampler.
```
cdiener: Sure will check.

cobra/flux_analysis/sampling.py
```diff
@@ -56,6 +56,40 @@
 """
 
 
+def mp_init(obj):
+    """Initialize the multiprocessing pool."""
+    global sampler
```
cdiener: Only needed for OptGP and I didn't want to pollute the module name space. But no strong feelings either way so whatever you prefer.

cdiener:
Okay added your suggestions @hredestig. Hope it's okay for you :smile:

cobra/flux_analysis/sampling.py
```diff
@@ -637,14 +681,17 @@ def sample(self, n, fluxes=True):
             n = n_process * self.np
             # No with statement or starmap here since Python 2.x
             # does not support it :(
-            mp = Pool(self.np)
+            mp = Pool(self.np, initializer=mp_init, initargs=(self,))
             chains = mp.map(
                 _sample_chain,
-                zip([self] * self.np, [n_process] * self.np, range(self.np)),
+                zip([n_process] * self.np, range(self.np)),
                 chunksize=1)
```
Midnighter: Have you played with `imap_unordered` and a larger chunksize at some point? Or is the communication cost negligible compared to in-process computation time?

cobra/flux_analysis/sampling.py
```diff
@@ -56,6 +56,40 @@
...
+
+    if data is not None:
+        if len(shape) != len(data.shape):
+            raise ValueError("`data` must have the same dimensions"
+                             "as the created array.")
+        same = all(x == y for x, y in zip(shape, data.shape))
+        if not same:
+            raise ValueError("`data` must have the same shape"
+                             "as the created array.")
+        np_array = data[:]
```
Midnighter: Shouldn't this be `np_array[:] = data` or does that have the same effect? Also, would you expect cases where your shape tests actually fail (given proper shape from the data array, of course)?

cdiener: No that would be wrong. np_array is the one in shared memory and you have to copy the non-shared values into it.

cobra/flux_analysis/sampling.py
cdiener: So, no pool in the process is ever recycled. A pool of size n will receive exactly n problems to sample, so imap_unordered does not really make sense IMO. A chunksize of one guarantees that each process will run one sampling chain otherwise it would *theoretically* be possible that a process received two chains and another none (unlikely).

cobra/flux_analysis/sampling.py
Midnighter: Exactly, that was my intention, too. I did some testing and actually either syntax ensures that the values are copied rather than just assigned to the variable name. I typically use the one I suggested in my comment above because `var = x` is usually assignment, so that `var[:] = ...` makes it clear to myself that I want to copy values into `var`. All good :)

cobra/flux_analysis/sampling.py
Midnighter: So what's a typical size for `n`? My strategy usually is to create a pool of size equal to the number of processors and not equal to the size of the problem. Maybe those numbers are equal here, I may have overlooked that.

cobra/flux_analysis/sampling.py
cdiener: Sorry, sorry, sorry. You are right! Will change it.

cobra/flux_analysis/sampling.py
cdiener: The n is user supplied. Sometimes you might not want to block the entire system for the sampling (e.g. shared server). So we let the user pick. The idea is that the problem is really easy to divide into completely independent sub-problems so you can create the optimal workflow for any number of processes easily.

cobra/flux_analysis/sampling.py
Midnighter: As I said, both syntaxes achieve the same thing so it's all good to me. Just didn't expect that :wink:

cdiener: But still it is an error on my part. Your way is much more explicit.

PR #523, opened 2017-06-11
remove rxn.x call in summary

pstjohn:
remove internal call to deprecated method


Midnighter:
I thought I left that there because the legacy solvers were using LegacySolution which doesn't have the new attributes. Since all the tests pass it seems fine, though. Glad to know that :)

PR #524, opened 2017-06-12
Add some pydocstyle options to `setup.cfg`

cdiener:
Adds some options for pydocstyle that:

1. use the numpy docstyle standard
2. ignore tests since those do not require docstrings

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ parse = (?P<major>\d+)
 	\.(?P<minor>\d+)
 	\.(?P<patch>\d+)
 	(?P<release>[a]*)(?P<num>\d*)
-serialize = 
+serialize =
 	{major}.{minor}.{patch}{release}{num}
 	{major}.{minor}.{patch}
 tag_name = {new_version}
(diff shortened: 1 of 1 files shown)


pstjohn:
So this will enforce numpy-style docstrings? Where does that check get executed, by codecov?

cdiener:
It will if you have `pydocstyle>=2.0` otherwise it gets ignored. We currently have no docstyle linter set up in the CI since a lot of source files do not have correct dosctrings yet. But if you run `pydocstyle` on the repo the option will be respected. This also works with linter plugins such as Atom's `linter-pydocstyle` (but not with `flake8_docstrings` due to a bug).

Midnighter:
I would love to move to flake8 and flake8-docstrings in the future (which works for me when pinning pydocstyle to an older version such as 1.1.1). At the moment the effort of switching to the more rigorous flake8 checks don't seem worth it, though. I would put that in with an eventual redesign of the library. It is a good idea to make sure we stick to the numpydoc style. So thanks for the PR!

cdiener:
Thanks @Midnighter! Does flake8-docstrings support reading pydocstyle config values? Last time I checked they ignored the pydocstyle section in the setup.cfg...

Midnighter:
I have successfully used the following:

```
[pydocstyle]
match_dir = ...
```

but I can't vouch for all options being picked up.

Issue #520, opened 2017-06-09
Summary method after pfba


ma-blaetke:
There is a problem with the summary method called on a model after doing pfba(). I get the following error message:

_OptimizationError: Likely no solution exists. Original solver message: Unable to retrieve attribute 'X'._
(Below, complete output)

Nevertheless, if I call e.g. the x_dict on the result of pfba(), there clearly is an optimal solution. What I also get from the x_dict, that now reversible reactions are splitted, instead of summed up as before. Is this on purpose?

```
---------------------------------------------------------------------------
OptimizationError                         Traceback (most recent call last)
<ipython-input-121-8a0c3b7fef90> in <module>()
      1 result_fba = cobra.flux_analysis.parsimonious.optimize_minimal_flux(cobra_model, solver="gurobi")
----> 2 cobra_model.summary()

//anaconda/lib/python2.7/site-packages/cobra/core/model.pyc in summary(self, threshold, fva, floatfmt, **kwargs)
    926         from cobra.flux_analysis.summary import model_summary
    927         return model_summary(self, threshold=threshold, fva=fva,
--> 928                              floatfmt=floatfmt, **kwargs)
    929 
    930     def __enter__(self):

//anaconda/lib/python2.7/site-packages/cobra/flux_analysis/summary.pyc in model_summary(model, threshold, fva, floatfmt, **solver_args)
    144             metabolite_fluxes[met] = {
    145                 'id': format_long_string(met.id, 15),
--> 146                 'flux': stoich * rxn.x}
    147 
    148     # Calculate FVA results if requested

//anaconda/lib/python2.7/site-packages/cobra/core/reaction.pyc in x(self)
    476         """
    477         warn("Please use reaction.flux instead.", DeprecationWarning)
--> 478         return self.flux
    479 
    480     @property

//anaconda/lib/python2.7/site-packages/cobra/core/reaction.pyc in flux(self)
    317             raise_from(OptimizationError(
    318                 "Likely no solution exists. Original solver message: {}."
--> 319                 "".format(str(err))), err)
    320 
    321     @property

//anaconda/lib/python2.7/site-packages/future/utils/__init__.pyc in raise_from(exc, cause)
    452             raise TypeError("exception causes must derive from BaseException")
    453         e.__context__ = sys.exc_info()[1]
--> 454         raise e
    455 
    456     exec('''

OptimizationError: Likely no solution exists. Original solver message: Unable to retrieve attribute 'X'.
```

[...]

PR #525, opened 2017-06-15
Fix FVA in summary calls

pstjohn:
We've got bigger issues, #520, but this fixes a bug with the FVA handling in summary.


pstjohn:
yeah I should probably write tests for these... lets see what happens in #520

PR #529, opened 2017-06-21
Update pubs and releases.

cdiener:
Regular updates for releases and publications.

### Minor design changes
Bottom of publication list shows the total number of pubs for that year.

### Preview 
https://opencobra.github.io/cobrapy-website

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@
 	<meta http-equiv="X-UA-Compatible" content="IE=edge">
 	<meta name="description" content="My personal website incuding Posts and my CV.">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
-	<meta name="generator" content="Hugo 0.21-DEV" />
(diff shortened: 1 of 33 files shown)


pstjohn:
Looks great. What do you guys think of the work going into the julia versions of cobra? I've been thinking that they have the advantage in dealing with all of the different C-level solvers, especially with something like [MathProgBase.jl](mathprogbasejl.readthedocs.io/en/latest). 

It doesn't look like there's a great Python -> Julia bridge at the moment though, so sadly I'm not sure that's something we can take advantage of here.

cdiener:
I think a bridge would also create a lot of overhead. As you said, the major problem for us is that there is no good C-library wrapping more of the solvers. Otherwise you could just write bindings to that directly using OpenMP or something. I think (but am not sure) that if you use openblas or mkl at least glpk should be able to use those... I did not really understand their Fig. 1A though :(

hredestig:
Looks great, thanks!

PR #535, opened 2017-07-04
feat: pfba-fva

hredestig:
Add the option to include a constraint on the summed absolute fluxes,
like pfba, when doing FVA to get (possibly) more realistic bounds.


hredestig:
Thanks for having a look, I just adjusted the doc. On the release number, indeed, from semantic version I also interpret it so, any new feature no matter how small should bump the middle number. However, I like to collect changes in release notes as we go along but not knowing what we're adding in unreleased devel branch I haven't bothered appropriate naming (to be fixed before release). A better way would have been `next-version.md`. Next time..

pstjohn:
> Setting this value may lead to more realistic predictions of the effective flux bounds.

That's interesting, is this method published anywhere? It would be cool to compare FVA vs pFVA bounds against some experimental data.

hredestig:
Actually not sure if there is a good reference, hence the 'may' :) @phantomas1234 do you know?

PR #545, opened 2017-07-17
feat: model.slim_optimize

### LINKED ISSUE #540: Implement quick-optimization function
We need a method on model to do what `model.optimize` does but which doesn't create a full solution to avoid the need for calling `model.solver.optimize` directly, and provide a method that we can document and explain when to use.

Could be implemented something like
```
def quick_optimize(self, value=0.):
    model.solver.optimize()
    try:
        assert_optimal(model)
        return model.solver.objective.value
    except OptimizationError:
        return value
```

An alternative would be to just provide better docs in the notebook for  how/when to use `model.solver.optimize`

Thoughts?

hredestig:
- Introduce new method on Model to do optimization without creating a solution.
- a couple of tests emitted visible warnings during tests, capture them appropriately
- fixes #540

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -23,7 +23,7 @@ from cobra.util.context import HistoryManager, resettable, get_context
 from cobra.util.solver import (
     SolverNotFound, get_solver_name, interface_to_str, set_objective, solvers,
     add_cons_vars_to_problem, remove_cons_vars_from_problem, choose_solver,
-    check_solver_status)
+    check_solver_status, assert_optimal)
 from cobra.util.util import AutoVivification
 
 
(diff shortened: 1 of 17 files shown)


cobra/core/model.py
```diff
@@ -778,6 +778,43 @@ def to_array_based_model(self, deepcopy_model=False, **kwargs):
         from cobra.core.arraybasedmodel import ArrayBasedModel
         return ArrayBasedModel(self, deepcopy_model=deepcopy_model, **kwargs)
 
+    def slim_optimize(self, error_value=0., message=None):
```
hredestig: I realised this and change it to `error_value=float('nan')` which I think should be less confusing. Personally I like to option get a value if error to avoid the long try/except blocks.  `raise_error=False` as you propose is also an option but then don't find it obvious what the return value should be..

cdiener: Yeah that work. Only if I want the return value to be `None` if not optimal, there is no way now, right?

PR #546, opened 2017-07-18
feat: add html representations

hredestig:
When using cobrapy in jupyter, it is useful to have html representations
of common classes. Add for model, reaction, metabolite and solution.

screen shot 2017-07-18 at 16 08 27

screen shot 2017-07-18 at 16 09 03

screen shot 2017-07-18 at 16 09 45

screen shot 2017-07-18 at 16 10 12

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -26,7 +26,6 @@ from cobra.util.solver import (
     check_solver_status, assert_optimal)
 from cobra.util.util import AutoVivification
 
-
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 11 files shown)


hredestig:
Cheers for the review! I tried with very long expression, e.g. even after `add_pfba` it looks quite okay I find and since the long string might be what you need, I somehow prefer to have the whole thing there. I have so many times now done the reaction string with names as well so really like having that one there, e.g. in yeast7 the ids are all cryptic :/

hredestig:
..should re-build all the notebooks prior to merge

Midnighter:
Big +1

Just wondering if we should communicate more clearly that 'Optimal' for the solution can be variable? Maybe by making it italic or in a `<pre>` block or so.

cobra/core/model.py
```diff
@@ -1020,3 +1019,35 @@ def merge(self, right, prefix_existing=None, inplace=True,
...
+                <td><strong>Compartments</strong></td>
+                <td>{compartments}</td>
+            </tr>
+          </table>""".format(
+            name=self.id,
+            address='0x0%x' % id(self),
+            num_metabolites=len(self.metabolites),
+            num_reactions=len(self.reactions),
+            objective=str(self.objective.expression),
+            compartments=", ".join(c for _, c in iteritems(self.compartments))
```
Midnighter: `itervalues` then?

hredestig: Yes of course 💦  but just realised values might not be defined so better use combination

hredestig:
Good point, but used `<em>` instead as `<pre>` makes ugly line break, guess this works as well.

PR #549, opened 2017-07-19
refactor: make model.medium pandas df

hredestig:
Make model.medium return data frame instead as nicer to read in
notebooks (backwards compatible)

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -9,6 +9,7 @@ from functools import partial
 from warnings import warn
 
 import optlang
+import pandas as pd
 import six
 import sympy
 from six import iteritems, string_types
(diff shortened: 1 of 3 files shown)


Midnighter:
I appreciate that you want to keep the same interface as before but I find a `DataFrame` with many columns and just one row a confusing format. In that case I'd rather use a `Series`. However, I would really like to set the upper and lower bound. So ideally `Model.medium` would return a data frame where the index is the Reaction ID, a column for the compound ID, compound name, lower bound, upper bound. When setting the compound columns could be optional. Can we make this happen with a conversion function or alternative method without breaking the old interface?

hredestig:
I agree.. that wide df is pretty ugly :/ Current behavior of `Model.medium` is however the upper bound of the import, keeping that feature as well as adding reaction ub, lb might get confusing though.. Will think about it..

mmundy42:
So the medium setter still uses a dict for the input but the medium getter returns a DataFrame? That seems a little confusing. I would prefer to keep the dict since I just reworked several packages to use the new medium getter and setter.

hredestig:
In this pr, you can use dataframe or dict for the setter as they behave very similarly, so this is currently backwards compatible. As discussed above though this requires a wide data frame which is kind of ugly...

hredestig:
I'm dropping this one, the benefit doesn't warrant the change.

PR #550, opened 2017-07-19
feat: essential genes and reactions

hredestig:
Add convenience functions for getting sets of essential genes and
reactions. Using this in two other packages that both depend on cobrapy
so better move here.


cdiener:
Those look like slower versions of `single_gene_deletion` and `single_reaction_deletion`. Or is there a difference I don't see?

hredestig:
You're right, making new pr. Although it does say something about unfortunate function name there..

cdiener:
Im also fine with renaming or refactoring the others, but probably leave that for after removing the old solver interface. Also commented on gitter. `abs(flux) > 0.` should be `abs(flux) > atol` where "atol" is the absolute tolerance of the solver.

hredestig:
Agreed, making fix!

PR #552, opened 2017-07-19
Add linear MOMA

cdiener:
Adds linear MOMA to `add_moma`.

### DIFF
--- a/cobra/flux_analysis/moma.py
+++ b/cobra/flux_analysis/moma.py
@@ -10,13 +10,16 @@ from cobra.solvers import get_solver_name, solver_dict
 
 
 def add_moma(model, linear=False):
-    """Add constraints and objective representing a MOMA
-    (minimization of metabolic adjustment) model.
+    """Add constraints and objective representing a MOMA (minimization of
+    metabolic adjustment) model.
 
     Parameters:
     -----------
 
-    model : a cobra model
+    model : cobra.Model
+        The model to add MOMA constraints and objective to.
+    linear : bool
+        Whether to use the linear MOMA formulation or not.
 
     Returns:
     --------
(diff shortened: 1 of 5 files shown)


cobra/flux_analysis/moma.py
```diff
@@ -9,7 +9,7 @@
 from cobra.solvers import get_solver_name, solver_dict
 
 
-def add_moma(model):
+def add_moma(model, linear=False):
     """Add constraints and objective representing a MOMA
     (minimization of metabolic adjustment) model.
```
hredestig: perhaps update the doc string, new parameter, here as well and brief explanation pro/con linear

cobra/util/solver.py
```diff
@@ -337,18 +338,32 @@ def add_absolute_expression(model, expression, name="abs_var", ub=None):
...
     ub : positive float
        The upper bound for the variable.
+    difference : positive float
+        The difference between the expression and the variable.
+    add : bool
+        Whether to add the variable to the model at once.
+
+    Returns
+    -------
+    list
```
hredestig: how about making the return value a `namedtuple('Components', ['variable', 'upper_constraint', 'lower_constraint'])` and avoid the need to remember the order they come? Tuples also work with `model.add_cons_vars`

cobra/flux_analysis/single_deletion.py
```diff
@@ -63,6 +63,10 @@ def single_reaction_deletion(cobra_model, reaction_list=None, solver=None,
     elif method == "moma":
         result = single_reaction_deletion_moma(cobra_model, reaction_list,
                                                solver=solver, **solver_args)
+    elif method == "linear moma":
```
hredestig: the docstring needs adjustment

hredestig:
Very nice, I just committed the suggested changes, if you agree with them, I'll go ahead and merge.

Midnighter:
Is this in line with the code in cameo? I didn't check yet.

cdiener:
Hi @hredestig found some other small points that need addressing here, just let me push those as well.

@Midnighter The API for MOMA in cobrapy is a bit different (see discussion in #367). The method used is the same (recentering) only that the cameo implementation needs to add twice as many variables and constraints since they define the absolute terms for forward and backward variables separately (which is surprising since they do not do that for the "normal" quadratic MOMA...).

cdiener:
Okay, let's see what the CI says. Also added a `solution` arg to MOMA to use a previous solution instead of recomputing it, makes the single_x_deletion functions with MOMA slightly faster and allows passing in solutions from for instance `pfba`.

PR #556, opened 2017-07-24
[sampling] fix inhomogeneous sampling and avoid recursion error

### LINKED ISSUE #558: Problem using the built-in sampling
I'm having an issue with using the built-in sampling in cobrapy v0.7.0. Upon sampling my model, I get identical values for all of the samples within each of my reactions of interest. I ran almost identical code (same reaction bounds, same model, same sampling parameters) on my computer using MATLAB and the optGpSampler code available [here](http://cs.ru.nl/~wmegchel/optGpSampler/) and received the desired result of varied values within my reactions of interest. I have tried using both the achr and optgp solving methods in cobrapy. The reason I would like to continue using the Python version of this software rather than sticking with the MATLAB version that works is that I would like to use the compute server that I have access to, but I had problems setting up optGpSampler on the server.

#### Code Sample

```python
import cobra
import pandas

from cobra.flux_analysis.sampling import sample

fluxes = pandas.read_excel("Flux_Values.xlsx") #Load flux values from Excel sheet
model = cobra.io.load_matlab_model("iCHOv1_gimme_final.mat", variable_name="gimmeK1") #Load CHOK1 model

for i in xrange(18):
    rxn_id = str(fluxes["Reaction"][i])
    rxn = model.reactions.get_by_id(rxn_id)
    flux_val = fluxes["Exchange rate(mmol/gDW/hr)"][i]
    if i < 6: #First 6 reactions in the Excel sheet are contraints
        rxn.lower_bound = flux_val
        rxn.upper_bound = flux_val
    else: #Last 12 reactions are the ones we want to sample
          #Set the fluxes to be +- 15% of the experimental seed value
        #We are sampling uptake rates, so all of the exchange reactions will have negative fluxes
        #Set bounds to be +- 15% of experimental seed value
        rxn.lower_bound = 1.15*flux_val
        rxn.upper_bound = .85*flux_val

model.objective = "biomass_cho_producing"

sModel = sample(model, 1500, thinning=1000, processes=4)  #Sample fluxes for model with 1500 returned points and 500 samples for each returned point. Uses 4 processes

sModel.to_excel('./outputs/output_new.xlsx') #Save flux values for later analysis

```

```matlab
load('CELS201_6mr9dq_Additional_Files\cels201_mmc2_6mr9dq\iCHOv1_gimme_final.mat','gimmeK1'); %Load CHOK1 model
[bounds, rxnList, raw] = xlsread('CalculatedfluxValues-AMBICExp1.xlsx', 'B2:C19'); %This is the same excel sheet as in the python code, I just renamed it
for i = 1:length(rxnList)
    if (i < 7)
        gimmeK1 = changeRxnBounds(gimmeK1, rxnList(i), bounds(i), 'b');
    else
        gimmeK1 = changeRxnBounds(gimmeK1, rxnList(i), bounds(i)*1.15, 'l');
        gimmeK1 = changeRxnBounds(gimmeK1, rxnList(i), bounds(i)*.85, 'u');
    end
end

model = changeObjective(gimmeK1,'biomass_cho_producing');
sModel = optGpSampler(model,[],1.5e3,1e3,4,'glpk',0);
```

#### Actual Output
As described above, the Python code returns and excel sheet that has identical values for all of the samples within many reactions, including the reactions that I am trying to study.
#### Expected Output
I expect the sampled values within each reaction, other than the reactions that I have specifically constrained, to vary. This is the result I achieve with the MATLAB version of the code.
#### Output of `cobra.show_versions()`

<details>

System Information
==================
OS                               Linux
OS-release  2.6.32-573.22.1.el6.x86_64
Python                          2.7.10

Package Versions
================
pip                              9.0.1
setuptools                      28.8.0
cobra                            0.7.0
future                          0.16.0
swiglpk                          1.4.3
optlang                          1.2.1
ruamel.yaml                    0.14.12
pandas                          0.19.2
numpy                           1.12.1
tabulate                         0.7.7
scipy                           0.19.0
matplotlib                       1.5.3

</details>

cdiener:
# Bug fixes

Gives a putative solution for recursion limit errors when using `multiprocessing.Pool`. Apparently arguments can not be a `zip` object. No idea why though...

---

Fixes bugs related to inhomogeneous sampling. Turns out the reprojection strategy was overly complicated and did not work well since it spans very little of the space. As a consequence inhomogeneous sampling often returned identical samples (fixes #558). Using `delta = any_warmup - center` as new direction already works for inhomogeneous problems since `S*(any_warmup - center) = b - b = 0`. This spans the search space much better and gives a much better distribution.   

# Minor fixes

The sampling problem now stores the nullspace rather than the entire projection (N * N.T) which reduces the memory footprint since it uses memory in the order of `2 * n_reaction * n_metabolites` instead of `4 * n_reactions^2`. Does not seem to influence sampling speed at all.

Log some minor issues the sampler resolves automatically as info for better debugging.

### DIFF
--- a/cobra/flux_analysis/sampling.py
+++ b/cobra/flux_analysis/sampling.py
@@ -33,6 +33,10 @@ feasibility_tol = bounds_tol
 nproj = 1000000
 """Reproject the solution into the feasibility space every nproj iterations."""
 
+nproj_center = 10000
+"""Reproject the center into the nullspace every nproj_center iterations.
+   Only used for inhomogeneous problems."""
+
 Problem = namedtuple("Problem",
                      ["equalities", "b", "inequalities", "bounds",
                       "variable_fixed", "variable_bounds", "nullspace",
@@ -66,7 +70,7 @@ def mp_init(obj):
     sampler = obj
 
 
-def shared_np_array(shape, data=None):
+def shared_np_array(shape, data=None, integer=False):
     """Create a new numpy array that resides in shared memory.
 
     Parameters
@@ -75,10 +79,17 @@ def shared_np_array(shape, data=None):
         The shape of the new array.
     data : numpy.array
         Data to copy to the new array. Has to have the same shape.
+    integer : boolean
+        Whether to use an integer array. Defaults to False which means
+        float array.
     """
     size = np.prod(shape)
-    array = Array(ctypes.c_double, int(size))
-    np_array = np.frombuffer(array.get_obj())
+    if integer:
+        array = Array(ctypes.c_int64, int(size))
+        np_array = np.frombuffer(array.get_obj(), dtype="int64")
+    else:
+        array = Array(ctypes.c_double, int(size))
+        np_array = np.frombuffer(array.get_obj())
     np_array = np_array.reshape(shape)
 
     if data is not None:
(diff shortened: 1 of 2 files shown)


config.sh
```diff
@@ -43,6 +43,8 @@ function run_tests_in_repo {
 		export PATH=$PWD:$PATH
 		# which pkg-config
 		# pip install matplotlib
+        yum install git
+        git --version
```
Midnighter: Which code down the line needs git? In general I'd prefer to pip install from archives rather than cloning repos (if possible).

hredestig: It's a requirement of pytest-benchmark apparently :/

cdiener:
Please don't merge yet, still debugging another issue...

cdiener:
Okay, should be done now. Results that there was a much simpler solution for inhomogeneous sampling :sweat_smile:

cdiener:
Yes, ready to merge :)

PR #565, opened 2017-08-04
removing solution links to model

pstjohn:
We talked about this a bit, but `solution.reactions[0].model` points back to the original model that generated the solution.

Do we really need these class members? Other than [this](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/test_solver_model.py#L39) line, everything else seems to pass fine when I remove them.

Thoughts on whether we want this? Let me know if I've missed anything?


hredestig:
Thanks for this!

PR #566, opened 2017-08-04
summary now accepts solution kwarg, resolves if missing

### LINKED ISSUE #520: Summary method after pfba

pstjohn:
This is a PR to fix #520 and de-couple the summary methods from connection via `reaction.flux` to the underlying solver in memory.

`model.summary()` now accepts a kwarg `solution`, which if given will re-use the fluxes from a previous solution. Otherwise puts in a call to `model.optimize()` to get an updated set of fluxes. Speeds don't seem to be affected that much, especially if the solver is able to re-use the solution basis.

### DIFF
--- a/cobra/core/metabolite.py
+++ b/cobra/core/metabolite.py
@@ -216,10 +216,13 @@ class Metabolite(Species):
 
         Parameters
         ----------
-        solution: cobra.core.solution
+        solution : cobra.core.Solution
             A previously solved model solution to use for generating the
             summary. If none provided (default), the summary method will
-            resolve the model.
+            resolve the model. Note that the solution object must match the
+            model, i.e., changes to the model such as changed bounds,
+            added or removed reactions are not taken into account by this
+            method.
 
         threshold : float
             a value below which to ignore reaction fluxes
--- a/cobra/core/metabolite.py
+++ b/cobra/core/metabolite.py
@@ -206,8 +206,8 @@ class Metabolite(Species):
         """
         self._model.remove_metabolites(self, destructive)
 
-    def summary(self, solution=None, threshold=0.01, fva=False, floatfmt='.3g',
-                **kwargs):
+    def summary(self, solution=None, threshold=0.01, fva=False,
+                floatfmt='.3g'):
         """Print a summary of the reactions which produce and consume this
         metabolite.
 
@@ -237,7 +237,7 @@ class Metabolite(Species):
         """
         from cobra.flux_analysis.summary import metabolite_summary
         return metabolite_summary(self, solution=solution, threshold=threshold,
-                                  fva=fva, floatfmt=floatfmt, **kwargs)
+                                  fva=fva, floatfmt=floatfmt)
 
     def _repr_html_(self):
         return """
--- a/cobra/core/metabolite.py
+++ b/cobra/core/metabolite.py
@@ -206,7 +206,8 @@ class Metabolite(Species):
         """
         self._model.remove_metabolites(self, destructive)
 
-    def summary(self, threshold=0.01, fva=False, floatfmt='.3g', **kwargs):
+    def summary(self, solution=None, threshold=0.01, fva=False, floatfmt='.3g',
+                **kwargs):
         """Print a summary of the reactions which produce and consume this
         metabolite.
 
@@ -215,6 +216,11 @@ class Metabolite(Species):
 
         Parameters
         ----------
+        solution: cobra.core.solution
+            A previously solved model solution to use for generating the
+            summary. If none provided (default), the summary method will
+            resolve the model.
+
         threshold : float
             a value below which to ignore reaction fluxes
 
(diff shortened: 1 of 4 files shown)


cobra/core/metabolite.py
```diff
@@ -206,7 +206,8 @@ def remove_from_model(self, destructive=False):
         """
         self._model.remove_metabolites(self, destructive)
 
-    def summary(self, threshold=0.01, fva=False, floatfmt='.3g', **kwargs):
+    def summary(self, solution=None, threshold=0.01, fva=False, floatfmt='.3g',
+                **kwargs):
```
hredestig: Suggest to remove the `**kwargs` here as only really relevant for the legacy solvers and those are not supported anyway

cobra/flux_analysis/summary.py
```diff
@@ -19,29 +19,39 @@ def format_long_string(string, max_length):
...
+        Whether or not to include flux variability analysis in the output.
+        If given, fva should be a float between 0 and 1, representing the
+        fraction of the optimum objective to be searched.
+
+    floatfmt : string
         format method for floats, passed to tabulate. Default is '.3g'.
 
     """
+    if solution is None:
+        solution = met.model.optimize(reactions=met.reactions, **solver_args)
```
hredestig: `Model.optimize` doesn't use the reactions parameter so perhaps change this to 
```
met.model.slim_optimize(error_value=None)
solution = get_solution(reactions=met.reactions)
```

cobra/flux_analysis/summary.py
```diff
@@ -129,23 +146,29 @@ def model_summary(model, threshold=1E-8, fva=None, floatfmt='.3g',
         raise NotImplementedError(
```
hredestig: I think can drop the `solver` , `solver_args` parameter and this exception altogether since we anyway introduce changes in behaviour

cobra/flux_analysis/summary.py
```diff
@@ -129,23 +146,29 @@ def model_summary(model, threshold=1E-8, fva=None, floatfmt='.3g',
...
             "Summary support for legacy solvers was removed.")
 
-    # Create a dataframe of objective fluxes
     objective_reactions = linear_reaction_coefficients(model)
-    obj_fluxes = pd.DataFrame({key: key.flux * value for key, value in
-                               iteritems(objective_reactions)},
+    boundary_reactions = model.exchanges
+
+    if solution is None:
+        solution = model.optimize(
```
hredestig: same as above, better use `slim_optimize` and `get_solution`

pstjohn:
Awesome, thanks for the comments -- agreed on all those points.

PR #587, opened 2017-08-29
refactor: transform production envelope to wide format with multiple carbon sources

### LINKED ISSUE #582: ZeroDivisionError in production_envelope
#### Problem description

Tried to compute a production envelope with the yeast 7.6 model but got a `ZeroDivisionError`. We should guard against that and make the result either `0`, `inf`, or `nan`. It works in cameo where this is being done.

#### Code Sample

```python
import cobra.flux_analysis as cflux
yeast7 = read_sbml_model("yeast_7.6_cobra.xml.gz")
biomass = yeast7.reactions.r_2111
c_source = yeast7.reactions.r_1714
target = yeast7.reactions.r_1546
plane = cflux.production_envelope(yeast7, [biomass], objective=target, c_source=c_source)
```

#### Actual Output

```python
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-16-160ec91a503e> in <module>()
----> 1 plane = cflux.production_envelope(yeast7, [biomass], objective=target, c_source=c_source)

cobra/flux_analysis/phenotype_phase_plane.py in production_envelope(model, reactions, objective, c_source, points, solver)
    391                          points, endpoint=True) for rxn in reactions]
    392         grid_list = list(product(*grid))
--> 393         result = envelope_for_points(model, reactions, grid_list, carbon_io)
    394 
    395     return pd.DataFrame(result)

cobra/flux_analysis/phenotype_phase_plane.py in envelope_for_points(model, reactions, grid, carbon_io)
    413                     if carbon_io[0] is not None:
    414                         results['carbon_yield'].append(carbon_yield(carbon_io))
--> 415                         results['mass_yield'].append(mass_yield(carbon_io))
    416     for key, value in results.items():
    417         results[key] = array(value)

cobra/flux_analysis/phenotype_phase_plane.py in mass_yield(c_input_output)
    466     except ValueError:
    467         return nan
--> 468     mol_prod_mol_src = product_flux / source_flux
    469     x = mol_prod_mol_src * c_product.formula_weight
    470     return x / c_source.formula_weight

ZeroDivisionError: float division by zero
```

#### Expected Output

Should not raise an exception.

#### Output of `cobra.show_versions()`

<details>

System Information
==================
OS                        Linux
OS-release     4.4.0-92-generic
Python                    3.5.2

Package Versions
================
pip                       9.0.1
setuptools               36.2.7
cobra                     0.8.2
future                   0.16.0
swiglpk                   1.4.4
optlang                   1.2.1
ruamel.yaml             0.14.12
pandas                   0.20.3
numpy                    1.13.1
tabulate                  0.7.7
python-libsbml           5.15.0
lxml                      3.8.0
scipy                    0.19.1
matplotlib                2.0.2
palettable                3.0.0

</details>

Midnighter:
fix #582 

So far this PR re-organizes the code for `production_envelope` a little bit and catches the potential `ZeroDivisionError` in `mass_yield`. However, I suggest the following further changes which is why I marked this PR as WIP.

1. Allow specification of more than one carbon source.
2. Transform the resulting data frame into wide format (the way it's done in cameo) meaning that maximum and minimum are separate columns.
3. Use the configured solver threshold to convert below threshold values to zero.

### DIFF
--- a/cobra/flux_analysis/phenotype_phase_plane.py
+++ b/cobra/flux_analysis/phenotype_phase_plane.py
@@ -3,17 +3,15 @@
 from __future__ import absolute_import, division
 
 from warnings import warn
-from collections import defaultdict
-from operator import itemgetter
-from itertools import product
 from six import iteritems
+from itertools import product
 from multiprocessing import Pool
+
 import pandas as pd
 from optlang.interface import OPTIMAL
-
 from numpy import (
     nan, abs, arange, dtype, empty, int32, linspace, meshgrid, unravel_index,
-    zeros, array)
+    zeros, full)
 
 from cobra.solvers import get_solver_name, solver_dict
 import cobra.util.solver as sutil
@@ -307,17 +305,18 @@ def calculate_phenotype_phase_plane(
     return data
 
 
-def production_envelope(model, reactions, objective=None, c_source=None,
+def production_envelope(model, reactions, objective=None, carbon_sources=None,
                         points=20, solver=None):
     """Calculate the objective value conditioned on all combinations of
     fluxes for a set of chosen reactions
 
-    The production envelope can be used to analyze a models ability to
+    The production envelope can be used to analyze a model's ability to
     produce a given compound conditional on the fluxes for another set of
-    reaction, such as the uptake rates. The model is alternately optimize
-    with respect to minimizing and maximizing the objective and record the
-    obtained fluxes. Ranges to compute production is set to the effective
-    bounds, i.e. the minimum / maximum fluxes that can be obtained given
+    reactions, such as the uptake rates. The model is alternately optimized
+    with respect to minimizing and maximizing the objective and the
+    obtained fluxes are recorded. Ranges to compute production is set to the
+    effective
+    bounds, i.e., the minimum / maximum fluxes that can be obtained given
     current reaction bounds.
 
     Parameters
@@ -325,21 +324,22 @@ def production_envelope(model, reactions, objective=None, c_source=None,
     model : cobra.Model
         The model to compute the production envelope for.
     reactions : list or string
-        A list of reactions, reaction identifiers or single reaction
-    objective : string, dict, model.solver.interface.Objective
+        A list of reactions, reaction identifiers or a single reaction.
+    objective : string, dict, model.solver.interface.Objective, optional
         The objective (reaction) to use for the production envelope. Use the
-        model's current objective is left missing.
-    c_source : cobra.Reaction or string
-       A reaction or reaction identifier that is the source of carbon for
-       computing carbon (mol carbon in output over mol carbon in input) and
-       mass yield (gram product over gram output). Only objectives with a
-       carbon containing input and output metabolite is supported.
-    points : int
+        model's current objective if left missing.
+    carbon_sources : list or string, optional
+       One or more reactions or reaction identifiers that are the source of
+       carbon for computing carbon (mol carbon in output over mol carbon in
+       input) and mass yield (gram product over gram output). Only objectives
+       with a carbon containing input and output metabolite is supported.
+       Will identify active carbon sources in the medium if none are specified.
+    points : int, optional
        The number of points to calculate production for.
-    solver : string
+    solver : string, optional
        The solver to use - only here for consistency with older
        implementations (this argument will be removed in the future). The
-       solver should be set using `model.solver` directly. Only optlang
+       solver should be set using ``model.solver`` directly. Only optlang
        based solvers are supported.
 
     Returns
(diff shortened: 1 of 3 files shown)


hredestig:
On 2, being used to ggplot's interface, I find long format much more convenient as the column names remain fixed and then easier to write general plotting code. But, I guess that may be somewhat niche case and that wide format feels more familiar.

Midnighter:
Actually, even for ggplot2 I find the wide variant more convenient because I opted to use `geom_ribbon`.

```R
ggplot2::ggplot(total_df, ggplot2::aes(x = biomass,
                                       ymin = flux_minimum,
                                       ymax = flux_maximum,
                                       color = label,
                                       fill = label)) +
  ggplot2::geom_ribbon(alpha=0.3) +
  ggplot2::scale_color_brewer(name = "Model", palette = "Set1") +
  ggplot2::scale_fill_brewer(name = "Model", palette = "Set1") +
  ggplot2::labs(x = expression(paste("Growth [", g[DW], h^{-1}, "]")),
                y = expression(paste("Flux [", mmol, g[DW]^{-1}, h^{-1}, "]")))
```

Which can produce something like the following plot, for example.

file:///home/moritz/Projects/fumarate/reports/figures/tmyl_flux_ppp.png

hredestig:
Sure that works as well, I find it easiest to know the right format when having tried to implement many different plots which I didn't do yet (e.g. to recycle code with plotting e.g. fva, however that should be done the best way.). Something to think of when implementing overall better plot functions for cobra. In the mean time, switch to wide format here totally fine with me.

Midnighter:
1. [x]
2. [x]
3. [ ]

Any other ideas that should be included here?

cobra/flux_analysis/phenotype_phase_plane.py
```diff
@@ -374,168 +372,201 @@ def production_envelope(model, reactions, objective=None, c_source=None,
...
-                    if carbon_io[0] is not None:
-                        results['carbon_yield'].append(carbon_yield(carbon_io))
-                        results['mass_yield'].append(mass_yield(carbon_io))
-    for key, value in results.items():
-        results[key] = array(value)
-    if carbon_io[0] is not None:
-        results['carbon_source'] = carbon_io[0].id
-    return results
+        with model:
+            model.solver.objective.direction = sense
```
hredestig: This doesn't use context manager, perhaps better formulate as `model.objective = ..` ? Otherwise I guess will exit with objective changed to max if starting with min..

cobra/flux_analysis/phenotype_phase_plane.py
```diff
@@ -374,168 +372,201 @@ def production_envelope(model, reactions, objective=None, c_source=None,
...
+            for i in range(len(grid)):
+                with model:
+                    for rxn in reactions:
+                        point = grid.at[i, rxn.id]
+                        rxn.bounds = point, point
+                    model.slim_optimize()
```
hredestig: perhaps `obj_value = model.slim_optimize()` and then use `obj_value` below

cobra/flux_analysis/phenotype_phase_plane.py
```diff
@@ -374,168 +372,201 @@ def production_envelope(model, reactions, objective=None, c_source=None,
...
+                    if model.solver.status != OPTIMAL:
+                        continue
+                    if abs(model.solver.objective.value) < \
```
hredestig: how about making the tolerance an optional argument and have thereby have this clipping documented?

cobra/flux_analysis/phenotype_phase_plane.py
```diff
@@ -374,168 +372,201 @@ def production_envelope(model, reactions, objective=None, c_source=None,
...
+        output_flux, output_elements, consumption=False)
     try:
         return carbon_output_flux / carbon_input_flux
     except ZeroDivisionError:
         return nan
 
 
-def mass_yield(c_input_output):
-    """Gram product divided by gram of carbon input source
+def reaction_components(reaction, atom='C'):
```
hredestig: `element` not atom.. And since you add this, perhaps do it all the way and allow user to calculate any element yield by having e.g. `yield_element='N'` as argument to `production_envelope`? Or do later and remove parametrisation here..

cobra/flux_analysis/phenotype_phase_plane.py
Midnighter: Agreed.

cobra/flux_analysis/phenotype_phase_plane.py
Midnighter: Good idea.

cobra/flux_analysis/phenotype_phase_plane.py
Midnighter: Removing parametrization here for now. It's a little annoying with column naming so unless there is a need for it I'd skip it.

Midnighter:
This PR is now based on #588 and the new `Model.objective_direction` property.

cobra/flux_analysis/phenotype_phase_plane.py
```diff
@@ -374,168 +374,193 @@ def production_envelope(model, reactions, objective=None, c_source=None,
...
+                    obj_val = model.slim_optimize()
+                    if model.solver.status != OPTIMAL:
+                        continue
+                    grid.at[i, 'flux_{}'.format(direction)] = \
+                        0.0 if abs(obj_val) < threshold else obj_val
+                    if c_input is not None:
+                        grid.at[i, 'carbon_yield_{}'.format(direction)] = \
+                            total_yield([rxn.flux for rxn in c_input],
+                                        input_components,
+                                        model.solver.objective.value,
```
hredestig: `obj_val` ?

Midnighter: Absolutely!

PR #588, opened 2017-09-12
feat: add revertible property objective_direction [WIP]

Midnighter:
Let me know if I should make use of this new property in other code in this PR or if we should slowly migrate code whenever/wherever we come across it.

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -979,21 +979,15 @@ class Model(Object):
         return self.solver.objective.direction
 
     @objective_direction.setter
+    @resettable
     def objective_direction(self, value):
         """"""
-        reverse_value = self.solver.objective.direction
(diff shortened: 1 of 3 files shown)


cobra/core/model.py
```diff
@@ -967,6 +967,34 @@ def objective(self, value):
...
+        Get or set the objective direction.
+
+        When using a `HistoryManager` context, this attribute can be set
+        temporarily, reversed when exiting the context.
+
+        """
+        return self.solver.objective.direction
+
+    @objective_direction.setter
+    def objective_direction(self, value):
```
pstjohn: would this work with the simple `@resettable` decorator?

Midnighter: Good call. It should indeed work.

PR #589, opened 2017-09-12
Add some helpers for media

cdiener:
This is a proposal to add a media module with a few helper functions for media. Major player is the `minimal_medium` function to calculate the smallest total import flux or smallest set of import reactions that allow growth at a specific rate.

This is still missing docs.

## Example:

```Python

In [3]: from cobra.media import *

In [4]: minimal_medium(mod, min_growth=0.8)
Out[4]: 
EX_glc__D_e    10.000000
EX_nh4_e        4.362240
EX_o2_e        18.579253
EX_pi_e         2.942960
dtype: float64

In [5]: minimal_medium(mod, min_growth=0.1, minimize_components=True)
Out[5]: 
EX_glc__D_e    10.000000
EX_nh4_e        1.042503
EX_pi_e         0.703318
dtype: float64

In [6]: # this is anaerobic growth :D
```

## Discussion points

- this also adds a new `exchanges` function that does some more work to only find exchanges and not demand and sink reactions, so should `model.exchanges` default to that one and `model.boundaries` use the old one?
- currently returns a pandas Series. Is that ok?

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -24,6 +24,7 @@ from cobra.util.solver import (
     add_cons_vars_to_problem, remove_cons_vars_from_problem, choose_solver,
     check_solver_status, assert_optimal)
 from cobra.util.util import AutoVivification, format_long_string
+from cobra.media import exchanges
 
 LOGGER = logging.getLogger(__name__)
 
@@ -233,18 +234,21 @@ class Model(Object):
             elif reaction.products:
                 reaction.upper_bound = bound
 
+        exchange_rxns = set(self.exchanges)
+
         # Set the given media bounds
         media_rxns = list()
         for rxn_id, bound in iteritems(medium):
             rxn = self.reactions.get_by_id(rxn_id)
+            if rxn not in exchange_rxns:
+                LOGGER.warn("%s does not seem to be an" % rxn.id +
+                            " an exchange reaction. Applying bounds anyway.")
             media_rxns.append(rxn)
             set_active_bound(rxn, bound)
-
-        boundary_rxns = set(self.exchanges)
         media_rxns = set(media_rxns)
 
         # Turn off reactions not present in media
-        for rxn in (boundary_rxns - media_rxns):
+        for rxn in (exchange_rxns - media_rxns):
             set_active_bound(rxn, 0)
 
     def __add__(self, other_model):
@@ -729,7 +733,16 @@ class Model(Object):
     def exchanges(self):
         """Exchange reactions in model.
 
-        Reactions that either don't have products or substrates.
+        Reactions that exchange mass with the exterior. Uses annotations
+        and heuristics to exclude non-exchanges such as sink reactions.
+        """
+        return exchanges(self, ext_compartment=None)
+
+    @property
+    def boundary(self):
+        """Boundary reactions in the model.
+
+        Reactions that either have no substrate or product.
         """
         return [rxn for rxn in self.reactions if rxn.boundary]
 
--- a/cobra/flux_analysis/loopless.py
+++ b/cobra/flux_analysis/loopless.py
@@ -9,8 +9,7 @@ from six import iteritems
 from optlang.symbolics import Zero
 
 from cobra.core import Metabolite, Reaction, get_solution
-from cobra.util import (linear_reaction_coefficients,
-                        create_stoichiometric_matrix, nullspace)
+from cobra.util import create_stoichiometric_matrix, nullspace
 from cobra.manipulation.modify import convert_to_irreversible
 
 
@@ -84,6 +83,7 @@ def add_loopless(model, zero_cutoff=1e-12):
 def _add_cycle_free(model, fluxes):
     """Add constraints for CycleFreeFlux."""
     model.objective = Zero
+    objective_vars = []
     for rxn in model.reactions:
         flux = fluxes[rxn.id]
         if rxn.boundary:
@@ -91,14 +91,13 @@ def _add_cycle_free(model, fluxes):
             continue
         if flux >= 0:
             rxn.lower_bound = max(0, rxn.lower_bound)
-            model.objective.set_linear_coefficients(
-                {rxn.forward_variable: 1, rxn.reverse_variable: -1})
+            objective_vars.append(rxn.forward_variable)
         else:
             rxn.upper_bound = min(0, rxn.upper_bound)
-            model.objective.set_linear_coefficients(
-                {rxn.forward_variable: -1, rxn.reverse_variable: 1})
+            objective_vars.append(rxn.reverse_variable)
 
-    model.solver.objective.direction = "min"
+    model.objective.set_linear_coefficients(dict.fromkeys(objective_vars, 1.0))
+    model.objective.direction = "min"
 
 
 def loopless_solution(model, fluxes=None):
@@ -117,10 +116,6 @@ def loopless_solution(model, fluxes=None):
         A dictionary {rxn_id: flux} that assigns a flux to each reaction. If
         not None will use the provided flux values to obtain a close loopless
         solution.
-        Note that this requires a linear objective function involving
-        only the model reactions. This is the case if
-        `linear_reaction_coefficients(model)` is a correct representation of
-        the objective.
 
     Returns
     -------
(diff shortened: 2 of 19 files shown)


Midnighter:
This looks really useful!

* I would argue for having four methods: `boundaries`, `exchanges`, `demands`, and `sinks`.
* I prefer a `pandas.DataFrame` that also includes the name (and ID) of the metabolite being exchanged. The name is important for models with non-intuitive identifiers.

joaocardoso:
Hi @cdiener!

This looks really nice. I would like to share a few ideas and thoughts:

1. You could add an integer cut to enumerate alternative medium compositions of the same size. There might be some alternative compositions.

2. I think the user should be able to set if oxygen is required or not.

3. Should we leave micronutrients free? Or make a mixed objective where phosphate, sulfur, carbon and a nitrogen sources have a binary indicator, but iron, magnesium, cobalt, etc. have a linear variable?

4. Can we state the mathematical formulation in the documentation?

Cheers,

Joao

cdiener:
@joaocardoso Thanks for the feedback.

1. sounds good, just how many should we allow. This would mean adding an additional constraint so that
    sum(previously_used_indicator) < opt
2. and 3. are difficult since all annotations are optional in cobrapy. If the model specifies its metabolites with IDs such as "m1234", "m12345" which one will be oxygen. All bounds are respected if you don't set `open_exchanges` so you can simply force a non-zero oxygen import by hand before calling the function. So it is already supported more or less.

4. Yes definitely. Still have to write the docs before merge.

Midnighter:
I think right now it's probably better to go with a `pandas.Series` as you proposed. It's compatible with the current interfaces. We can revisit this if we ever make the medium a `DataFrame` as is done in cameo.

joaocardoso:
@cdiener you're welcome!

1. We can set a `max_iteration' and run the loop until it reaches the maximum iteration or there are no more solutions. We can get the max size from the first solution.

So after the first iteration we just need add the sum(previously_used_indicator) < opt and sum(all_indicator) == opt.

The solution will be infeasible if there are no more alternatives.

2. This would be something we do when the metabolites have chemical formulas. That would work despite the identifiers.

cdiener:
Okay, I added the alternative solutions now. Seems to work but depends a lot on the integrality tolerance. Added a section to the notes that discusses that.

```Python
In [1]: from cobra.media import *

In [2]: from cobra.test import create_test_model

In [4]: minimal_medium(mod, 0.8, minimize_components=5, open_exchanges=True)
Out[4]:
                      0           1           2           3           4
EX_fru_e       0.000000  523.104557    0.000000    0.000000    0.000000
EX_glc__D_e    0.000000    0.000000  523.104557  521.357767  519.750758
EX_gln__L_e    0.000000    0.000000    0.000000   40.698058    0.000000
EX_glu__L_e  348.101944   83.995843   83.995843    0.000000    0.000000
EX_nh4_e       0.000000    0.000000    0.000000    0.000000   81.026921
EX_o2_e      500.000000    0.000000    0.000000    0.000000    0.000000
EX_pi_e       66.431529   56.667310   56.667310   54.913419   54.664344
```

@joaocardoso So what would happen if we would not have the formula? For now I think we could leave that to the user since it is pretty much straight forward and uses the same API as all other knockouts:

```Python

In [2]: from cobra.media import *

In [4]: minimal_medium(mod, 0.2)
Out[4]:
EX_glc__D_e    3.490332
EX_nh4_e       1.090560
EX_o2_e        5.460817
EX_pi_e        0.735740
dtype: float64

In [5]: mod.reactions.EX_o2_e.knock_out()

In [6]: minimal_medium(mod, 0.2)
Out[6]:
EX_glc__D_e    10.00000
EX_h2o_e        3.06276
EX_nh4_e        1.09056
EX_pi_e         0.73574
dtype: float64
```

cdiener:
Here is what is still missing:

1. Write the docs. 
2. Decide what should happen with the exchanges function and model.exchanges. I tend to having model.exchanges and model.boundary (or model.boundary_reactions).

cdiener:
Okay, added the docs. Can not really test them since I have not manged to build the docs locally (getting all kinds of errors). 

We could still change the behavior of `model.exchanges` if we want to...

Midnighter:
> We could still change the behavior of `model.exchanges` if we want to...

I think that would be nice. My preferred way would still be `boundary` is the superset of `exchanges`, `demands`, and `sinks`. In testing for the subsets I would first look for the relevant SBO terms and as a fallback assume the BiGG like identifiers `EX_`, `DM_`, and `SK_`.

I will take a look at locally building the docs and if that points to any serious issues.

cdiener:
> In testing for the subsets I would first look for the relevant SBO terms and as a fallback assume the BiGG like identifiers EX_, DM_, and SK_.

Sounds like a good idea. Do you have a link to a complete list of SBO terms I would need to check?

Midnighter:
We have the following in `cobra.manipulation.validate` but I haven't double checked them:

```
NOT_MASS_BALANCED_TERMS = {"SBO:0000627",  # EXCHANGE
                           "SBO:0000628",  # DEMAND
                           "SBO:0000629",  # BIOMASS
                           "SBO:0000631",  # PSEUDOREACTION
                           "SBO:0000632",  # SINK
                           }
```

Midnighter:
@cdiener could you also handle #628 in this PR?

cdiener:
@Midnighter yeah sure. For this it might make sense to also add a function (maybe `media.make_exchange`) that makes a a reaction a valid exchange reaction (for instance by setting the SBO term), so that a user might fix the warning by explicitly denoting a reaction as an exchange... What do you think?

Midnighter:
We should definitely add SBO terms to our `add_boundary` method but I wouldn't do it here. The way that #628 came up was that the user tried to take a transport reaction and add it to the medium. I think it would be a bad idea to allow the user to try and turn that into an exchange. It might even break the model.

cdiener:
Yeah, I meant apart from `model.media`. I also would not want that to convert everything to exchanges. I just want a way for the user to declare reactions as exchanges in case we do not identify them correctly as such.

Midnighter:
I see what you mean. In that case I find `mark_exchange` (and correspondingly, demand, sink, and biomass) clearer. I'd rather handle that in a different PR then because it seems to me to be more about annotation than the medium.

Midnighter:
I'm also not 100% sure about the module name/location. Are you planning a lot more functionality for media which is why you're starting a new sub-package? Otherwise my suggestion would be to put it in `cobra.manipulation.medium`.

cdiener:
At least I see it as a first class citizen. A lot of stuff in the manipulation module is deprecated by now so I was not sure whether it would be around in the long run...

cdiener:
Okay, added the SBO terms and the boundary/exchanges split in model. This is a backwards incompatible change.

Midnighter:
@cdiener would you mind a PR to this branch? I have a few small changes in mind and would like to adjust before approving this PR. Overall it seems very solid to me!

cdiener:
@Midnighter sure go ahead

cobra/core/model.py
```diff
@@ -729,7 +733,16 @@ def constraints(self):
     def exchanges(self):
         """Exchange reactions in model.
 
-        Reactions that either don't have products or substrates.
+        Reactions that exchange mass with the exterior. Uses annotations
+        and heuristics to exclude non-exchanges such as sink reactions.
+        """
+        return exchanges(self, ext_compartment=None)
```
Midnighter: I would like to add corresponding properties for demands and sinks.

cdiener: As single one for both? Or a separate one for each?

cobra/core/model.py
```diff
@@ -233,18 +234,21 @@ def set_active_bound(reaction, bound):
...
                 reaction.upper_bound = bound
 
+        exchange_rxns = set(self.exchanges)
+
         # Set the given media bounds
         media_rxns = list()
         for rxn_id, bound in iteritems(medium):
             rxn = self.reactions.get_by_id(rxn_id)
+            if rxn not in exchange_rxns:
+                LOGGER.warn("%s does not seem to be an" % rxn.id +
```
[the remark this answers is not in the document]
cdiener: Will do

cobra/core/model.py
```diff
@@ -233,18 +234,21 @@ def set_active_bound(reaction, bound):
             elif reaction.products:
                 reaction.upper_bound = bound
 
+        exchange_rxns = set(self.exchanges)
```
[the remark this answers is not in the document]
cdiener: Will do

cobra/core/model.py
Midnighter: A separate one for each.

cdiener:
Closing in favor of a new clean PR. The discussion will be linked there.

Issue #590, opened 2017-09-13
Add the method for loopless FVA from cameo


Midnighter:
cameo uses a different method for removing cycles in the FVA result. I ran this on iJR904 and cobra takes 3 min whereas cameo 15 s.

cdiener:
They use exactly the same method, but I will look into what is making the cobrapy version so slow...

cdiener:
Okay, I checked it and there is actually something weird going on in cameo. For now cameo only checks reactions for cycles if the flux is larger than 666 (?). For the textbook model that only applies to 4 reactions. Obviously that will not work for any model which as maximum bounds smaller than that and ,even if not, it is not guaranteed to remove cycles at all due to that restriction. Calling it directly without the sloppy argument, ergo `_cycle_free_fva(mod, sloppy=False)` is actually slower than the cobrapy implementation...

Midnighter:
@phantomas1234 can you comment on this condition? It seems rather odd to check for diabolic flux bounds.

phantomas1234:
I am sorry for the satanic and arbitrary threshold here. I'll need to check my implementation, but if I remember correctly what I did was to introduce  `sloppy` to check if reactions are in cycles only if they have a large flux (to speed up cyclefreeflux). Setting `sloppy=False` makes it indeed very slow (I never really understood why tbh). Doesn't `sloppy` make sense though? Wouldn't most cycles have very large fluxes? Also, I've yet to see a model to have default bounds < 1000 (I agree though that sloopy should probably be based on some actual bound in the model). The textbook model has only one cycle and it is detected even when `sloppy=True`. @cdiener are you referring to the toy example on the docs regarding the cycle that is not close to the reaction bounds?

cdiener:
Thanks for the reply @phantomas1234 :)

I agree that not many models have flux bounds different from 1000 but I think it would be a pretty strong assumption to make about any model. For instance models constrained with fluxome data would deviate from that assumption... 

[...]

PR #594, opened 2017-09-13
Several small adjustments to loopless

cdiener:
Removes the requirement of the fixed objective. Now gives the closest objective value without loops as cameo does. This also make loopless FVA a bit faster. Still pretty slow for large models, but the limiting factor here is changing the objective and getting the solution.

### DIFF
--- a/cobra/flux_analysis/loopless.py
+++ b/cobra/flux_analysis/loopless.py
@@ -9,8 +9,7 @@ from six import iteritems
 from optlang.symbolics import Zero
 
 from cobra.core import Metabolite, Reaction, get_solution
-from cobra.util import (linear_reaction_coefficients,
-                        create_stoichiometric_matrix, nullspace)
+from cobra.util import create_stoichiometric_matrix, nullspace
 from cobra.manipulation.modify import convert_to_irreversible
 
 
@@ -84,6 +83,7 @@ def add_loopless(model, zero_cutoff=1e-12):
 def _add_cycle_free(model, fluxes):
     """Add constraints for CycleFreeFlux."""
     model.objective = Zero
+    objective_vars = []
     for rxn in model.reactions:
         flux = fluxes[rxn.id]
         if rxn.boundary:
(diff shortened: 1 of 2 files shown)


Midnighter:
I have a novice question regarding the use of lower level optlang interfaces. Doesn't that circumvent the reversibility of the context introduced before? That's one reason I created `Model.objective_direction` but now I'm starting to think we should build the context into optlang?

cdiener:
So one cheat I use is setting the objective to zero in the context with `model.objective = S.Zero` than modify it only with the low level optlang interface. This way it is reset, but only once.

cdiener:
Hi, @Midnighter was there still something you would like added here?

Midnighter:
The changes itself look good to me but the following still took 22 minutes for me:

```python

if __name__ == "__main__":
    model = read_sbml_model("iJO1366.xml.gz")
    model.solver = "glpk"
    fva_result_loopless = flux_variability_analysis(model, loopless=True)
```

With CPLEX I killed it after 2 hrs. Didn't try Gurobi yet. I can merge these changes but then would like to find the best implementation between cobrapy and cameo (fully acknowledging that the implementation in cameo may be faulty).

cdiener:
Yeah I agree that it is still slow but I don't see how to make it faster without dropping down to C. The cameo implementation essentially does very little (only checks a few select reactions). You can emulate that behavior trivially with cobrapy by just passing a few reactions to fva. 

Are you saying that clpk takes 22 min but cplex 2h for you? That sounds weird. For me the loopless FVA runs in about the same time with glpk and cplex...

Midnighter:
> Are you saying that clpk takes 22 min but cplex 2h for you? That is suprising. For me the loopless FVA runs in about the same time with glpk and cplex...

Strange indeed. I should probably double check my cplex installation.

cdiener:
> Strange indeed. I should probably double check my cplex installation.

If it is reproducible that would indeed be a bug in cobrapy or optlang but would probably have to be addressed in another PR...

phantomas1234:
Hi, see also my response to https://github.com/opencobra/cobrapy/issues/590. I never really understood why cyclefreeflux is so slow in reality (at least in my implementation, maybe one should benchmark against the R implementation). On paper it looks very fast. Is it even faster than MILP loopless?

cdiener:
@Midnighter ok, done.

PR #596, opened 2017-09-20
adding reaction `__radd__`

pstjohn:
Are we still supporting these overloaded methods? If we are, this is a pretty basic change that essentially lets you `sum` an iterable of reactions.

Python's `sum` starts with a 0, and then adds each element in turn. So previously this would fail with
```python
>>> sum(model.metabolites.etoh_c.reactions)
TypeError: unsupported operand type(s) for +: 'int' and 'Reaction'
```

This PR
1) allows an `__radd__` function and
2) explicitly allows `reaction + 0`

So now you can 
```python
>>> sum(model.metabolites.etoh_c.reactions).reaction
'etoh_e + h_e + nad_c <=> acald_c + 2.0 h_c + nadh_c'
```

### DIFF
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -608,8 +608,14 @@ class Reaction(Object):
 
         """
         new_reaction = self.copy()
-        new_reaction += other
+        if other == 0:
+            return new_reaction
+        else:
+            new_reaction += other
+
         return new_reaction
+
+    __radd__ = __add__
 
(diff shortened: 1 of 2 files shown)


pstjohn:
Yeah, probably a bit of a niche case. In my particular use, I wanted to get lumped reaction expressions for a whole pathway. So I'd just build up a list of the reaction objects and `sum` them all together.

PR #600, opened 2017-09-26
[website] Use new builtin syntax highlighting in Hugo

cdiener:
Hugo now uses [chroma](https://github.com/alecthomas/chroma) for native syntax highlighting so we can get rid of prism. Also updates the publication list with 3 recent ones.

If you would like another [theme](https://help.farbox.com/pygments.html) for the syntax highlighting please comment.

Preview: https://opencobra.github.io/cobrapy-website/


Midnighter:
I'm a big fan of solarized light/dark code themes. Will see if someone made that (was not on the page you linked).

cdiener:
Yeah chroma only supports the ones [here](https://github.com/alecthomas/chroma/tree/master/styles). There is monokai if you like that one...

pstjohn:
unrelated but we could probably remove the `model.optimize()` line now that solution solves the model automatically.

I did like the background coloring of the previous version, maybe `friendly` would replicate that?

cdiener:
@pstjohn you mean `summary` solves the model automatically, right?

If it is only the background color I think I can force a different one with CSS as well...

pstjohn:
oops, right - summary

Midnighter:
Is this an option? https://pypi.python.org/pypi/pygments-style-solarized

cdiener:
@Midnighter hugo does not use Pygments anymore (really slow) but Chroma which is a port of pygments to Go. So you need the style in chroma to work.

PR #616, opened 2017-10-26
feat: sort out the ordering issues with round tripping models

Midnighter:
Currently, the JSON and YAML writers (inheriting from dict) sort the model attributes. This may cause confusion and is unexpected behaviour. This PR aims to improve that situation and starts out with a series of clear test scenarios.

### DIFF
--- a/cobra/test/test_io_order.py
+++ b/cobra/test/test_io_order.py
@@ -72,8 +72,10 @@ def get_ids(iterable):
 
 @pytest.mark.parametrize("read, write, ext", [
     ("read_sbml_model", "write_sbml_model", ".xml"),
-    ("read_legacy_sbml", "write_legacy_sbml", ".xml"),
-    ("load_matlab_model", "save_matlab_model", ".mat"),
+    pytest.mark.skip(("read_legacy_sbml", "write_legacy_sbml", ".xml"),
+                     reason="Order for legacy SBML I/O is uninteresting."),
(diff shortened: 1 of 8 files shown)


cobra/test/test_io_order.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.fixture(scope="module")
+def minimized_core():
+    model = cio.read_sbml_model(join(dirname(__file__), "data",
+                                "textbook.xml.gz"))
+    model.id = "minimized_core"
+    chosen = sample(model.reactions, 10)
+    model.remove_reactions(set(model.reactions).difference(chosen),
+                           remove_orphans=True)
```
pstjohn: Don't you risk removing all the metabolites if the 10 chosen reactions don't overlap?

Midnighter: The metabolites of those reactions should remain, right? That seems like enough metabolites to play with.

pstjohn: Oh I guess you're right, nevermind then, yeah that's probably enough

Midnighter:
Let's see what the tests say but I consider this done. What I finally did:
* Introduce tests
* Make sorting of metabolites, reactions, and genes optional
* Do sort additional attributes such as annotations and notes

Midnighter:
Had to pin the scipy version since the sparse matrices changed quite a bit with the release of 1.0.

PR #619, opened 2017-10-30
Skip tests for ArrayBasedModel with scipy>=1.0.0

cdiener:
Workaround in order to not have to pin the scipy version which may cause problems. ArrayBasedModel is deprecated anyways...

### DIFF
--- a/cobra/test/test_model.py
+++ b/cobra/test/test_model.py
@@ -932,7 +932,8 @@ class TestCobraModel:
 class TestStoichiometricMatrix:
     """Test the simple replacement for ArrayBasedModel"""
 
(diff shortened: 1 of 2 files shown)


Midnighter:
My preferred way is to remove the deprecated parts rather than skipping the tests. What do you think?

cdiener:
You mean only the tests or the entire ArrayBasedModel? Either one is fine with me but the second option requires a new minor release version (backwards incompatible).

Midnighter:
I meant the entire `ArrayBasedModel`. Let's chat about it briefly during the call but there will be a few minor releases coming up.

cdiener:
Closing in favor of removing `ArrayBasedModek`.

PR #624, opened 2017-11-03
Double deletion

sgalkina:


cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+
+def defaultdict_to_dict(def_dict):
+    if not isinstance(def_dict, defaultdict):
+        return def_dict
+    for k, v in def_dict.items():
+        def_dict[k] = defaultdict_to_dict(v)
+    return dict(def_dict)
+
+
+def _reactions_knockouts_with_restore(model, reactions):
```
Midnighter: The 'restore' functionality is exactly provided by the model context manager. Any model modifications are reversed upon leaving the context. You can find examples [here](http://cobrapy.readthedocs.io/en/latest/getting_started.html?highlight=context#Making-changes-reversibly-using-models-as-contexts).

sgalkina: that's true, though the speed issue was exactly here: it was much faster in my benchmarks to restore the simple case of few reactions fluxes like that than to use the context manager

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+                (err_type, err_val, err_tb) = sys.exc_info()
+                raise_(err_type, err_val, err_tb)  # reraise for Python2&3
+
+        if legacy:
+            raise ValueError(
+                "Legacy solvers are not supported any longer. Please use one of"
+                " the optlang solver interfaces instead.")
+
+        if num_cpu is None:
+            num_cpu = kwargs.get("number_of_processes")
```
Midnighter: I think extracting arguments from the `kwargs` is a leftover from the old code and not very elegant. I'd rather make everything proper arguments and get rid of `kwargs` entirely.

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # Reactions with 0 flux in the current conditions have no impact on
+        # deletion. Determine those reactions now to speed up computation later.
+        if method == 'moma':
+            moma.add_moma(model)
+        if method == 'linear moma':
+            moma.add_moma(model, linear=True)
+        solution = model.optimize()
+        if solution.status == "optimal":
+            kwargs["growth"] = solution.objective_value
```
Midnighter: Same as above about `kwargs`.

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+            kwargs["zero_flux_reactions"] = \
+                {rxn for (rxn, flux) in solution.fluxes.items()
```
Midnighter: Please import `iteritems` from `six` and use that where possible.

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+        for comb in product(*element_lists):
+            growth = knockout_results[unordered_combination(comb)]
+            current = double
+            for v in comb[:-1]:
+                current = current[v.id]
+            current[comb[-1].id] = growth
+        return defaultdict_to_dict(double)
+
+
+def _element_lists(entities, *ids):
```
Midnighter: In the various functions you're converting between identifiers and objects quite a lot (it seems). Is it possible to be more consistent, for example, work with identifiers most of the time unless reactions are needed for knockouts?

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+            kwargs["zero_flux_reactions"] = \
+                {rxn for (rxn, flux) in solution.fluxes.items()
+                 if abs(flux) < zero_cutoff}
+        else:
+            warn("Non-optimal ({0:s}) base solution.".format(solution.status))
+
+        def unordered_combination(reactions):
+            return frozenset([i.id for i in reactions if i.id not in kwargs["zero_flux_reactions"]])
+
+        args = set([unordered_combination(comb) for comb in product(*element_lists)])
```
Midnighter: One strategy I rather liked in the old implementation was that not only are the `"zero_flux_reactions"` uninteresting for further computations but it's also clear that any double (multiple) knockout combination involving an essential reaction (from the single knockouts) will be essential. I don't know how much you gain from that in practice but it seems like the computation of a lot of combinations could be avoided that way.

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+
+    Returns
+    -------
+    dict of dict
+        A sparse representation of all combinations of pairs of reaction
+        deletions without replacements.
+    """
+    with cobra_model as model:
+        WORKERS = dict(gene=_gene_deletion_worker, reaction=_reaction_deletion_worker)
+        worker_function = WORKERS[entity]
```
Midnighter: I would skip assigning the `dict` to anything. i.e., `worker_function = dict(...)[entity]`.

cobra/flux_analysis/deletion.py
cdiener: It should not really make a difference if you use nested contexts, like in the current implementation:

```Python
        with cobra_model as m:
            m.solver = solver
            for gene in gene_list:
                with m:
                    gene.knock_out()
                    obj_value = m.slim_optimize()
                    status_dict[gene.id] = m.solver.status
                    growth_rate_dict[gene.id] = obj_value
```

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+def _reactions_knockouts_with_restore(model, reactions):
+    to_restore = {}
+    growth = None
+    for reaction in reactions:
+        to_restore[reaction] = reaction.bounds
+        reaction.knock_out()
+    try:
+        solution = model.optimize()
```
cdiener: This should be `slim_optimize` if you only want the growth rate. Will be much faster.

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+
+        def unordered_combination(reactions):
+            return frozenset([i.id for i in reactions if i.id not in kwargs["zero_flux_reactions"]])
+
+        args = set([unordered_combination(comb) for comb in product(*element_lists)])
+        worker_function.model = model
+        if num_cpu > 1:
+            chunk_size = len(args) // num_cpu
+            pool = multiprocessing.Pool(num_cpu)
+            results = pool.imap_unordered(worker_function, args, chunksize=chunk_size)
```
cdiener: I think the pool has to be terminated or closed after running the map. Otherwise you will spawn a lot of zombie processes.

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+        model,
+        [model.reactions.get_by_id(r_id) for r_id in ids]
+    )
+
+
+def _gene_deletion_worker(ids):
+    model = _gene_deletion_worker.model
+    all_reactions = []
+    for g_id in ids:
+        all_reactions.extend(find_gene_knockout_reactions(model, (model.genes.get_by_id(g_id),)))
```
cdiener: This will be slow. Preferably `find_gene_knockout_reactions` should be called once before the map step with all genes that are knocked out.

cdiener:
Hi, great API :) Some random comments:

When parallelizing the deletions `Pool` will pickle the model and copy it across the processes. With optlang this will be pretty slow so there is a big performance hit there. Also the when unpickling any previous solution basis will be lost so the first solve on every process starts *de novo* which is also slower. So I would expect that this will only be faster than a single process for many knock outs.

The zero flux exclusion is a great idea but has some traps. Due to floating point issues having a flux <1e-12 often means it is effectively zero, however that is not guaranteed. Depending on how unstable the model is it can happen that a flux has a value of 1e-12 in the maximization but upon setting it to zero the growth rate drops to zero (flux is essential but with a tiny value). So there is some danger there. 

For the double deletion one does not only have to avoid pairs where one of the reactions is essential but also the case where pairs have been calculated previously. For instance if you calculated the knock out for `(r1, r2)` there is no point in also calculating it for `(r2, r1)`. So if you imagine the double knockouts as a rn x rn table you only need to calculate the upper or lower triangular matrix corresponding to that.

In case you accept PRs to the double_deletion branch I could also send my suggestions that way.

Cheers

cobra/flux_analysis/deletion.py
sgalkina: the current interface for the knockout functions allows to accept both ids and objects, and this ambiguity propagated down here. But I agree there are maybe some unnecessary ones that could be cleaned up

cobra/flux_analysis/deletion.py
sgalkina: but to calculate essentiality, wouldn't you need to perform single knockouts for all the reactions and optimise? I think it'll bring more of extra optimisations to the process which can slow it down

sgalkina:
thank you @Midnighter and @cdiener for the good comments, I'm working on them now.
couple of comments from my side

> So I would expect that this will only be faster than a single process for many knock outs.
since model is pickled only once per process, and the number of processes is usually ~4, it is still pretty fast. And speeding it up only makes sense for many knockouts indeed, for a few knockouts speed is not the issue anyway. Maybe it even makes sense to have a threshold for a number of `args` below which just one process should be used.

>The zero flux exclusion is a great idea but has some traps.
thanks for noticing that, do you think the smaller threshold of f.e  <1e-15 would make sense?

> For the double deletion one does not only have to avoid pairs where one of the reactions is essential but also the case where pairs have been calculated previously
It is already done with capturing the required ids with `frozenset`, so `(r1, r2)` and `(r2, r1)` are the same and duplications in `args` are thus removed

@cdiener you can definitely send prs to this branch if you feel like it, though the comments are also fine

cobra/flux_analysis/deletion.py
Midnighter: Yes, you would have to do that. Which is why I also suggested the more iterative approach in my review comment. I'm not sure the strategy is better. The problem is also that it's, of course, model dependent whether it's worth doing or not. Let *N* be the set of reactions with non-zero flux. Let *n* be the cardinality |*N*|. Let *K* be the set of essential reactions with *K* being a subset of *N*. Let *k* be the cardinality |*K*|. Let *C* be the binomial coefficient. Then it's worth doing this if *k* < *C*(*n*^2, 2) - *C*((*n* - *k*)^2, 2).

Midnighter:
> Also the when unpickling any previous solution basis will be lost so the first solve on every process starts de novo which is also slower.

@cdiener can we set a basis on the workers somehow so that the solvers can hot start?

cobra/flux_analysis/deletion.py
```diff
@@ -0,0 +1,348 @@
...
+    to_restore = {}
+    growth = None
+    for reaction in reactions:
+        to_restore[reaction] = reaction.bounds
+        reaction.knock_out()
+    try:
+        solution = model.optimize()
+        if solution.status != 'infeasible':
+            growth = solution.objective_value
+    except:
```
Midnighter: Once you use `slim_optimize` I would not expect an error here. If anything something more specific should be caught such as `cobra.exceptions.OptimizationError` or `optlang.SolverError`.

cdiener:
Thanks for the quick reply @sgalkina!

> since model is pickled only once per process, and the number of processes is usually ~4, it is still pretty fast. And speeding it up only makes sense for many knockouts indeed, for a few knockouts speed is not the issue anyway. Maybe it even makes sense to have a threshold for a number of args below which just one process should be used.

Yes that is true. The factor seems to be about 1:50 for me (resolving vs. pickling). I think it is sufficient to have the user choose the number of processes (maybe call it `n_jobs` as in scipy) and have a section in the docs that recommends only using more processes if there are more than 100 reactions to be checked or something like that.

> thanks for noticing that, do you think the smaller threshold of f.e <1e-15 would make sense?

I don't think so. It is more an issue of the solver tolerance. So any exactly zero flux will give you a value in [tol, -tol] where "tol" is the solver tolerance (I think 1e-6 by default). The problem is that a flux with an exact solution of 1e-12 will give you any value in [-tol + 1e-12, tol + 1e-12]. Those intervals overlap so there is no way to tell whether a flux is really zero using a non-exact solver. So there is always a chance you wrongly call a flux as not having impact on the growth rate even so there is. How much that happens will depend on the model you use. I think in the end we would have to prioritize slow correct results over fast incorrect ones and do what everybody else does and run all deletions. For the double deletion case you can still save some time by first running all single deletions and excluding the ones that give zero or near zero *growth* (since pairing those with any other will not drop the growth rate further).

> It is already done with capturing the required ids with frozenset, so (r1, r2) and (r2, r1) are the same and duplications in args are thus removed

So sorry, you are absolutely correct :(

Looks like it is in really good hands though. Looking forward to having the new API merged.

Cheers

Midnighter:
Looks great now. Thank you for all the hard work and bearing with our change requests.

cdiener:
Are there some benchmarks to verify it didn't get slower?

Midnighter:
Hmm, not sure how systematic they are but @sgalkina was comparing her implementation to the previous one.

cdiener:
Did a quick check and speed looks good. There are few issues though, especially for several cases I now get incorrect results. For instance the MOMA code gives incorrect results for infeasible solutions and the code does not seems to work with symengine.

PR #640, opened 2017-12-10
refactor: use the optlang interface clone method [question]

Midnighter:
So I played around a bit with addressing [this problem](https://github.com/opencobra/memote/issues/262) in memote. One way of doing so is by using `optlang.interface.Model.clone` instead of a `copy` or `deepcopy`.

That causes two problems:
1. The times are abysmal in comparison or the names might change.
```python
In [5]: from cobra.io import read_sbml_model   

In [6]: m = read_sbml_model("iJO1366.xml.gz")  

In [7]: %timeit m._solver.clone(m._solver)     
1.35 s ± 6.56 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)                           

In [8]: %timeit m._solver.clone(m._solver, use_json=False)                                    
4.62 s ± 14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)                             

In [9]: %timeit m._solver.clone(m._solver, use_json=False, use_lp=True)                       
/home/moritz/.virtualenvs/testcopy/lib/python3.6/site-packages/optlang/interface.py:1096 UserWarning: Cloning with LP formats can change variable and constraint ID's.                      
291 ms ± 5.55 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [10]: %timeit deepcopy(m._solver)
286 ms ± 3.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

2. It fails one of our tests `test_fixed_seed` because cloning the solver model rather than copying changes the order of variables.

I'm also not sure what really causes this. I was not able to reproduce it with below code. The situation in memote is slightly more complicated but I have made certain that in the fixture in memote the model has variables before copying and no variables on the copy (GLPK), whereas with CPLEX it copies the variables just fine.

```python
import pytest

MODEL = read_sbml_model("iJO1366.xml.gz")


@pytest.fixture(scope="session")
def read_only_model():
    return MODEL


@pytest.fixture(scope="function")
def model(read_only_model):
    return read_only_model.copy()


def test_copy_has_variables(model):
    assert len(model.variables) > 0


def test_copy_has_same_number_variables(read_only_model, model):
    assert len(read_only_model.variables) == len(model.variables)


def test_copy_has_same_variables(read_only_model, model):
    assert set(v.name.split("_rev")[0] for v in read_only_model.variables) == \
        set(v.name.split("_rev")[0] for v in model.variables)
```

I'm not suggesting to merge this but I wanted a place to discuss this issue and make the necessary changes (if any). @KristianJensen any idea under what circumstances the variables of the GLPK interface would not be copied?

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -319,11 +319,8 @@ class Model(Object):
                 new_gene = new.genes.get_by_id(gene.id)
                 new_reaction._genes.add(new_gene)
                 new_gene._reaction.add(new_reaction)
-        try:
-            new._solver = deepcopy(self.solver)
-            # Cplex has an issue with deep copies
-        except Exception:  # pragma: no cover
-            new._solver = copy(self.solver)  # pragma: no cover
+
+        new._solver = self._solver.clone(self._solver)
 
         # it doesn't make sense to retain the context of a copied model so
         # assign a new empty context


KristianJensen:
Cloning the solver should not change the order of the variables. With which cloning method(s) does this happen?
In cameo we used to copy models all the time and I don't think we had any problems with variables disappearing. With that said I'm not sure how reliably copy/deepcopy works with c-extensions, so I wouldn't be surprised if there were issues with that. Maybe it would be possible to implement a `__copy__` method in swiglpk. That would be the best way to deal with it I guess.

Midnighter:
> Cloning the solver should not change the order of the variables. With which cloning method(s) does this happen?

I haven't fully checked this. `test_fixed_seed` samples from the model with a seed using `random.sample`. It started failing when I introduced cloning (with the default `use_json=True` argument) so I'm assuming the order changes.

> In cameo we used to copy models all the time and I don't think we had any problems with variables disappearing.

I'm also very puzzled why this happens. I cannot reproduce it in any other scenario but reliably in memote.

cdiener:
AFAIK `test_fixed_seed` does not depend directly on the order of variables but it requires that given a particular problem the solver returns exactly the same solution even if there are many possible solutions. I guess that this could depend on the ordering of the variables (simplex should be sensitive to that). In your examples I think `[9]` and `[10]` are doing the same thing...

Those methods might potentially get faster if the intermediate representation is written to RAM rather than an intermediate file as is done now...

Midnighter:
I admit I didn't look into `test_fixed_seed` very thoroughly.

> In your examples I think [9] and [10] are doing the same thing...

The timings are very similar but they should not do the same thing at all. Indeed having a `__copy__` method in swiglpk sounds very desirable. Is that a quick fix @phantomas1234 or something more involved?

cdiener:
> The timings are very similar but they should not do the same thing at all.

Last time I checked the optlang code both wrote the model to a `*.lp` representation and read it again from there (possibly with a different interface).

Midnighter:
> Last time I checked the optlang code both wrote the model to a *.lp representation and read it again from there (possibly with a different interface).

So you're saying `deepcopy` writes to `.lp` as well? Sounds very inefficient but sure that could be the implementation.

cdiener:
I think so at least. Porting all variables and constraints between interfaces by hand is much slower since the direct writing and reading from the LP format is usually implemented directly in C. Normal deepcopy would not work since the solver objects are C objects (pointers) so Python does not know how to copy those....

KristianJensen:
Yes that is true. Copying of optlang models is implemented through the `__getstate__` and `__setstate__` methods (same as are used for pickling), which in the case of GLPK uses an lp serialization as intermediate. The implementation of the GLPK `glp_read_prob` method requires a filename to read from, so currently the lp format is actually being written to a temporary file. However a quick profiling of GLPK model copying shows that 244 ms out of 264 ms is actually spent constructing the model (variable and constraint objects) after the lp file has been read, so the file writing/reading is not the bottleneck.

KristianJensen:
@Midnighter I think I have a hunch what might be going on. The `glp_write_prob` and `glp_read_prob` functions that are used to copy the problem can fail silently, with a non-zero return value. Currently the return values of these functions are not checked, so if the writing and/or reading fails for whatever reason, the copied problem will simply be an emtpy newly initialized problem. Can you verify that this is what is going on, by adding a check in either of `__getstate__` or `__setstate__` of `optlang.glpk_interface`. According to the GLPK manual it should also print an error message, but I don't think that gets propagated by swig.

If this is the problem, that still leaves the question of why the writing/reading fails of course...

Midnighter:
A concrete hint 😺 I'll check it out.

cdiener:
Thanks for the explanation @KristianJensen :D I can confirm that swiglpk will not print the error messages since many glpk print statements are removed in it's setup.py due to a compatibility problem.

Midnighter:
So thanks to some more hints from @KristianJensen and lots of fun with the debugger I found that `swiglpk.glp_write_prob` was writing floats with a Danish decimal sign but `swiglpk.glp_read_prob` is unaware of the locale and cannot read that file.

@phantomas1234 I'll create a test scenario for swiglpk but I'm not sure at which level this problem is occurring right now.

I'm closing this PR since it is clearly nothing cobrapy related.

cdiener:
Awesome. Nice detective work.

PR #642, opened 2017-12-18
docs: update readme

Midnighter:
* Remove waffle
* Fix PyPI (why were we linking to cameo? :P)
* Fix AppVeyor badge
* style


Midnighter:
Failing codecov for an edit to the README doesn't make any sense to me.

cdiener:
@Midnighter that's from the sampling code. There is a part of the code that sometimes fixes bounds violations if they arise due to numerical issues, but it did not have to do that in this particular run. This is essentially random so I would ignore that for now...

PR #643, opened 2017-12-18
fix: drop and warn Python 3.4

Midnighter:
(@cdiener and @pstjohn I'm not waiting the customary 24 h since this is breaking all our builds. I'm happy to revert the changes later on.)


pstjohn:
No worries, that's fine. Thanks for being on top of this!

cdiener:
Sure, good call :)

Issue #655, opened 2018-01-26
Error when removing and re-adding reaction with context manager


pstjohn:
When removing a reaction from a model inside a context, it appears the reaction is added back to the model but the underlying solver objects are mangled.

```python
import cobra.test
model = cobra.test.create_test_model('textbook')

sol = model.optimize()
model.summary(sol)
print('PDH_FLUX: {}'.format(sol.fluxes['PDH']))
print()

with model:
    model.reactions.PDH.remove_from_model()
    sol = model.optimize()
    model.summary(sol)
    print()

sol = model.optimize()
model.summary(sol)
print('PDH_FLUX: {}'.format(sol.fluxes['PDH']))
```

```
IN FLUXES        OUT FLUXES    OBJECTIVES
---------------  ------------  ----------------------
o2_e      21.8   h2o_e  29.2   Biomass_Ecol...  0.874
glc__D_e  10     co2_e  22.8
nh4_e      4.77  h_e    17.5
pi_e       3.21
PDH_FLUX: 9.282532599166627

IN FLUXES        OUT FLUXES    OBJECTIVES
---------------  ------------  ----------------------
o2_e      21.3   h2o_e  24.2   Biomass_Ecol...  0.797
glc__D_e  10     h_e    23.7
nh4_e      4.34  co2_e  18.4
pi_e       2.93  for_e   7.74

IN FLUXES        OUT FLUXES    OBJECTIVES
---------------  ------------  ----------------------
o2_e      21.3   h2o_e  24.2   Biomass_Ecol...  0.797
glc__D_e  10     h_e    23.7
nh4_e      4.34  co2_e  18.4
pi_e       2.93  for_e   7.74
PDH_FLUX: 0.0
```

Midnighter:
Definitely unintended behavior. Is there any particular reason, though, why you are removing the reaction rather than knocking it out? I'm fairly sure that knock-outs are reverted as intended.

pstjohn:
Indeed, knockouts are handled fine. This was more just an easy example to show the issue.
For me, it's because I'm passing the model to a visualization library and don't want the reactions to show up :). But more generally, if its a feature we want to support we should probably make sure it works correctly. I wonder if it was a recent break or if contexted model additions / subtractions never got added back to the solver?

pstjohn:
So the context manager is adding the variables back to model.solver.variables... 

```python
In [4]: id(model.reactions.PDH.forward_variable)
Out[4]: 4703595688

[...]

PR #656, opened 2018-01-27
attempts to make reversible reaction deletion work

pstjohn:
Some in-progress attempts at fixing `model.remove_reactions` and `model._populate_solver`. Still not working.

Addresses #655

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -615,6 +615,8 @@ class Model(Object):
                 reaction._model = None
 
                 if context:
+                    context(self.solver.update)
+                    context(partial(self._populate_solver, [reaction]))
                     context(partial(setattr, reaction, '_model', self))
                     context(partial(self.reactions.add, reaction))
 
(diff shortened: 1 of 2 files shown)


cdiener:
The problem is at https://github.com/opencobra/cobrapy/blob/76ed164d4fada06395a8fc9be80e6bedde825c8e/cobra/core/model.py#L613 . This will already trigger a re-add of the variables on context exit. But `_populate_solver` is also called on exit and also adds the variables. Maybe adding `_populate_solver` to the context *before*  the call to `remove_cons_vars` works.

pstjohn:
Thanks for the tip! This works locally, lets see what CI says.

Midnighter:
@pstjohn are the calls to `self.solver.update()` strictly necessary? I'm asking because updating the solver after every change defeats the purpose of collecting them and then doing batch modifications.

pstjohn:
Doesn't look like it, good call.

pstjohn:
@Midnighter, this is RTM from my end. Any other thoughts?

cobra/core/model.py
```diff
@@ -745,16 +747,24 @@ def _populate_solver(self, reaction_list, metabolite_list=None):
...
+                    reaction.id, lb=forward_lb, ub=forward_ub)
+                reverse_variable = self.problem.Variable(
+                    reaction.reverse_id, lb=reverse_lb, ub=reverse_ub)
 
-            forward_variable = self.problem.Variable(
-                reaction.id, lb=forward_lb, ub=forward_ub)
-            reverse_variable = self.problem.Variable(
-                reaction.reverse_id, lb=reverse_lb, ub=reverse_ub)
+                self.add_cons_vars([forward_variable, reverse_variable])
+                self.solver.update()
```
Midnighter: This one is necessary?

Midnighter:
If you can comment on that one line, I'm happy to merge.

pstjohn:
Oh, sorry, thought you meant the context'ed model.solver.update. No, I don't think we need that line. I'll delete it

pstjohn:
should fix #661 as well

PR #658, opened 2018-01-31
refactor: add or change `num_cpu` argument

Midnighter:
The `find_essential_genes` and reactions functions did not have an argument to set the number of processes thus the underlying functions were always executed with multiprocessing. This PR fixes that. I have (re-)named the argument `num_cpu` everywhere. I found `num_jobs` slightly confusing since it's about the number of processes and not jobs. I chose `num_cpu` since it is closer to `os.` or `multiprocessing.cpu_count` but would be fine with `num_proc`, too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+.pytest_cache/
 nosetests.xml
 coverage.xml
 *,cover
--- a/cobra/flux_analysis/deletion.py
+++ b/cobra/flux_analysis/deletion.py
@@ -71,7 +71,7 @@ def _init_worker(model):
     _model = model
 
 
-def _multi_deletion(model, entity, element_lists, method="fba", num_cpu=None):
+def _multi_deletion(model, entity, element_lists, method="fba", num_proc=None):
     """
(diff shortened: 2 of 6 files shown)


cdiener:
I think it would be good to take this occasion to homogenize the naming of that particular argument the sampling code as well where it is still called `np` (number of processes). An alternative suggestion would be `num_workers`. `num_cpu` is also fine with we even though I can run our codes with `num_cpu=8` even though I only have 2 CPUs :smile:

Midnighter:
I agree that `num_cpu` is not ideal either. A common name is also `num_threads` although that is more on the C level and should not be confused with Python threading. So my personal favorites are:
1. `num_proc`
2. `num_workers`

cdiener:
Both of those are very nice in my opinion :+1:

pstjohn:
scikit-learn and joblib use `n_jobs`, pymc3 uses `njobs`. In case you didn't have enough options 😄

Midnighter:
Went with `num_proc`. Seems the most accurate to me. Not a fan of jobs (😝) and worker like an async setting.

Midnighter:
Alright, @phantomas1234 made a good point that the argument in multiprocessing itself is actually `processes` so I went with that.

PR #664, opened 2018-02-05
fix: support non-mapped optlang exception

joaocardoso:
By using a default dict, all other exceptions will return an generic OptimizationError


joaocardoso:
I'll submit a better implementation

PR #665, opened 2018-02-05
Fix: capture all optlang exceptions

joaocardoso:
Optlang returns more solver status than the ones mapped in the current implementation.

To address that problem, we raise a cobrapy specific exception if it is mapped or a OptimizationError.

We include the solver status in the message to provide the user with a cause for failure.


Midnighter:
@cdiener & @pstjohn: Sorry for not waiting the customary 24 h. It's a small change but we need it in order to re-submit the cameo paper today.

PR #672, opened 2018-02-27
Stabilize sampling

cdiener:
Yet again, several small fixes and improvements. There are limitations though, so I also added a section to the docs how to filter samples for feasible ones.

## Fixed issues:

- correctly samples from 2D problems
- redundant/duplicated search directions and zero directions are now removed from the warmup
- throws an error for completely fixed problems
- stabilized the online mean calculation (still numerically unstable though)
- sampling will no longer throw a RecursionError when getting stuck but throws an informative error
- reprojection frequency is now set via a heuristic to improve stability for small models but maintain speed for larger ones
- samplers gain the `nproj` and `retries` attributes to monitor stability problems


cdiener:
[And here is the benchmark.](https://github.com/cdiener/cobra-docker/blob/master/sampling_benchmark.ipynb)

Midnighter:
I don't have a problem to test this on but the change set looks good to me. My only problem is why that first commit f26802a is in there?

cdiener:
Yeah no idea. Do you have some pointers on how to fix that, because I am not sure how to do that...

Midnighter:
`git rebase -i HEAD~5` and then drop that commit should fit it.

cdiener:
Hmm, a lot of stuff seems broken now. Will need more time to fix all of that.

Midnighter:
Were the tests never deterministic or did this now uncover a problem with not updating the model excessively?

cdiener:
More that I got sent another model to test and this uncovered some problems in how I select the sampling directions/warmup points.

cdiener:
Ok, looks like it might work now, but I think it would help if someone could test it some more.

PR #677, opened 2018-03-13
Quick edit to reaction.add_metabolites

pstjohn:
Previously calling `model.reactions.rxn1 + model.reactions.rxn2` would cause the metabolites from rxn2 to be added to the new lumped reaction, while the lumped reaction didn't belong to any model.

This ensures that metabolites being added to a reaction belong to the same model, or are duplicated prior to adding.

@cdiener, locally I'm getting an error with the sampling code based on the fixed seed. Any idea why that might be? Should I just change the test to reflect the new value?

```python
    def test_fixed_seed(self, model):
        s = sample(model, 1, seed=42)
>       assert numpy.allclose(s.TPI[0], 9.12037487)
E       assert False
E        +  where False = <function allclose at 0x1159ace18>(9.002806727659744, 9.12037487)
E        +    where <function allclose at 0x1159ace18> = numpy.allclose

cobra/test/test_flux_analysis.py:762: AssertionError
```

### DIFF
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -743,8 +743,8 @@ class Reaction(Object):
             # Make sure metabolites being added belong to the same model, or
             # else copy them.
             if isinstance(metabolite, Metabolite):
-                if ((metabolite.model is not None) &
-                        (metabolite.model != self._model)):
+                if ((metabolite.model is not None) and
+                        (metabolite.model is not self._model)):
                     metabolite = metabolite.copy()
 
             met_id = str(metabolite)
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -739,6 +739,14 @@ class Reaction(Object):
         _id_to_metabolites = dict([(x.id, x) for x in self._metabolites])
 
         for metabolite, coefficient in iteritems(metabolites_to_add):
+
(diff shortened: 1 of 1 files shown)


pstjohn:
Hmm, guess that doesn't complain on the CI tests. Nevermind then on the sampling error

cobra/core/reaction.py
```diff
@@ -739,6 +739,14 @@ def add_metabolites(self, metabolites_to_add, combine=True,
         _id_to_metabolites = dict([(x.id, x) for x in self._metabolites])
 
         for metabolite, coefficient in iteritems(metabolites_to_add):
+
+            # Make sure metabolites being added belong to the same model, or
+            # else copy them.
+            if isinstance(metabolite, Metabolite):
+                if ((metabolite.model is not None) &
+                        (metabolite.model != self._model)):
```
Midnighter: Pardon my nitpicking. Can you turn the `&` into an `and`, please. In the second comparison we're not interested in model (in-)equality but identity, right? So `metabolite.model is not self._model` seems more idiomatic to me.

cdiener:
@pstjohn Everytime we change something in the setup code of the solver the seed test has to be adjusted (since the result will be different now). You probably used some branch which did not include the latest sampler changes.

PR #679, opened 2018-03-18
add room

synchon:
- is very slow when simulating large models like iAF1260

- LP variant stops to function due to infeasibility arising in the model when optimizing

- tests and documentation to be added

### DIFF
--- a/cobra/flux_analysis/room.py
+++ b/cobra/flux_analysis/room.py
@@ -22,10 +22,6 @@ def add_room(model, solution=None, delta=0.03, epsilon=0.001, linear=False):
     linear : bool
         Whether to use the linear ROOM formulation or not.
 
-    Returns
-    -------
-    Nothing.
-
     Notes
     -----
     The formulation used here is the same as stated in the original paper.
@@ -44,7 +40,7 @@ def add_room(model, solution=None, delta=0.03, epsilon=0.001, linear=False):
          w_i^l = w_i - \delta|w_i| - \epsilon
 
     So, for the linear version of the ROOM , constraint [3] is relaxed to
-    0 <=y_i <= 1.
+    0 <= y_i <= 1.
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


cobra/flux_analysis/room.py
```diff
@@ -0,0 +1,77 @@
...
+    if 'room_old_objective' in model.solver.variables:
+        raise ValueError('model is already adjusted for ROOM')
+
+    # optimizes if no reference solution is provided
+    if solution is None:
+        solution = model.optimize()
+
+    problem = model.problem
+    variable = problem.Variable("room_old_objective")
+    constraint = problem.Constraint(
```
cdiener: This does not yet constrain the old objective. For that you still need to set bounds on `variable` :)

cdiener:
Since the default formulation is a MILP the slowness is to be expected, so no worries there.

cobra/flux_analysis/room.py
synchon: Okay! I will fix it.

synchon: But, when I compare it to the cameo implementation, it's still slow by almost an order of 4. And, also what might be the reason that the LP variant is causing infeasibility in the model?

cdiener: Is that using symengine or sympy? Yeah there should not be an infeasibility for the linear version, is that with the fix for the growth rate constraint? Otherwise the problem might be the unbounded variable.

synchon: I have set the upper bound of `variable` equal to `solution.objective_value` and it gave me the same result.
I used sympy. Even after making the variable bounded, the linear variant still gives infeasibility. I tried relaxing the growth rate constraint too but it doesn't help.

Midnighter:
Just did some quick timings with and without symegine.
* sympy: `4min 56s ± 7.87 s per loop (mean ± std. dev. of 7 runs, 1 loop each)`
* symengine: `4.13 s ± 27.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
So almost two orders of magnitude.

Midnighter:
Superseded by #708.

PR #680, opened 2018-03-20
Drop usage of unsupported pip api

kvikshaug:
[From pip v9.0.2, importing pip after certain packages, like requests, fails](https://github.com/pypa/pip/issues/5079). That can cause problems with applications using cobrapy. Additionally, pip now [makes it pretty clear that the internal api is not supported](https://mail.python.org/pipermail/distutils-sig/2017-October/031642.html).

This PR removes package print output from `show_versions` and instead encourages issue reporters to include the output of `pip list`.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -20,9 +20,9 @@ Create a [minimal, complete, verifiable example
 
 #### Expected Output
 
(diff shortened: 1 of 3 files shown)


Midnighter:
I agree, #697 is the better solution to the problem. Thank you for spotting this so early, though. We should have been warned :/

PR #687, opened 2018-04-05
Gapfill fix to support new metabolites

maureencarey:
Updated flux_analysis/gapfilling to support gapfill solutions that add metabolites; tests modified to highlight this case. So, if a gapfill solution adds a reaction that contains metabolites not in the original model, corresponding demand or other reactions will also be added to clear the new metabolites. (Note: several additional commits to test with travis, not to address the gapfill modifications.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,13 @@ branches:
  only:
  - master
  - devel
- - gapfill_demand_fix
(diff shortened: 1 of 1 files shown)


maureencarey:
Bug addressed in Gapfill bugfix #744

PR #689, opened 2018-04-06
Changes summaries to use a fractional threshold rather than absolute

pstjohn:
This generalizes better to models that have a wide range of different flux values.


Midnighter:
Can you rebase your branch on the latest devel, please? That should have taken care of the AppVeyor problems.

PR #697, opened 2018-04-23
Repair Appveyor

cdiener:
So this is an attempt to repair appveyor. 

I suspect the problem is that conda has not yet updated pip to version 10. So even if you update pip with conda you get pip 9. this triggers a warning from pip to stderr which appveyor interprets as failure (wrongly so).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ environment:
 
   global:
     PIP_CACHE_DIR: "pip_cache"
+    PIP_DISABLE_PIP_VERSION_CHECK: "yes"
 
   matrix:
     - PYTHON: "C:\\Miniconda-x64"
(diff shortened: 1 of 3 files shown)


cdiener:
Ok, looks like it works.

appveyor.yml
```diff
@@ -47,10 +47,10 @@ install:
       if ($env:CONDA -eq "true") {
         conda config --set always_yes yes --set changeps1 no;
         conda install -q pip setuptools wheel numpy scipy;
-        pip install -r develop-requirements.txt;
-        pip install .
+        pip install --disable-pip-version-check -r develop-requirements.txt;
```
Midnighter: Rather than setting these options on each command I prefer to set an environment variable `PIP_DISABLE_PIP_VERSION_CHECK=yes`.

cobra/util/version_info.py
```diff
@@ -55,7 +55,8 @@ def get_pkg_info():
     # using requirements files that can be read in.
     dependencies = frozenset(PKG_ORDER)
     blob = dict()
-    for dist in pip.get_installed_distributions():
+    dists = [d for d in pkg_resources.working_set]
```
Midnighter: Since `dists` is not used for anything else can you combine these two lines to:

```python
for dist in pkg_resources.working_set:
```
please?

appveyor.yml
Midnighter: In fact, I just set this variable through the interface so you don't have to worry about it. These options can simply be removed.

cdiener: Would it bother if I still set the environment variable in the config? Just so that it is explicit and people could fork the config and it would still work.

Midnighter: Good idea, please do.

Issue #698, opened 2018-04-27
Loopless FVA produces odd results


tpfau:
#### Loopless FVA produces odd results

Please explain:
Using the very simple model [attached](https://github.com/opencobra/cobrapy/files/1954149/Test.xml.zip) I tried to run loopless FVA. What I got was all 0 fluxes except for the exchange reactions. I assume this is due to the use of CycleFreeFVA to achieve looplessness. I think the reason is that an objective coefficient of 0 is incompatible with CycleFreeFVA, as that was what I originally used. This is due to a zero coefficient leading to the CycleFreeFVA solution being always zero for reactions which can have any flux in a cycle.  
However, when using a higher objective coefficient, I get non zero flux through R1 and  R2 due to the forced flux through the objective reaction. In fact, the "cycling" reactions (R2 and R3) are then forced to carry flux depending on the coefficient, while the exchange reactions are pushed towards 0.
Not sure, but this looks pretty odd to me.

#### Code Sample
```python
model = cobra.io.read_sbml_model( "Test.xml")
A:
>>> fva = flux_variability_analysis(model, loopless=True, fraction_of_optimum=0)
B:
>>> fva = flux_variability_analysis(model, loopless=True, fraction_of_optimum=0.001)
C:
>>> fva = flux_variability_analysis(model, loopless=True, fraction_of_optimum=1)
```
#### Actual Output
```
A:
>>> fva
    maximum  minimum
R1      0.0      0.0
R2      0.0      0.0
R3      0.0      0.0
EA   1000.0      0.0
EB   1000.0      0.0
B:
>>> fva
    maximum  minimum
R1   1000.0      1.0
R2      1.0      1.0
R3      1.0      1.0
EA    999.0      0.0
EB    999.0      0.0

C:
>>> fva
    maximum  minimum
R1   1000.0   1000.0
R2   1000.0   1000.0
R3   1000.0   1000.0
EA      0.0      0.0
EB      0.0      0.0
```

#### Expected Output
```
A:
>>> fva
    maximum  minimum
R1    1000.0      0.0
R2      0.0      0.0
R3      0.0      0.0
EA   1000.0      0.0
EB   1000.0      0.0
B:
>>> fva
    maximum  minimum
R1   1000.0      1.0
R2      0.0      0.0
R3      0.0      0.0
EA    1000.0      1.0
EB    1000.0      1.0

C:
>>> fva
    maximum  minimum
R1   1000.0   1000.0
R2      0.0      0.0
[...]

PR #699, opened 2018-04-27
Fix a regression bug in loopless FVA

cdiener:
Used the loopless objective as the target flux. Also there was a missing with block which would not reset the exchange fluxes correctly. Some other small fixes that do not actually affect anything but make the method more explicit.

### DIFF
--- a/cobra/flux_analysis/loopless.py
+++ b/cobra/flux_analysis/loopless.py
@@ -228,7 +228,6 @@ def loopless_fva_iter(model, reaction, solution=False, zero_cutoff=1e-6):
         reaction.bounds = bounds
         # find the reactions with loops using the current reaction and remove
         # the loops
-        print(ll_sol, almost_ll_sol)
         for rxn in model.reactions:
             rid = rxn.id
             if ((abs(ll_sol[rid]) < zero_cutoff) and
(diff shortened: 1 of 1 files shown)


Midnighter:
Partially fix #698

cobra/flux_analysis/loopless.py
```diff
@@ -221,11 +221,11 @@ def loopless_fva_iter(model, reaction, solution=False, zero_cutoff=1e-6):
         # almost loopless solution containing only loops including the current
         # reaction. Than remove all of those loops.
         ll_sol = get_solution(model).fluxes
-        bounds = reaction.bounds
         reaction.bounds = (current, current)
         model.slim_optimize()
         almost_ll_sol = get_solution(model).fluxes
-        reaction.bounds = bounds
+        print(ll_sol, almost_ll_sol)
```
Midnighter: Could you remove this again, please? I suppose it was for debugging. Otherwise maybe introduce a logger here.

cdiener: Ouch, sorry. fixed now.

cdiener:
Well, I'm the one that broke it in the first place :sweat_smile:

PR #704, opened 2018-05-11
refactor: simplify dependency information

Midnighter:
Simplify the dependency information by depending on an external package.


Midnighter:
Merged without review due to being trivial.

PR #705, opened 2018-05-11
fix: repair AppVeyor builds

Midnighter:
Probably the first of many attempts to fix the AppVeyor builds.


Midnighter:
I merged this since it doesn't touch any code and will restore AppVeyor builds immediately.

PR #708, opened 2018-05-17
Added ROOM with tests from original paper

synchon:
Updated version of #679. Separate PRs due to commit conflict issues. Needs performance improvement when using Sympy. Symengine works perfectly.

### DIFF
--- a/cobra/flux_analysis/deletion.py
+++ b/cobra/flux_analysis/deletion.py
@@ -13,6 +13,7 @@ import pandas as pd
 from cobra.manipulation.delete import find_gene_knockout_reactions
 import cobra.util.solver as sutil
 from cobra.flux_analysis.moma import add_moma
+from cobra.flux_analysis.room import add_room
 
 LOGGER = logging.getLogger(__name__)
 
@@ -72,7 +73,7 @@ def _init_worker(model):
 
 
 def _multi_deletion(model, entity, element_lists, method="fba",
-                    solution=None, processes=None):
+                    solution=None, processes=None, **kwargs):
     """
     Provide a common interface for single or multiple knockouts.
 
@@ -93,6 +94,13 @@ def _multi_deletion(model, entity, element_lists, method="fba",
         The number of parallel processes to run. Can speed up the computations
         if the number of knockouts to perform is large. If not passed,
         will be set to the number of CPUs found.
+    **kwargs:
+        delta: float
+            The relative tolerance for ROOM.
+            Default is 0.03.
+        epsilon: float
+            The absolute tolerance for ROOM.
+            Default is 0.001.
 
     Returns
     -------
@@ -123,6 +131,15 @@ def _multi_deletion(model, entity, element_lists, method="fba",
     with model:
         if "moma" in method:
             add_moma(model, solution=solution, linear="linear" in method)
+        if "room" in method:
+            if kwargs:
+                delta = float(kwargs['delta'])
+                epsilon = float(kwargs['epsilon'])
+            else:
+                delta = 0.03
+                epsilon = 0.001
+            add_room(model, solution=solution, linear="linear" in method,
+                     delta=delta, epsilon=epsilon)
 
         args = set([frozenset(comb) for comb in product(*element_lists)])
         processes = min(processes, len(args))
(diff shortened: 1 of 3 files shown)


cobra/test/test_flux_analysis.py
```diff
@@ -61,6 +63,65 @@ def ll_test_model(request):
...
+def construct_papin_2003_model():
+    test_model = Model('papin_2003')
+    test_model.add_metabolites(Metabolite("A"))
+    test_model.add_metabolites(Metabolite("B"))
+    test_model.add_metabolites(Metabolite("C"))
+    test_model.add_metabolites(Metabolite("D"))
+    test_model.add_metabolites(Metabolite("E"))
+    test_model.add_metabolites(Metabolite("byp"))
+    test_model.add_metabolites(Metabolite("cof"))
+    v1 = Reaction("v1")
```
Midnighter: I think a less verbose way to construct the model would be:
```python
test_model = Model("papin_2003")
v1 = Reaction("v1")
# other reactions...
test_model.add_reactions([v1, ...])
v1.reaction = "A -> B"
v2.reaction = "2 B -> C + byp"
# etc.
```

Midnighter: This will automatically create the necessary metabolites.

cobra/test/test_flux_analysis.py
```diff
@@ -315,6 +376,25 @@ def test_single_gene_deletion_benchmark(self, model, benchmark,
         model.solver = solver
         benchmark(single_reaction_deletion, model=model, processes=1)
 
+    @pytest.mark.parametrize("solver", optlang_solvers)
+    def test_single_reaction_deletion_room(self, solver):
+        model = construct_papin_2003_model()
+        model.solver = solver
+        wt_sol = construct_papin_2003_solution()
```
Midnighter: You don't actually use the `wt_sol` and thus also don't need the `construct_papin_2003_solution` function.

cobra/test/test_flux_analysis.py
```diff
@@ -315,6 +376,25 @@ def test_single_gene_deletion_benchmark(self, model, benchmark,
...
+                           'b3': 5.0})
+        with model:
+            add_room(model, solution=wt_sol, delta=0.0, epsilon=0.0)
+            room_sol = model.optimize()
+        with model:
+            add_room(model, solution=wt_sol, delta=0.0, epsilon=0.0,
+                     linear=True)
+            linear_room_sol = model.optimize()
+
+        assert expected.eq(room_sol.fluxes).all()
```
Midnighter: Since we're not using exact solvers here it's more robust to say:
```python
assert numpy.allclose(room_sol.fluxes, expected)
```

cobra/test/test_flux_analysis.py
```diff
@@ -315,6 +376,25 @@ def test_single_gene_deletion_benchmark(self, model, benchmark,
         model.solver = solver
         benchmark(single_reaction_deletion, model=model, processes=1)
 
+    @pytest.mark.parametrize("solver", optlang_solvers)
+    def test_single_reaction_deletion_room(self, solver):
```
Midnighter: Can you please split this into two test cases one `test_single_reaction_deletion_room` and another `test_single_reaction_deletion_room_linear`.

cobra/test/test_flux_analysis.py
Midnighter: For the tolerance we have done elsewhere:
```python
assert numpy.allclose(room_sol.fluxes, expected, 1E-06)
```

cobra/test/test_flux_analysis.py
synchon: The `wt_sol` is being used as the reference solution when performing `add_room`.

cobra/flux_analysis/deletion.py
```diff
@@ -123,6 +131,15 @@ def _multi_deletion(model, entity, element_lists, method="fba",
     with model:
         if "moma" in method:
             add_moma(model, solution=solution, linear="linear" in method)
+        if "room" in method:
+            if kwargs:
```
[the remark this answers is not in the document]
synchon: The tolerances need to specified by default and the test case needs different tolerance values. So, I can move this condition in the `add_room` and clean the test interface.

cobra/test/test_flux_analysis.py
Midnighter: I overlooked that, sorry.

PR #709, opened 2018-05-20
refactor: unify objective creation

Midnighter:
This pull request:

* add benchmarks for ROOM
* fix some solver argument usage to test cases
* in most cases makes use of `Objective.set_linear_coefficients`
* brings down the time for `add_moma` and `add_room` for iML1515 from ~5 min to ~6 s

### DIFF
--- a/cobra/flux_analysis/loopless.py
+++ b/cobra/flux_analysis/loopless.py
@@ -82,7 +82,8 @@ def add_loopless(model, zero_cutoff=1e-12):
 
 def _add_cycle_free(model, fluxes):
     """Add constraints for CycleFreeFlux."""
-    model.objective = Zero
+    model.objective = model.solver.interface.Objective(
+        Zero, direction="min", sloppy=True)
     objective_vars = []
     for rxn in model.reactions:
         flux = fluxes[rxn.id]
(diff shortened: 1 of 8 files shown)


cobra/flux_analysis/loopless.py
```diff
@@ -96,8 +97,7 @@ def _add_cycle_free(model, fluxes):
             rxn.bounds = min(flux, rxn.lower_bound), min(0, rxn.upper_bound)
             objective_vars.append(rxn.reverse_variable)
 
-    model.objective.set_linear_coefficients(dict.fromkeys(objective_vars, 1.0))
-    model.objective.direction = "min"
+    model.objective.set_linear_coefficients({v: 1.0 for v in objective_vars})
```
cdiener: Was there a particular reason for that? I always thought that `dict.fromkeys` is actually faster than a dict comprehension...

Midnighter: In my tests the dict comprehension was slightly faster (hardly worth mentioning). I did it more because we were already using comprehensions in other places and I find it a bit more intuitive.

Midnighter:
As shown in the notebook, `optlang.symbolics.add` (which is really `sympy.Add._from_args`) is comparable in speed to `set_linear_coefficients`. Sometimes faster, sometimes slower. I must admit that I didn't check the quadratic expressions as thoroughly.

Midnighter:
I will merge this one now to move on with MOMA and ROOM in other regard but I'm happy to keep tweaking the expressions!

PR #710, opened 2018-05-23
Added feature geometric FBA

### LINKED ISSUE #650: Support geometric FBA?
This is called `geometricFBA` in Matlab's of COBRA, and the method is explained in this paper http://www.sciencedirect.com/science/article/pii/S0022519309000204.

synchon:
This PR adds geometric FBA with tests from original paper and also adds documentation in the notebooks. This also resolves #650. Please review the code @Midnighter.

### DIFF
--- a/cobra/flux_analysis/geometric.py
+++ b/cobra/flux_analysis/geometric.py
@@ -11,7 +11,7 @@ from cobra.flux_analysis.variability import flux_variability_analysis
 from cobra.core import get_solution
 
 
-def geometric_fba(model, epsilon=1E-06, max_tries=30):
+def geometric_fba(model, epsilon=1E-06, max_tries=200):
     """Perform geometric FBA to obtain a unique, centered flux distribution.
 
     Geometric FBA [1]_ formulates the problem as a polyhedron and
@@ -26,7 +26,7 @@ def geometric_fba(model, epsilon=1E-06, max_tries=30):
     epsilon: float, optional
         The convergence tolerance of the model (default 1E-06).
     max_tries: int, optional
-        Maximum number of iterations (default 30).
+        Maximum number of iterations (default 200).
 
     Returns
     -------
@@ -49,7 +49,7 @@ def geometric_fba(model, epsilon=1E-06, max_tries=30):
         # vars and consts storage variables
         consts = []
         obj_vars = []
-        triples = []
+        updating_vars_cons = []
 
         # first iteration
         prob = model.problem
@@ -68,10 +68,10 @@ def geometric_fba(model, epsilon=1E-06, max_tries=30):
                                           name="geometric_fba_upper_const_" +
                                           rxn.id)
             lower_const = prob.Constraint(rxn.flux_expression + var,
-                                          lb=fva_sol["minimum"][rxn.id],
+                                          lb=fva_sol.at[rxn.id, "minimum"],
                                           name="geometric_fba_lower_const_" +
                                           rxn.id)
-            triples.append((var.name, upper_const.name, lower_const.name))
+            updating_vars_cons.append((rxn.id, var, upper_const, lower_const))
             consts.extend([var, upper_const, lower_const])
             obj_vars.append(var)
         model.add_cons_vars(consts)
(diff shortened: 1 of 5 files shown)


cobra/flux_analysis/geometric.py
```diff
@@ -0,0 +1,86 @@
...
+from optlang.symbolics import Zero
+
+from cobra.flux_analysis.parsimonious import add_pfba
+from cobra.flux_analysis.variability import flux_variability_analysis
+from cobra.core import get_solution
+
+
+def geometric_fba(model, epsilon=1E-06):
+    """Perform geometric FBA to obtain a unique
+    centered flux.
```
Midnighter: This summary must be one line exactly. It's fine to not start directly after the quotes, however, if size is a problem. I suggest the following change:
```python
"""
Perform geometric FBA to obtain a unique, centered flux distribution.
```

synchon: Size won't be a problem, I can fit it after the quotes.

cobra/flux_analysis/geometric.py
```diff
@@ -0,0 +1,86 @@
...
+                                          name="geometric_fba_lower_const_" +
+                                          rxn.id)
+            consts.extend([var, upper_const, lower_const])
+            obj_vars.append(var)
+        model.add_cons_vars(consts)
+        # minimize distance between flux and centre
+        model.objective = prob.Objective(Zero, sloppy=True, direction="min")
+        model.objective.set_linear_coefficients({v: 1.0 for v in obj_vars})
+        model.optimize()
+        while delta > epsilon:
```
Midnighter: We need a condition here like `and max_tries < x` otherwise it will be annoying if this never converges (in some freak cases).

synchon: So, what shall be the default number of tries?

cobra/flux_analysis/geometric.py
```diff
@@ -0,0 +1,86 @@
...
+            fva_sol = (flux_variability_analysis(model))
+            v_n = 0.5 * abs(fva_sol['maximum'] + fva_sol['minimum'])
```
Midnighter: Take advantage of the faster numpy/pandas operations :)
```python
v_n = (fva_sol['maximum'] + fva_sol['minimum']).abs() / 2
```

synchon: Right! Totally forgot about it.

cobra/flux_analysis/geometric.py
```diff
@@ -0,0 +1,86 @@
...
+            for rxn in model.reactions:
```
Midnighter: I wonder if it's better to create a list of the triples before rather than performing these string concatenations over-and-over again. Say you have a list of the triples `(var, lower_const, upper_const)` then:
```python
for rxn, (var, lower_const, upper_const) in zip(model.reactions, triples):
    var.ub = v_n[rxn.id]
    # etc.
```
Something more descriptive than `v_n` would also be nice as a variable name.

cobra/flux_analysis/geometric.py
```diff
@@ -0,0 +1,86 @@
...
+                model.variables["geometric_fba_" + rxn.id].ub = v_n[rxn.id]
+                model.constraints["geometric_fba_upper_const_" + rxn.id].ub = \
+                    v_n[rxn.id]
+                model.constraints["geometric_fba_lower_const_" + rxn.id].lb = \
+                    fva_sol['minimum'][rxn.id]
+            model.optimize()
+            delta = max(fva_sol['maximum'] - fva_sol['minimum'])
```
Midnighter: Use `(fva_sol['maximum'] - fva_sol['minimum']).max()`. Also it'd be nice to be consistent with using `'` or `"` as quotes.

cobra/flux_analysis/room.py
```diff
@@ -22,7 +22,7 @@ def add_room(model, solution=None, linear=False, delta=0.03, epsilon=1E-03):
         A previous solution to use as a reference.
     linear : bool
         Whether to use the linear ROOM formulation or not.
-        Default is False.
+        (default False).
```
Midnighter: One line and one sentence if it fits, please.

cobra/flux_analysis/geometric.py
synchon: Can we go for a `or` condition?

synchon: Yes, makes sense.

synchon: Ok cool!

cobra/flux_analysis/geometric.py
```diff
@@ -42,45 +42,65 @@ def geometric_fba(model, epsilon=1E-06):
...
-                                ub=v_n[rxn.id])
+                                ub=mean_flux[rxn.id])
             upper_const = prob.Constraint(rxn.flux_expression - var,
-                                          ub=v_n[rxn.id],
+                                          ub=mean_flux[rxn.id],
                                           name="geometric_fba_upper_const_" +
                                           rxn.id)
             lower_const = prob.Constraint(rxn.flux_expression + var,
-                                          lb=fva_sol['minimum'][rxn.id],
+                                          lb=fva_sol["minimum"][rxn.id],
```
Midnighter: Sorry, I meant to comment on that earlier but forgot. The better way to access a single element of a data frame is:
```python
fva_sol.at[rxn.id, "minimum"]
```

Midnighter: Or you immediately loop over the data frame here since you only use the reaction ID and don't need the reaction object itself:
```python
for (rxn_id, min_flux, max_flux) in fva_sol.itertuples(index=True):
```

synchon: I am using `rxn.flux_expression`, so I think the `Reaction` object is needed. I think it's better to go with `fva_sol.at[rxn.id, "minimum"]`.

Midnighter: Oh yes.

PR #713, opened 2018-05-28
ROOM sanity check fix

synchon:
Aim:
1. This PR addresses the issue of the ROOM sanity check issue with default model. 
2. It also modifies the existing tests just to be close to the aim of the method.

Modifications:
1. The knock-out reaction has been taken as PYK as the paper states it is a better candidate for this purpose. This solves the GLPK solution issue.
2. The assertions have been changed to fit the proper criteria.
3. The ROOM and MOMA methods have been modified so as to make the solution being optional as opposed to the positional (previously).

@Midnighter please review the changes!

### DIFF
--- a/cobra/test/test_flux_analysis.py
+++ b/cobra/test/test_flux_analysis.py
@@ -444,10 +444,11 @@ class TestCobraFluxAnalysis:
         flux_change = (sol.fluxes - knock_sol.fluxes).abs().sum()
         flux_change_room = (sol.fluxes - room_sol.fluxes).abs().sum()
         flux_change_room_ref = (sol.fluxes - room_sol_ref.fluxes).abs().sum()
-        # Expect the ROOM solution to have smaller flux changes in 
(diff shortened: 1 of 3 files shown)


cobra/test/test_flux_analysis.py
```diff
@@ -391,73 +391,71 @@ def test_single_gene_deletion_linear_room_benchmark(self, model,
...
+        assert flux_change_room_ref >= flux_room_ref
 
     @pytest.mark.parametrize("solver", optlang_solvers)
     def test_single_reaction_deletion_room(self, solver):
-        model = construct_papin_2003_model()
+        model = construct_room_model()
         model.solver = solver
-        sol = construct_papin_2003_solution()
+        sol = construct_room_solution()
         expected = Series({'v1': 10.0, 'v2': 5.0, 'v3': 0.0, 'v4': 5.0,
```
Midnighter: Aren't the expected fluxes exactly those in the constructed solution? (Same in line 476.)

PR #714, opened 2018-05-29
Some proposed changes to the solver utilities

cdiener:
This fixes some smaller issues with the solver utilities.

### check_solver_status

The only states currently allowed are OPTIMAL and INFEASIBLE (which throws a warning). However, there are many states that also allow returning fluxes (primal values) and are actually less problematic than INFEASIBLE. For larger models there is FEASIBLE, SUBOPTIMAL or NUMERIC which indicate that the solver found a feasible solution but is not sure whether it is really an optimum (it usually is). I think it should at least be possible to obtain the fluxes from those solver states but currently this is prohibited.

### Gurobi compatibility

optlang has implemented the QP solver for Gurobi since some time now and it seems to work well in my tests (better numerical stability than the cplex one). So I enabled it for the QP problems.

### Bump optlang version

Bumped the optlang version to 1.4.2 to have the QP solver for Gurobi and to avoid the Gurobi bug from 1.4.1 .

*Those changes are not backwards compatible.*

### DIFF
--- a/cobra/util/solver.py
+++ b/cobra/util/solver.py
@@ -407,7 +407,7 @@ def check_solver_status(status, raise_error=False):
     elif (status in has_primals) and not raise_error:
         warn("solver status is '{}'".format(status), UserWarning)
     elif status is None:
-        raise RuntimeError(
+        raise OptimizationError(
             "model was not optimized yet or solver context switched")
     else:
         raise OptimizationError("solver status is '{}'".format(status))
--- a/cobra/util/solver.py
+++ b/cobra/util/solver.py
@@ -20,6 +20,8 @@ from warnings import warn
 
 import optlang
 from optlang.symbolics import Basic, Zero
+from optlang.interface import (OPTIMAL, NUMERIC, FEASIBLE, INFEASIBLE,
+                               SUBOPTIMAL, ITERATION_LIMIT)
 
 from cobra.exceptions import OptimizationError, SolverNotFound,\
     OPTLANG_TO_EXCEPTIONS_DICT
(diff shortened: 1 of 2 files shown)


cobra/util/solver.py
```diff
@@ -397,9 +402,9 @@ def fix_objective_as_constraint(model, fraction=1, bound=None,
 
 def check_solver_status(status, raise_error=False):
     """Perform standard checks on a solver's status."""
-    if status == optlang.interface.OPTIMAL:
+    if status == OPTIMAL:
         return
-    elif status == optlang.interface.INFEASIBLE and not raise_error:
+    elif (status in has_primals) and not raise_error:
         warn("solver status is '{}'".format(status), UserWarning)
     elif status is None:
```
Midnighter: I wonder if we should raise an `OptimizationError` here rather than a `RuntimeError`.

cobra/util/solver.py
```diff
@@ -31,7 +33,10 @@
            for match in dir(optlang) if "_interface" in match}
 
 # Defines all the QP solvers implemented in optlang.
-qp_solvers = ["cplex"]  # QP in gurobi not implemented yet
+qp_solvers = ["cplex", "gurobi"]
+
+# optlang solution statuses which still allow retrieving primal values
+has_primals = [NUMERIC, FEASIBLE, INFEASIBLE, SUBOPTIMAL, ITERATION_LIMIT]
```
Midnighter: I wonder if this is something that should be tackled in optlang. `has_primals` (and `has_duals`?) could be an attribute on the `Model` class.

cdiener: Yeah that would be much cleaner...

Midnighter: I'll open an issue at optlang but will merge this for now in order to keep moving.

PR #719, opened 2018-05-30
Advanced handling of exchanges and growth media

cdiener:
This extends #589. It implements all changes made there and adds additional detection methods for the boundary type of a reaction (exchange, demand, sink).

I am not sure if the boundary type API is super elegant. So would be good if somebody could take a look.

### DIFF
--- a/cobra/medium/boundary_types.py
+++ b/cobra/medium/boundary_types.py
@@ -1,0 +1,126 @@
+"""Contains function to identify the type of boundary reactions.
+
+This module uses various heuristics to decide whether a boundary reaction
+is an exchange, demand or sink reaction. It mostly orientates on the following
+paper:
+
+Thiele, I., & Palsson, B. Ø. (2010, January). A protocol for
+generating a high-quality genome-scale metabolic reconstruction.
+Nature protocols. Nature Publishing Group.
+http://doi.org/10.1038/nprot.2009.203
+"""
+
+from collections import Counter
+import logging
+
+LOGGER = logging.getLogger(__name__)
+
+excludes = {"demand": ["SN_", "SK_", "sink", "EX_", "exchange"],
+            "exchange": ["demand", "DM_", "biosynthesis", "transcription",
+                         "replication", "SN_", "SK_", "sink"],
+            "sink": ["demand", "DM_", "biosynthesis", "transcription",
+                     "replication", "EX_", "exchange"]}
+"""A list of sub-strings in reaction IDs that usually indicate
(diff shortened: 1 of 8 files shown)


cobra/medium/boundary_types.py
```diff
@@ -0,0 +1,128 @@
...
+        if most == "e":
+            return "e"
+        else:
+            LOGGER.warn("There is an `e` compartment but it does not look "
+                        "like it is the actual external compartment.")
+        return most
+    return most
+
+
+def is_type(reaction, btype, ext_compartment):
```
Midnighter: I would like the publicly visible names to be a little more verbose. In this case, for example, I think `is_boundary_type` is more descriptive. Similarly the arguments could be `boundary_type` and `external_compartment`.

cobra/medium/minimal_medium.py
```diff
@@ -0,0 +1,229 @@
...
+    Arguments
+    ---------
+    model : cobra.model
+        The model to modify.
+    """
+    if len(model.variables) > 1e4:
+        LOGGER.warning("the MIP version of minimal media is extremely slow for"
+                       " models that large :(")
+    exchange_rxns = find_boundary_types(model, "exchange")
+    M = max(np.max(np.abs(r.bounds)) for r in exchange_rxns)
```
Midnighter: I think numpy is overkill here. (It will convert each tuple into a new array.) In the following `bounds` is an array of pairs.

```python
In [10]: %timeit max(np.max(np.abs(p)) for p in bounds)
9.2 ms ± 67 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```python
In [13]: %timeit max(abs(x) for p in bounds for x in p)
421 µs ± 2.66 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

cobra/medium/minimal_medium.py
```diff
@@ -0,0 +1,229 @@
...
+            medium[rxn.id] = -flux
+        elif not export:
+            medium[rxn.id] = flux
+    if not exports:
+        medium = medium[medium > 0]
+
+    return medium
+
+
+def minimal_medium(model, min_growth=0.1, exports=False,
```
Midnighter: Maybe `min_growth` should be called `min_objective_value`?

Midnighter:
There are a few typos in the notebook. Should I fix those after merging or do you want to correct them now?

cdiener:
Whatever you prefer. You can also just push to the branch I think.

PR #744, opened 2018-08-09
Gapfill bugfix

maureencarey:
bug fix referenced in Pull Request #687 (Gapfill fix to support new metabolites) using latest devel branch

Updated flux_analysis/gapfilling to support gapfill solutions that add metabolites; tests modified to highlight this case. So, if a gapfill solution adds a reaction that contains metabolites not in the original model, corresponding demand or other reactions will also be added to clear the new metabolites. 

Also updated flux_analysis/gapfilling to check if universal model contains exchange reactions. If exchange_reactions = TRUE, the model is only extended by exchange reactions not already in the universal model.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: python
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
+python: 3.5
 sudo: false
 cache:
   pip: true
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,6 +10,7 @@ branches:
  only:
  - master
  - devel
+ - gapfill_bugfix
```
Midnighter: Please remove this line. Your pull request will be tested on Travis anyway because it is a PR for the devel branch.

PR #752, opened 2018-08-24
refactor: io tests

synchon:
- [ ] YAML
- [ ] JSON
- [ ] MAT
- [ ] SBML


cobra/test/test_io/test_json.py
```diff
@@ -0,0 +1,41 @@
...
+
+"""Test functionalities of json.py"""
+
+from __future__ import absolute_import
+
+from os.path import join
+from warnings import warn
+
+import pytest
+from cobra import io
```
Midnighter: I think, it'd be better to import the `json` stdlib directly here and `import cobra.io as cio` in order to not confuse it with the `io` standard package.

cobra/test/test_io/test_json.py
```diff
@@ -0,0 +1,41 @@
...
+from cobra.test.test_io.conftest import compare_models
+
+try:
```
Midnighter: Take a look at [this section](https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency) and the pytest skipif functionality in general for a better approach to running or not running certain tests. (I know it's done this way in existing tests but we are trying to improve, right? :wink: )

cobra/test/test_io/test_json.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_load_json_model(data_directory, mini_model):
+    """Test the reading of JSON model."""
+    json_model = io.load_json_model(join(data_directory, "mini.json"))
+    assert compare_models(mini_model, json_model) is None
+
+
+def test_save_json_model(tmpdir, mini_model):
+    """Test the writing of JSON model."""
```
Midnighter: It'd also be great to validate the output against the schema.

cobra/test/test_io/test_yaml.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+"""Test functionalities provided by yaml.py"""
+
+from __future__ import absolute_import
+
+from os.path import join
+
+from cobra import io
```
Midnighter: Same as above.

cobra/test/test_io/test_yaml.py
```diff
@@ -0,0 +1,23 @@
...
+from cobra.test.test_io.conftest import compare_models
+
+
+def test_load_yaml_model(data_directory, mini_model):
```
Midnighter: Loading the YAML at a lower level gives you a dictionary structure. It'd be great to test that object structure against the JSONSchema as well.

cobra/test/test_io/test_json.py
synchon: pytest sure has a lot of tricks! :P

cobra/test/test_io/test_json.py
synchon: The schema needs to be updated for both JSON and YAML. For now, I have marked the tests as xfail.

synchon:
The gene comparison is failing for the `.mat` test and it is due to the outdated model annotations stored in it. Also, the order is not being preserved for the genes when they are imported. The rest test cases don't have a problem with the comparison as I have observed.

cobra/test/test_io/test_pickle.py
```diff
@@ -38,7 +38,7 @@ def test_write_pickle(tmpdir, mini_model, dump_function):
         pytest.skip()
 
     output_file = tmpdir.join("mini.pickle")
-    with open(str(output_file), "wb") as outfile:
+    with open(output_file.strpath, "wb") as outfile:
```
Midnighter: I actually find the previous version using `__str__` quite idiomatic. Easy to grasp what's meant.

cobra/test/test_io/test_sbml.py
```diff
@@ -46,9 +46,9 @@ def test_write_sbml_model(tmpdir, mini_fbc1_model, mini_cobra_model):
     mini_cobra_output_file = tmpdir.join("mini_cobra.xml")
 
     # convert to str object before passing the filename
-    io.write_legacy_sbml(mini_fbc1_model, str(mini_fbc1_output_file),
+    io.write_legacy_sbml(mini_fbc1_model, mini_fbc1_output_file.strpath,
```
Midnighter: Same as above, `str(mini_fbc1_output_file)` is nice.

cobra/test/test_io/test_pickle.py
synchon: If you are not merging yet, I'll push another commit addressing this.

Issue #720, opened 2018-05-31
JSON export version attribute doesn't match JSON-Schema


zakandrewking:
#### Problem description

Our team (credit @dbourdeveloper) noticed that the JSON output has a version attribute as a string, but the JSON-Schema specifies that it should be a number:

https://github.com/opencobra/cobrapy/blob/1ca1bf34508e7380e386935e1664ceab21359603/cobra/io/json.py#L13
https://github.com/opencobra/cobrapy/blob/1ca1bf34508e7380e386935e1664ceab21359603/cobra/io/json.py#L148

Since I'm here, a small suggestion: It would be great to pull the JSON-Schema document out into an independent file, so it can be more easily incorporated into non-Python projects.

Thanks!!

Midnighter:
> It would be great to pull the JSON-Schema document out into an independent file, so it can be more easily incorporated into non-Python projects.

Sure thing. I would change the type to string, I think so that we can use a semantic version in future.

Do you have any other thoughts regarding the schema? Is it still serving your purposes?

zakandrewking:
It's been great for us. Semantic versioning is cool, but you'll have to
think about what constitutes a "non-breaking change". Some previous
discussion:

https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/

We have one now for ME-models too. In fact, I think we should JSON-Schema
everything.

zakandrewking:
@Midnighter I can submit a PR for this if the solution is to just switch the type in the JSON Schema to "string".

I can also pull the schema out into a separate file if you want, or just leave it as is.

Midnighter:
My rationale for waiting on this was that we will soon have compartment classes and thus need a version `"2"` as well. If this is very urgent for you then feel free to make a PR, please. Type string and a separate file would be nice. If you can load that file using https://importlib-resources.readthedocs.io/en/latest/ I would much appreciate. Otherwise as announced earlier, we will do a lot of work on cobrapy in August and cover this as well.

zakandrewking:
Ok. I'm not in a rush. Let's leave this open to wait for version "2".

Hemant27031999:
Are there any updates on this? Though the solution of this issue is quite simple and has already been discussed, json_schema still shows the version type to be an integer instead of string.
Thanks.

Midnighter:
The basic problem was addressed by #956.

PR #754, opened 2018-08-24
Feature/implement groups

### LINKED ISSUE #543: Add groups feature
## Feature description

Allow cobrapy objects to be annotated with groups like the legacy `subsystem`. This is supposed to increase compatibility with [SBML groups](http://co.mbine.org/system/files/sbml-groups-version-1-release-1.pdf).

## Possible API

Add a groups attribute to metabolites, reactions and genes that can be used to store arbitrary grouping attributes. Could be a lightweight object or list. Example:

```Python
r = model.reactions[0]
r.groups
# Output = ["TCA cycle", "mitochondrial metabolism"]
```

The SBML interface would read and write the groups.

or to support the `kind` attribute
```
r.groups
# Output = [Group(kind="classification", name="TCA cycle"), Group(kind="classification", name="mitochondrial metabolism"]
```

## Problems

SBML groups allows nested groups like (group1, species1, species2). This is not compatible with the current cobrapy API that assigns groups as properties onto metabolites, reactions or genes. However, building a full fledged group system like `model.groups` would be hard to maintain since one has to sync those with the context manager and model modifications which is not necessary in the proposed API. However, to use it like that we would have to ignore all nested groups since SBML specifically prohibits flattening of the group hierarchy:

> If a Member references a Group, it is the Group itself that is considered to be a member of the parent Group, not 33
the corresponding referenced element (or elements).

gregmedlock:
Implements the `group` class and the `Model.groups` attribute for better compliance with the SBML FBC package, fixes #543. 

Some key implementation choices:
* reactions, metabolites, or genes can belong to groups
* groups can only take "kind" values as specified in the FBC specification
* each `Group` has a `members` attribute, which is a set of reactions, metabolites, and/or genes.
* whenever reactions, metabolites, or genes are removed from a model, group membership is updated.
* a `Group` that is not associated with a model can be generated in isolation and associated with cobrapy objects; upon addition of the `Group` to the model, all cobrapy objects associated with the `Group` are also added.
* when a `Group` is removed from a model, no modifications are made to cobrapy objects that were previously members of the group.
* `model.get_associated_groups()` is used to dynamically look up group membership. This PR does not have a model-based helper function for retrieving group information en masse (e.g. group membership info for all reactions, metabolites and genes with a single function)

Models needed to be repickled for tests to pass; one interesting artifact is a "version" tag in mini.json that is now being regenerated with a string in the corresponding field as "1" instead of the int (1) that is compliant with the jsonschema. I'm not sure what the source of this is, since it seems like it's been a while since the models were repickled in a PR. So, here I just manually corrected the issue in mini.json (changing "1" to 1).

I think I will add documentation for groups in a separate PR; would it make the most sense to add a section in "building a model" and "reading and writing a model"?

Code review and comments for improvement are much appreciated; I tried to keep things consistent with the rest of the codebase where possible. @matthiaskoenig please let me know if I missed some major functionality that would make the libsbml integration smoother!


Midnighter:
I haven't had the time yet to review your code but your description of the PR sounds really promising!

The problem with the JSON version that you see is related to #720 and will be fixed in #725 where we switch to only strings. So no manual editing to stick with the current schema is necessary.

gregmedlock:
Ok, I reran update_pickles.py so the files represent their true state. Also did it to retrigger the travis build; looks like the python 2.7 build timed out without any error.

Midnighter:
Hi @gregmedlock, thanks again for your work here.Your code has been merged as part of #685. I have some code locally that will extend the JSON schema with groups. So that still needs to be done but this code is finally in devel. I'm closing the PR because your code was merged locally.

PR #601, opened 2017-09-27
update modify.py for rename_genes

maureencarey:
updated to permit modifications to a subset of gene ids and to avoid adding genes without GPRs

see below for example motivation:

```python
import cobra
from cobra import Model, Reaction, Metabolite
import os
print cobra.__version__
```

    0.8.2

```python
model = Model()
reaction = Reaction('reaction1')
reaction.name = 'first reaction'
reaction.lower_bound = 0.  # This is the default
reaction.upper_bound = 1000.  # This is the default
met1_c = Metabolite('met1_c', formula='C', name='metabolite 1', compartment='c')
met2_c = Metabolite('met2_c', formula='C', name='metabolite 2', compartment='c')
met3_c = Metabolite('met3_c', formula='C', name='metabolite 3', compartment='c')
met4_c = Metabolite('met4_c', formula='C', name='metabolite 4', compartment='c')
reaction.add_metabolites({met1_c: -1.0, met2_c: 1.0, met3_c: 0, met4_c: 0 })
reaction.gene_reaction_rule = '( old_gene1a and old_gene1b )'
model.add_reactions([reaction])

reaction = Reaction('reaction2')
reaction.name = 'second reaction'
reaction.lower_bound = 0.  # This is the default
reaction.upper_bound = 1000.  # This is the default
reaction.add_metabolites({met1_c: 0, met2_c: -1.0, met3_c: +1.0, met4_c: 0 })
reaction.gene_reaction_rule = '( old_gene2 )'
model.add_reactions([reaction])

reaction = Reaction('reaction3')
reaction.name = 'third reaction'
reaction.lower_bound = 0.  # This is the default
reaction.upper_bound = 1000.  # This is the default
reaction.add_metabolites({met1_c: 0, met2_c: 0, met3_c: -1.0, met4_c: +1.0 })
reaction.gene_reaction_rule = '( old_gene3 )'
model.add_reactions([reaction])

reaction = Reaction('ex1')
reaction.name = 'exchange reaction'
reaction.lower_bound = 0.  # This is the default
reaction.upper_bound = 1000.  # This is the default
reaction.add_metabolites({met1_c: +1.0, met2_c: 0, met3_c: 0, met4_c: 0 })
reaction.gene_reaction_rule = '( old_gene4 or old_gene5 )'
model.add_reactions([reaction])

reaction = Reaction('ex2')
reaction.name = 'exchange reaction'
reaction.lower_bound = 0.  # This is the default
reaction.upper_bound = 1000.  # This is the default
[...]

PR #756, opened 2018-08-30
Bug/rename genes

maureencarey:
Addressing the issues originally brought up in closed PR #601. modify/rename_genes and associate test have been updated to permit the renaming of only a subset of genes and to avoid adding genes without GPRs


Midnighter:
Thank you for picking this up again :smiley: I just pushed one minor change. I'll merge after the CI passes.

Issue #741, opened 2018-07-29
Create a global configuration object (singleton)


Midnighter:
This class should allow setting of default bounds and solver etc.

PR #758, opened 2018-09-11
feature: Configuration class

synchon:
This PR addresses https://github.com/opencobra/cobrapy/issues/741.

- is singleton
- solves setting of default parameters

- [x] make it intelligent
- [x] add tests
- [ ] global-model basis **and/or** per-model basis

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -10,7 +10,7 @@ from os.path import dirname as _dirname
 
 from cobra import flux_analysis, io
 from cobra.core import (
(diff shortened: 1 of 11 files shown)


Midnighter:
Might think about the lower bound on the configuration object being the negative bound of a reversible reaction. Zero as the default lower bound for irreversible reactions seems pretty much the only way to go.

synchon:
Probably now we are left with setting the lower bound of a reversible reaction and deciding if we should extend the configuration to per-model basis too.

Issue #763, opened 2018-09-20
Add functions for feasbility objectives and lexicographic constraints


pstjohn:
Related to #452, i think the functions @synchon started for the dFBA implementation would be useful in cobrapy. Specifically, what would people think about adding some version of

```python
def add_lp_feasibility(model):
    
    obj_vars = []
    prob = model.problem
    for met in model.metabolites:
        s_plus = prob.Variable("s_plus_" + met.id, lb=0)
        s_minus = prob.Variable("s_minus_" + met.id, lb=0)

        model.add_cons_vars([s_plus, s_minus])
        model.constraints[met.id].set_linear_coefficients({s_plus: 1.0, s_minus: -1.0})
        obj_vars.append(s_plus)
        obj_vars.append(s_minus)

    model.objective = prob.Objective(Zero, sloppy=True, direction="min")
    model.objective.set_linear_coefficients({v: 1.0 for v in obj_vars})


def add_lexicographic_constraints(model, objective_list, objective_direction):
    
    for rxn_id, obj_dir in zip(objective_list, objective_direction):
        model.objective = model.reactions.get_by_id(rxn_id)
        model.objective_direction = obj_dir
        cobra.util.fix_objective_as_constraint(model)
```

to cobra.util.solver? I might also edit cobra.util.fix_objective_as_constraint to return the objective / constraint value to avoid having to re-run a solution in a dFBA implementation.

On dFBA specifically, I'm not sure what the plans are for [synchon/dynamic-fba](https://github.com/synchon/dynamic-fba), but I think with the above functions integrated into cobrapy, dFBA would be relatively easy to implement in a single jupyter notebook in the cobrapy docs

davidtourigny:
Hi Peter. I have just started looking into working with cobrapy (and following this thread) as part of new research project with a view to do exactly that (i.e. implement a version of dFBA). However, I am relatively inexperienced with python and much more familiar with C++, which I use write integration platforms for ODE models incorporating various plugins. My first naive attempt at dFBA with cobrapy in python suffers from the same problems you mentioned previously (speed, infeasibility of the LP), and so I started to look into a possible DFBAlab-hybrid, but lack the cobrapy knowledge to progress very quickly. I would therefore be very keen for something like you propose and to remain included in the discussion/process of development- if not to contribute very much, more so just to learn! Thanks, David

synchon:
[...]

Issue #452, opened 2017-03-10
dynamic FBA implementation


phantomas1234:
A much requested feature here at DTU Biosustain. Any particular preference if this should be part of cobrapy or a separate package? I am leaning more towards the first.

pstjohn:
i've coded up a few dynamic fba implementations, a really basic one is [here](https://github.com/pstjohn/dFBA). There are better algorithms out there, [this](https://www.ncbi.nlm.nih.gov/pubmed/23055276) for instance. If I remember correctly all their stuff was in either messy fortran or matlab 😄 .

In terms of speed, you'll really want to make sure the ODE is running in a compiled language. That will be the most difficult part of getting something implemented, since we'll want people to be able to specify uptake rate rules in python. You could probably use things like theano, tensorflow, or [casadi](casadi.org) to construct the symbolic graph in python and then do a jit compilation. 

There's also the option of doing a collocation-based approach, [like this](https://www.ncbi.nlm.nih.gov/pubmed/17022660) that works really well in casadi, but you start to run into limits on how large of a model you can simulate. Also, the IPOPT extension he mentions seems to no longer be supported.

cdiener:
Yes would be interesting. @pstjohn what you're saying sounds interesting but like you said would be a of of C code since you have to implement the LP solving there as well. So is the function call overhead for Python larger than solving the LP in lower-level language?

pstjohn:
If you're following the approach from bruce tidor's group, you only end up resolving the LP when you change solution bases.  If you're using the collocation-based approach, you actually don't end up solving the LP at all, you embed the optimality by using the KKT conditions in the nonlinear program.

If you're doing it naively, the way I did, then you need to resolve the LP at each interval. The bounds are only going to change very slightly, so my expectation is the solve would be very fast - in terms of overhead, just think about all the conversations we've had about whether or not a solution object should pull the flux values from the solver 😄 . That says nothing about whether its mathematically advisable to be solving the LP at each call to f(x,p), I'm sure the writers of the ODE integration code would scoff at all the assumptions you're breaking on continuity, etc.

Solving an ODE will involve hundreds to thousands of calls to dx/dt = f(x, p), so the overhead is nontrivial. Also worth considering (if we're trying to use the method from tidor's group) is that we'll need rootfinding in the ODE integration. I don't think scipy's solvers have that, nor do the SUNDIALS wrappers from casadi (casadi/casadi#974). SUNDIALs proper does though, but that's in C.

[...]

PR #764, opened 2018-09-20
adding dfba utilities and example

pstjohn:
Relevant to #763 and #452, this implements some utility functions to the model solution to make a simple dFBA implementation feasible. 

I also added an [example](https://github.com/pstjohn/cobrapy/blob/solver_util_edits/documentation_builder/dfba.ipynb) of the resulting DFBA implementation to the documentation.

Lots of credit to @synchon for the code + ideas, and my apologies if this is stepping on SoC project. I ended up needing something simple for a dfba project here and figured others might make use as well.

Definitely let me know if you don't think this belongs in cobrapy proper, but I figured it doesn't actually end up adding all that much code


pstjohn:
rebased this on the latest devel. The util.__init__ script should already be pulling everything from the solver utils, so these functions should be importable from cobra.util.

pstjohn:
hold off on merging this for today -- I'm trying out a couple other options.. I'll let you know when this is RTM.

pstjohn:
@Midnighter + @synchon, thoughts on this?

synchon:
I think the inclusion of utility functions will help implementation of various methods and especially DFBA. :)

For the example in the docs, I feel we should state the edge cases and also warn the users that it is just a simple prototype and might lead to convergence issues in some cases. I think it would be best to showcase this as an example of the utilities available in cobrapy, which allows for customized implementations. @Midnighter does that sound good?

@pstjohn won't the use of slack variables cause the constraints to be soft (as compared to metabolites which make them hard constraints)? Is the use of soft constraints desired?

pstjohn:
Hmm, I suppose what I've included is what they do in Eq. 3 of [DFBAlab](https://doi.org/10.1186/s12859-014-0409-8). The lp feasibility problem largely seems to work. 

The edge cases I get seems to be in the lexicographic optimization, the tolerances don't always work out and sometimes I'll get an 'infeasible solution' if for instance, a knocked out product is maximized and then used as a constraint for a subsequent product. 

I'd be happy to put in more qualifying language in the documentation saying this isn't a 'production' DFBA implementation. I've used it for some results here though (for a small-scale model), and might be useful to have to compare against a future implementation that's a bit more developed.

Midnighter:
Thanks for adding this. I would just do some nitpicking with docstrings and the like. Can you give me permission to push a commit to your branch? Otherwise I have to push to devel and it's not transparent.

pstjohn:
are you not able to push to pstjohn/solver_util_edits? I have 'Allow edits from maintainers' checked, I'm not sure how else to give you access

Midnighter:
Sorry, I didn't try before asking :smiley:

PR #768, opened 2018-09-26
refactor: refactor core tests

synchon:
* Also refactors `util` tests as they were in the same file previously.
* Remnants of repeated code stay but will be removed once the upcoming features are added.


Midnighter:
Just something happened with the indentations there. Please fix that and I'll merge this.

PR #774, opened 2018-10-16
Better identification of the extracellular compartment

### LINKED ISSUE #747: Checking model.exchanges membership
I am currently trying to check the membership of model.exchanges to make sure all of the metabolites I have added exchanges for are taken care of properly. When I check for previously existing exchanges as a test I get the correct result of True, but when I check for those I added via add_boundary I always get False returned. This is using cobrapy version 0.13.0

I am adding reactions like this:
```python
cpd00141_e = cobra.Metabolite(
    'cpd00141_e',
    formula='C3H5O2',
    name='Propionate',
    compartment='Extracellular')

model.add_boundary(cpd00141_e, type='exchange', reaction_id='EX_cpd00141_e', lb=-1000.0, ub=1000.0)
```

And checking membership this way, which always returns False:
```python
model.reactions.get_by_id('EX_cpd00141_e') in model.exchanges
```

But the reaction is definitely in place when 
```python
model.reactions.get_by_id('EX_cpd00141_e')
```

This metabolite (cpd00141_e) is already included in a transport reaction I added prior to this check.

cdiener:
This improves handling of the extracellular compartment. In particular it fixes #747 by now raising an error when trying to add an exchange reaction for a Metabolite not in the extracellular compartment.

```Python3

In [2]: from cobra import Metabolite

In [4]: m = Metabolite("new_met", compartment="something")

In [5]: mod.add_boundary(m, type="exchange")
ValueError                                Traceback (most recent call last)
<ipython-input-5-4de374dce0c6> in <module>
----> 1 mod.add_boundary(m, type="exchange")

~/code/cobrapy/cobra/core/model.py in add_boundary(self, metabolite, type, reaction_id, lb, ub)
    511                                  " its compartment or rename the model"
    512                                  " compartments to fix this." %
--> 513                                  (metabolite.compartment, external))
    514         if type in types:
    515             prefix, lb, ub = types[type]

ValueError: The metabolite is not an external metabolite (is `something` but should be `e`). Either change its compartment or rename the model compartments to fix this.
```

I did not implement a new attribute to specify the extracellular compartment but rather ask for the user to name the compartment correctly. There is now a larger set of recognized names consistent with `memote`.

## Additional changes

- annotations for compartments by names and reactions are now refactored in its own module and the idea is that other packages can just import it and use it which to have a common source of annotations across cobra packages
- naming now always dominates the number of boundary reactions which should fix problems like @pstjohn was describing


pstjohn:
I'd probably prefer a warning rather than an exception for adding an exchange for an internal species. 
I'll sometimes add exchanges for ATP / NADH in debugging a model, and wouldn't want to add transport / extra species to make sure they're formally defined.

cdiener:
@pstjohn You can still do that by adding those as demand and sinks. An exchange reaction must come from the outside whereas the other two don't.

```python3

In [2]: from cobra import Metabolite

In [4]: m = Metabolite("new_met", compartment="c")

In [5]: mod.add_boundary(m, type="sink")
Out[5]: <Reaction SK_new_met at 0x1328044400>
```

So this will only throw an error if you specify it as an exchange (`type="exchange"`). The dosctring of `Model.add_boundary` specifies how it interprets `type`.

pstjohn:
ah ok, cool. 👍

cobra/core/model.py
```diff
@@ -500,8 +500,18 @@ def add_boundary(self, metabolite, type="exchange", reaction_id=None,
         >>> demand.build_reaction_string()
         'atp_c --> '
         """
-        types = dict(exchange=("EX", -ub, ub), demand=("DM", 0, ub),
-                     sink=("SK", -ub, ub))
+        types = {"exchange": ("EX", -ub, ub),
```
Midnighter:
```suggestion
        types = {
            "exchange": ("EX", -ub, ub),
            "demand": ("DM", 0, ub),
            "sink": ("SK", -ub, ub)
        }
```

Midnighter: Also, this function should make use of the bounds of the global configuration singleton rather than using these default arguments.

cobra/medium/annotations.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
Midnighter: Should we load these from little data files (JSON, YAML, CSV)?

cobra/medium/boundary_types.py
```diff
@@ -49,25 +35,47 @@ def find_external_compartment(model):
...
+        return most[0]
+        LOGGER.warning("Could not identify an external compartment by name and"
+                       " choosing one with the most boundary reactions. That "
+                       "might be complete nonsense or change suddenly. "
+                       "Consider renaming your compartments using "
+                       "`Model.compartments` to fix this.")
+    # No info in the model, so give up
+    raise RuntimeError("The heuristic for discovering an external compartment "
+                       "relies on names and boundary reactions. Yet, there "
+                       "are no compartments with recognized names and no "
```
Midnighter:
```suggestion
                       "are neither compartments with recognized names nor "
```

cobra/test/test_medium.py
```diff
@@ -143,6 +166,14 @@ class TestErrorsAndExceptions:
     def test_no_boundary_reactions(self, empty_model):
         assert medium.find_boundary_types(empty_model, 'e', None) == []
 
-    def test_no_boundary_reactions(self, empty_model):
+    def test_no_names_or_boundary_reactions(self, empty_model):
         with pytest.raises(RuntimeError):
             medium.find_external_compartment(empty_model)
+
+    def test_bad_exchange(self, model):
```
Midnighter: I'd suggest to use parametrization:

```python
    @pytest.parametrize("metabolite", [
        pytest.param(Metabolite("baddy", compartment="nonsense"), mark=pytest.mark.raises(ValueError)),
        Metabolite("goody", compartment="e")
    ])
    def test_bad_exchange(self, model, metabolite): 
        rxn = model.add_boundary(metabolite)
        assert isinstance(rxn, Reaction)
```

cdiener:
@Midnighter Cool thanks for the comments will integrate them. Are there dos for the new config singleton? I don't know how to use that yet...

Midnighter:
If you build the documentation locally then there will be a little section on the configuration. Unfortunately, our builds on readthedocs are broken right now :disappointed: It should be something like:

```python
from cobra.core import Configuration

CONFIG = Configuration()


def add_boundary(self, metabolite, type="exchange", reaction_id=None,
                     lb=None, ub=None):
    ub = CONFIG.upper_bound if ub is None else ub
    lb = CONFIG.lower_bound if lb is None else lb
```

and then adjust the `types` dictionary to use `lb`, `ub`, I guess? It would be some behavioral change but I don't think it'd practically affect any users.

cobra/test/test_medium.py
cdiener: That does not work:

```
cobra/test/test_medium.py:164: in <module>
    class TestErrorsAndExceptions:
cobra/test/test_medium.py:175: in TestErrorsAndExceptions
    mark=pytest.mark.raises(ValueError)),
../../anaconda3/lib/python3.6/site-packages/_pytest/mark/structures.py:241: in __call__
    store_mark(func, self.mark)
../../anaconda3/lib/python3.6/site-packages/_pytest/mark/structures.py:276: in store_mark
    obj.pytestmark = get_unpacked_marks(obj) + [mark]
E   TypeError: can't set attributes of built-in/extension type 'ValueError'
```

Using `pytest.mark.xfail(raises=ValueError)` does not work either.

cobra/medium/annotations.py
Midnighter: Let's leave as is. Too much work for little gain.

cobra/test/test_medium.py
Midnighter: My bad, need to make sure that `pytest-raises` is installed. Then it should be:
```python
pytest.param(Metabolite("baddy", compartment="nonsense"), mark=pytest.mark.raises(exception=ValueError))
```
with keyword argument.

cobra/core/model.py
```diff
@@ -500,8 +500,22 @@ def add_boundary(self, metabolite, type="exchange", reaction_id=None,
         >>> demand.build_reaction_string()
         'atp_c --> '
         """
-        types = dict(exchange=("EX", -ub, ub), demand=("DM", 0, ub),
-                     sink=("SK", -ub, ub))
+        ub = CONFIGURATION.upper_bound if ub is None else ub
+        lb = CONFIGURATION.lower_bound if lb is None else lb
+        types = {
+            "exchange": ("EX", -ub, ub),
```
Midnighter: It would be a behavioral change but this should probably become for consistency with the configuration.

```suggestion
            "exchange": ("EX", lb, ub),
```

cobra/core/model.py
```diff
@@ -500,8 +500,22 @@ def add_boundary(self, metabolite, type="exchange", reaction_id=None,
...
+            "demand": ("DM", 0, ub),
+            "sink": ("SK", -ub, ub)
```
Midnighter: As above

```suggestion
            "sink": ("SK", lb, ub)
```

cobra/test/test_medium.py
cdiener: Okay, what about just refactoring that into two tests (one for good metabolites and one for bad) and just using the normal parameterize? This way we don't have to introduce new test dependencies.

cobra/core/model.py
cdiener: Yeah, actually that is missing some flexibility since you can not specify a sole import or export reaction (lb=0 or ub=0). But the behavior you suggest makes sense!

cdiener:
Hey, I think this is good to go and would allow us to get rid of #747. The only open discussion is on the elegance of the test functions and that might be better addressed in a new PR that also adds the new dependencies.

Midnighter:
Yes, good idea. I'll try to manage the conflicts.

PR #788, opened 2018-11-29
refactor: make add_boundary arguments more flexible

### LINKED ISSUE #783: model.add_boundary ignores optional argument reaction_id
#### Problem description

When adding a boundary reaction to the model passing a custom id for the new boundary reactions (see code) the model rises an exception.

After inspecting cobra.core.model I've found:
    503        if type in types:
    504             prefix, lb, ub = types[type]
    505             reaction_id = "{}_{}".format(prefix, metabolite.id)
    506         if reaction_id in self.reactions:
--> 507             raise ValueError('boundary %s already exists' % reaction_id)
    508         name = "{} {}".format(metabolite.name, type)
    509         rxn = Reaction(id=reaction_id, name=name, lower_bound=lb,

The expected behavior would be that if reaction_id is not None, the variable name should not be overwritten.

Hope its helps!

Please explain:
* **what** adding a demand variable and set a custom reaction_id
* **how** using the model function add_boundary
* **why** Because it ignores the reaction_id argument
  you expected instead.

#### Code Sample

```python
# Paste your code here.
model.add_boundary(model.metabolites.trp_L_c, type='demand', reaction_id='MY_DEMAND')
```

#### Actual Output
ValueError: boundary DM_trp_L_c already exists
#### Expected Output
the new reaction variable with the passed id
#### Dependency Information

Please paste the output of `python -c "from cobra import show_versions;
show_versions()"` in the details block below.

<details>

[Paste output here.]

</details>

Midnighter:
Fix #783

### DIFF
--- a/cobra/core/model.py
+++ b/cobra/core/model.py
@@ -19,7 +19,7 @@ from cobra.core.object import Object
 from cobra.core.reaction import Reaction, separate_forward_and_reverse_bounds
 from cobra.core.solution import get_solution
 from cobra.exceptions import SolverNotFound
-from cobra.medium import find_boundary_types
+from cobra.medium import find_boundary_types, sbo_terms
 from cobra.util.context import HistoryManager, get_context, resettable
 from cobra.util.solver import (
     add_cons_vars_to_problem, assert_optimal, interface_to_str,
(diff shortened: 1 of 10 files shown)


cobra/core/model.py
```diff
@@ -500,17 +505,34 @@ def add_boundary(self, metabolite, type="exchange", reaction_id=None,
...
         'atp_c --> '
         """
-        types = dict(exchange=("EX", -ub, ub), demand=("DM", 0, ub),
-                     sink=("SK", -ub, ub))
+        if ub is None:
+            ub = CONFIGURATION.upper_bound
+        if lb is None:
+            lb = CONFIGURATION.lower_bound
+        types = {
+            "exchange": ("EX", lb, ub, "SBO:0000627"),
```
cdiener: I would prefer taking the SBO terms from `cobra.medium.boundary_types.sbo_terms`. Might make it a bit uglier but future proof since there is a single source of truth for the SBO terms (for instance if they change at some point). So fir instance use `sbo_terms["exchange"]` rather than `"SBO:0000627"`.

Midnighter: I anyway want to make arguing about the ontology possible in future so you're absolutely right.

PR #791, opened 2018-11-30
Fix inf bounds

### LINKED ISSUE #784: Handle explicit infinity values for unbounded constraints
#### Problem description

Using values like `math.inf` for bounds results in problems with de/serialization (although simulations seem to run fine).

I suggest that using `math.inf` / `float('inf')`, `numpy.inf` ,`None`,  and `numpy.nan` should all treat the value as unbounded.

#### Code Sample

```python
from cobra.io import load_json_model, save_json_model
import math

model = load_json_model('e_coli_core.json')
model.reactions.ACALD.upper_bound = math.inf
save_json_model(model, 'e_coli_core.json')
```

#### Actual Output

```
ValueError: Out of range float values are not JSON compliant: inf
```

#### Expected Output

Serialized model written to disk with unbounded constraint for the chosen reaction.

#### Dependency Information

<details>

System Information
==================
OS                       Linux
OS-release 4.19.2-arch1-1-ARCH
Python                   3.6.7

Package Versions
================
cobra        0.13.4
depinfo       1.4.0
future       0.17.1
numpy        1.15.3
optlang       1.4.3
pandas       0.23.4
pip            18.0
ruamel.yaml 0.15.75
setuptools   40.6.2
six          1.11.0
swiglpk      4.65.0
tabulate      0.8.2
wheel        0.32.3
</details>

Midnighter:
Fix #784 

While trying to handle infinite values for bounds, I tripped over the behavior that adjusts both bounds when the lower bound is larger than the upper bound or vice versa. Is this something people use a lot? To me that seems like trying to be too smart for the user's own good.

### DIFF
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -184,20 +184,6 @@ class Reaction(Object):
         cop = deepcopy(super(Reaction, self), memo)
         return cop
 
-    @property
-    def lower_bound(self):
-        """Get or set the lower bound
-
-        Setting the lower bound (float) will also adjust the associated optlang
-        variables associated with the reaction. Infeasible combinations,
-        such as a lower bound higher than the current upper bound will
-        update the other bound.
-
-        When using a `HistoryManager` context, this attribute can be set
-        temporarily, reversed when the exiting the context.
-        """
-        return self._lower_bound
-
     @staticmethod
     def _check_bounds(lb, ub):
         if lb > ub:
@@ -231,10 +217,30 @@ class Reaction(Object):
                 ub=None if isinf(self._lower_bound) else -self._lower_bound
             )
 
+    @property
+    def lower_bound(self):
+        """Get or set the lower bound
+
+        Setting the lower bound (float) will also adjust the associated optlang
+        variables associated with the reaction. Infeasible combinations,
+        such as a lower bound higher than the current upper bound will
+        update the other bound.
+
+        When using a `HistoryManager` context, this attribute can be set
+        temporarily, reversed when the exiting the context.
+        """
+        return self._lower_bound
+
     @lower_bound.setter
     @resettable
     def lower_bound(self, value):
         if self._upper_bound < value:
+            warn("You are constraining the reaction '{}' to a fixed flux "
+                 "value of {}. Did you intend to do this? We are planning to "
+                 "remove this behavior in a future release. Please let us "
+                 "know your opinion at "
+                 "https://github.com/opencobra/cobrapy/issues/793."
+                 "".format(self.id, value), DeprecationWarning)
             self.upper_bound = value
         # Validate bounds before setting them.
         self._check_bounds(value, self._upper_bound)
@@ -259,6 +265,12 @@ class Reaction(Object):
     @resettable
     def upper_bound(self, value):
         if self._lower_bound > value:
+            warn("You are constraining the reaction '{}' to a fixed flux "
+                 "value of {}. Did you intend to do this? We are planning to "
+                 "remove this behavior in a future release. Please let us "
+                 "know your opinion at "
+                 "https://github.com/opencobra/cobrapy/issues/793."
+                 "".format(self.id, value), DeprecationWarning)
             self.lower_bound = value
         # Validate bounds before setting them.
         self._check_bounds(self._lower_bound, value)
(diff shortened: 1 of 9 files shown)


gregmedlock:
Agreed, I think that could lead to some unexpected model changes, e.g. if someone is increasing the lower bound on a demand reaction, they could end up fixing flux through the demand reaction at a particular value when the upper bound on the reaction is low to start. Better to just raise an error if the user tries to optimize a model with conflicting lb/ub, IMO.

Midnighter:
> with the approach of raising a ValueError you've implemented (e.g. _check_bounds in the setter for lower_bound and upper_bound), will the error come up as soon as the bound is improperly set on the reaction? If so, this could lead to some unexpected errors based on the order that users change bounds.

Yes, indeed. If you want to modify both bounds the preferred way is to use the `bounds` property of a reaction and modify both bounds at the same time.

> I think that could lead to some unexpected model changes

Please note that this has been the behavior for a long time now. I'm questioning it but I'm also concerned about breaking existing code. Although, maybe it's good to break that code and make the situation more explicit.

gregmedlock:
I see--I've always used `lower_bound` and `upper_bound` directly, so I made a small addition to the docs to demonstrate the preferred behavior in #792

Curious to hear if any users were intentionally using this one; definitely agree this aspect could break some things, but I favor your modification.

phantomas1234:
As a user you would have to start writing a lot of conditionals without this behavior (which I would find annoying). `reaction.bounds` didn't exist when I made this design decision unfortunately. I am not sure how much code it would break, but probably a lot in cameo for example.

Midnighter:
> As a user you would have to start writing a lot of conditionals without this behavior (which I would find annoying).

Can you give an example of this? I don't see what you mean right now.

> I am not sure what the old cobrapy behavior was?

So to be clear what the behavior has been:
```python
from cobra.test import create_test_model


model = create_test_model("textbook")
reaction = model.reactions.ACALD

# Verify current bounds.
assert reaction.lower_bound == -1000
assert reaction.upper_bound == 1000
assert reaction.forward_variable.lb == 0
assert reaction.forward_variable.ub == 1000
assert reaction.reverse_variable.lb == 0
assert reaction.reverse_variable.ub == 1000

# Set upper bound to a value that's lower than the current lower bound.
reaction.upper_bound = -1100
assert reaction.lower_bound == -1100
assert reaction.upper_bound == -1100
assert reaction.forward_variable.lb == 0
assert reaction.forward_variable.ub == 0
assert reaction.reverse_variable.lb == 1100
assert reaction.reverse_variable.ub == 1100

# Set lower bound to a value larger than the current upper bound.
reaction.lower_bound = 100
assert reaction.lower_bound == 100
assert reaction.upper_bound == 100
assert reaction.forward_variable.lb == 100
assert reaction.forward_variable.ub == 100
assert reaction.reverse_variable.lb == 0
assert reaction.reverse_variable.ub == 0
```

Setting the lower or upper bound individually was, of course, also always possible as long as the value observed `lower_bound <= upper_bound`. Additionally, since 2016 it is possible to modify both bounds at the same time which fully covers above use-cases.

```python
reaction.bounds = -1100, -1100
```

or

```python
reaction.bounds = 100, 100
```

So to me this sort of behavior was surprising (I never used it before) and so I'm curious if many people make use of it.

Midnighter:
Alright, had some offline discussion with @phantomas1234 and we will maintain the old behavior. I still find it surprising but there's no good reason to break existing behavior.

cobra/core/reaction.py
```diff
@@ -200,14 +198,48 @@ def lower_bound(self):
...
         return self._lower_bound
 
+    @staticmethod
+    def _check_bounds(lb, ub):
+        if lb > ub:
+            raise ValueError(
+                "The lower bound must be less than or equal to the upper "
+                "bound ({} <= {}).".format(lb, ub))
+
+    def update_variable_bounds(self):
```
pstjohn: Are we pretty sure this will play OK with the context manager? I suppose the tests all pass, so that should mean it works.

Midnighter:
> Perhaps the thing to do is give a warning so we could drop it later. At the very least the user should be warned something is happening they might not intend.

Good idea!

cobra/core/reaction.py
Midnighter: I did not test this explicitly but I only touched one test to change from `AssertionError` to `ValueError`.

PR #792, opened 2018-11-30
updated getting_started w/ preferred approach for modifying reaction bounds

gregmedlock:
Small addition to Reaction section in getting started documentation to demonstrate preferred method for modifying reaction bounds.


Midnighter:
Hi Greg, thank you for the initiative. I would like to resolve the discussion on #791 and then put whatever we conclude in the documentation. Does that sound fair?

gregmedlock:
Sounds great @Midnighter -- when the discussion is resolved I'll update the PR.

Midnighter:
Alright, #791 is now merged.

gregmedlock:
Good to go now. Not sure what's up, but the DeprecationWarning you added in #791 doesn't seem to be printing.

cdiener:
I think normal python warnings will not show in Jupyter notebooks but are redirected to the running terminal. We should probably convert all warnings to logger warnings at some point to give more control for handling as well.

PR #801, opened 2019-02-06
Lower log level for the expected case

kvikshaug:
With the merge of #774, this statement is now seen very frequently
in our applications. I suggest lowering the log level as this code path
seems to be the expected case, and therefore is a bit too verbose to be
informative.

### DIFF
--- a/cobra/medium/boundary_types.py
+++ b/cobra/medium/boundary_types.py
@@ -51,9 +51,9 @@ def find_external_compartment(model):
 
     if matches.sum() == 1:
         compartment = matches.index[matches][0]
-        LOGGER.info("Compartment `%s` sounds like an external compartment. "
-                    "Using this one without counting boundary reactions" %
-                    compartment)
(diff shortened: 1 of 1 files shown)


cdiener:
Dies that happen even if not setting the log level explicitly? I was under the impression the default is to only show warnings and errors... If you explicitly choose INFO you should expect to see a lot of messages since this is the intention of this log level. However, I agree it should not be the default to show this message.

kvikshaug:
We do indeed explicitly show `info` level from all loggers. I guess we just haven't noticed a lot of noise from cobrapy yet. If that's the intention, then it sounds like the level is correct and we should rather raise cobra's log level to `warning` in our app, so I'll close this. Thanks!

cdiener:
Thanks for getting back. That seems like the best behavior. I wouldn't really have a problem with changing it in this instance, but there is no guarantee cobrapy won't show more info messages in the future.

Issue #469, opened 2017-03-27
Implement better errors


hredestig:
Once fixed in optlang, use the better errors in `get_solution` etc to avoid the catch-alls

Midnighter:
The relevant issue on optlang is https://github.com/biosustain/optlang/issues/75.

synchon:
Do we have a rough outline of how we should define the error hierarchy?

synchon:
I was making some cosmetic changes to `cobra.exceptions` and thought maybe this is a good time to revisit this and maybe discuss a bit more. Would like to know others thoughts on this. :)

Midnighter:
Good idea to bring it up again 😃 Need to take a fresh look.

PR #805, opened 2019-02-19
update: Configuration

synchon:
This PR updates the Configuration object by adding ability to set solver tolerances. It also updates the representation of the object.

### DIFF
--- a/cobra/core/configuration.py
+++ b/cobra/core/configuration.py
@@ -99,10 +99,42 @@ class BaseConfiguration(object):
         self.upper_bound = bounds[1]
 
     def __repr__(self):
-        return "solver: {}\nlower_bound: {}\nupper_bound: {}\n" \
-            "processes: {}".format(
-                interface_to_str(self.solver), self.lower_bound,
-                self.upper_bound, self.processes)
+        return """
+        solver: {solver}
+        lower_bound: {lower_bound}
+        upper_bound: {upper_bound}
+        processes: {processes}""".format(
+            solver=interface_to_str(self.solver),
+            lower_bound=self.lower_bound,
+            upper_bound=self.upper_bound,
+            processes=self.processes,
+        )
+
+    def _repr_html_(self):
+        return """
+        <table>
+            <tr>
+                <td><strong>Solver</strong></td>
+                <td>{solver}</td>
+            </tr>
+            <tr>
+                <td><strong>Lower bound</strong></td>
+                <td>{lower_bound}</td>
+            </tr>
+            <tr>
+                <td><strong>Upper bound</strong></td>
+                <td>{upper_bound}</td>
+            </tr>
+            <tr>
+                <td><strong>Processes</strong></td>
+                <td>{processes}</td>
+            </tr>
+        </table>""".format(
+            solver=interface_to_str(self.solver),
+            lower_bound=self.lower_bound,
+            upper_bound=self.upper_bound,
+            processes=self.processes,
+        )
 
 
 class Configuration(with_metaclass(Singleton, BaseConfiguration)):
--- a/cobra/core/configuration.py
+++ b/cobra/core/configuration.py
@@ -9,7 +9,7 @@ import types
 from multiprocessing import cpu_count
 from warnings import warn
 
-from six import string_types, with_metaclass
+from six import string_types, viewvalues, with_metaclass
 
 from cobra.core.singleton import Singleton
 from cobra.exceptions import SolverNotFound
@@ -34,6 +34,8 @@ class BaseConfiguration(object):
     solver : {"glpk", "cplex", "gurobi"}
         The default solver for new models. The solver choices are the ones
         provided by `optlang` and solvers installed in your environment.
+    tolerances: dict('feasibility': float, 'optimality': float, 'integrality': float)
+        The tolerances for the solver being used (default is solver's initial tolerances).
     lower_bound : float
         The standard lower bound for reversible reactions (default -1000).
     upper_bound : float
@@ -50,6 +52,7 @@ class BaseConfiguration(object):
 
     def __init__(self):
         self._solver = None
+        self._tolerances = None
         self.lower_bound = None
         self.upper_bound = None
         # Set the default solver from a preferred order.
@@ -60,7 +63,15 @@ class BaseConfiguration(object):
                 continue
             else:
                 break
+        
+        self.tolerances = {
+            'feasibility': self.solver.Configuration().tolerances.feasibility,
+            'optimality': self.solver.Configuration().tolerances.optimality,
+            'integrality': self.solver.Configuration().tolerances.integrality,
+        }
+
         self.bounds = -1000.0, 1000.0
+
         try:
             self.processes = cpu_count()
         except NotImplementedError:
@@ -98,13 +109,32 @@ class BaseConfiguration(object):
         self.lower_bound = bounds[0]
         self.upper_bound = bounds[1]
 
+    @property
+    def tolerances(self):
+        return self._tolerances
+
+    @tolerances.setter
+    def tolerances(self, tolerance_dict):
+        if all(isinstance(x, float) 
+               for x in list(viewvalues(tolerance_dict))):
+            self._tolerances = {
+                'feasibility': tolerance_dict.get('feasibility', 1e-07),
+                'optimality': tolerance_dict.get('optimality', 1e-07),
+                'integrality': tolerance_dict.get('integrality', 1e-05),
+            }
+        else:
+            LOGGER.warning(
+                "Only float values are allowed for setting tolerances.")
+
     def __repr__(self):
         return """
         solver: {solver}
+        toleraces: {tolerances}
         lower_bound: {lower_bound}
         upper_bound: {upper_bound}
         processes: {processes}""".format(
             solver=interface_to_str(self.solver),
+            tolerances=self.tolerances,
             lower_bound=self.lower_bound,
             upper_bound=self.upper_bound,
             processes=self.processes,
@@ -116,6 +146,10 @@ class BaseConfiguration(object):
             <tr>
                 <td><strong>Solver</strong></td>
                 <td>{solver}</td>
+            </tr>
+            <tr>
+                <td><strong>Tolerances</strong></td>
+                <td>{tolerances}</td>
             </tr>
             <tr>
                 <td><strong>Lower bound</strong></td>
@@ -131,6 +165,7 @@ class BaseConfiguration(object):
             </tr>
         </table>""".format(
             solver=interface_to_str(self.solver),
+            tolerances=self.tolerances,
             lower_bound=self.lower_bound,
             upper_bound=self.upper_bound,
             processes=self.processes,
--- a/cobra/core/configuration.py
+++ b/cobra/core/configuration.py
@@ -12,7 +12,7 @@ from warnings import warn
 from six import string_types, with_metaclass
 
 from cobra.core.singleton import Singleton
-from cobra.exceptions import BoundsError, SolverNotFound
+from cobra.exceptions import SolverNotFound
 from cobra.util.solver import interface_to_str, solvers
 
 
@@ -100,11 +100,9 @@ class BaseConfiguration(object):
     @bounds.setter
     def bounds(self, bounds):
         # TODO: We should consider allowing `None` for free bounds.
-        if bounds[0] <= bounds[1]:
-            self.lower_bound = bounds[0]
-            self.upper_bound = bounds[1]
-        else:
-            raise BoundsError("Lower bound can't be greater than upper bound.")
+        assert bounds[0] <= bounds[1]
+        self.lower_bound = bounds[0]
+        self.upper_bound = bounds[1]
 
     def __repr__(self):
         return """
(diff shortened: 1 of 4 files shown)


cobra/core/configuration.py
```diff
@@ -94,15 +107,73 @@ def bounds(self):
     @bounds.setter
     def bounds(self, bounds):
         # TODO: We should consider allowing `None` for free bounds.
-        assert bounds[0] <= bounds[1]
-        self.lower_bound = bounds[0]
-        self.upper_bound = bounds[1]
+        if bounds[0] <= bounds[1]:
```
Midnighter: Why do you suggest this change in behavior? I think that actually raising an exception is important rather than an error which can be missed and will lead to problems later.

cobra/core/configuration.py
```diff
@@ -60,7 +65,15 @@ def __init__(self):
                 continue
             else:
                 break
+
+        self.tolerances = {
```
Midnighter: I'm not 100% sure about the current behavior in optlang but I know that not all solvers implement all of these tolerances. So I wonder if a simplification to a single tolerance value on the config class would make more sense.

cobra/test/test_core/test_configuration.py
```diff
@@ -32,3 +32,31 @@ def test_solver():
     assert interface_to_str(config.solver) == "glpk_exact"
     # Restore default solver.
     config.solver = "glpk"
+
+
+def test_default_tolerances():
+    """Verify the default solver tolerances."""
```
Midnighter: I wouldn't expect the default tolerance to always be the same. Rather, we should implement a default cobrapy tolerance across solvers.

cobra/core/configuration.py
synchon: Oh yes! I totally forgot to mention that I was thinking that adding a new exception like `BoundsError` might be a better option than `AssertionError`. This will not only give a proper context for the error but also can be introduced to other parts of the code base which might require it.

cobra/core/configuration.py
synchon: Although simplification to a single tolerance value might be an easier to understand design, it might make it a bit difficult if an user wants to access the other tolerance values. A possible solution might be to issue an warning when an user is trying to set a tolerance value which the solver does not possess. I feel we can provide a bit more information about the solvers and their tolerances in the docstring.

cobra/test/test_core/test_configuration.py
synchon: I thought that GLPK would always initialize with the same default tolerance.

cobra/core/configuration.py
cdiener: I agree that they do different things, unfortunately many solvers conflate those. For instance you can not set different tolerances for feasibility and optimality in GLPK. Optlang currently ignores all tolerances except feasibility for GLPK for instance. I think this is something that optlang needs to solve since it is supposed to give a unified interface to different solvers.

cobra/test/test_core/test_configuration.py
cdiener: The same version probably does but users might have different GLPK versions. Having a cobrapy specific value definitely makes sense.

cobra/core/configuration.py
Midnighter: Well, optlang offers these three settings but some of them don't have an effect for certain solvers. Maybe not the best design but somewhat unified. I wonder if we should go a step further in cobrapy and simply have one tolerance value that is then set on these three optlang attributes.

synchon: In my opinion, setting a default tolerance value for solvers in cobrapy seems a good idea.

Midnighter: So what I'm suggesting is something like:

```python
def __init__(...):
    ...
    self._tolerance = None
    self.tolerance = 1E-09

@property
def tolerance(self):
    return self._tolerance

@tolerance.setter
def tolerance(self, value):
    config = self.solver.Configuration()
    config.tolerances.feasibility = value
    config.tolerances.optimality = value
    config.tolerances.integrality = value
    self._tolerance = value
```

Midnighter: Or maybe those properties should actually be on the model class and the configuration holds a simple global default...

cobra/test/test_core/test_configuration.py
synchon: Does `1E-06` as the default cobrapy solver tolerance sound good?

cobra/core/configuration.py
synchon: I can imagine two situations:
- the user is in a notebook-like environment: Using `Configuration` is the recommended way.
- the user is running a program: Is directly setting the solver a better choice? How would using Configuration in that situation be?

I think we should have a separate section on configuring cobrapy internals.

cobra/test/test_core/test_configuration.py
Midnighter: I know of a few models that are feasible with a tolerance of 1E-06 but infeasible with something stricter or the exact solver. So I'm tending towards a stricter default tolerance. I'm not sure how to choose the right one, though.

cobra/core/configuration.py
Midnighter: I'm thinking that one might want to set the tolerance on a per model basis. Thus there should be a global default in the configuration with which models are initialized but then you should be able to change the parameter for each model. That would mean that `cobra.Configuration.tolerance` is a simple attribute and the properties I outlined above would go onto `cobra.Model.tolerance`.

synchon: So then `cobra.Model.tolerance` would inherit the value from `cobra.Configuration.tolerance` and `cobra.Model.tolerance` would set the tolerance value for the associated solver. And also the `cobra.Model.tolerance` would allow you to change the value which would in turn change the tolerance for the associated solver. Am I getting it right?

cobra/test/test_core/test_configuration.py
synchon: Okay then `1E-09` seems good for the default config.

cobra/core/configuration.py
Midnighter: Yep, exactly :smiley:

cobra/test/test_core/test_configuration.py
cdiener: > I know of a few models that are feasible with a tolerance of 1E-06 but infeasible with something stricter or the exact solver. So I'm tending towards a stricter default tolerance. I'm not sure how to choose the right one, though.

This might lead to serious performance regressions though. I would recommend to check with an FVA what impact that might have.

cobra/core/model.py
```diff
@@ -149,6 +154,18 @@ def solver(self, value):
             return
         self._solver = interface.Model.clone(self._solver)
 
+    @property
+    def tolerance(self):
+        return self._tolerance
+
+    @tolerance.setter
+    def tolerance(self, value):
+        solver_config = self._solver.configuration
```
Midnighter: Would this work?
```suggestion
        tolerances = self._solver.configuration.tolerances
```

cobra/core/model.py
```diff
@@ -149,6 +154,18 @@ def solver(self, value):
...
+        solver_config.tolerances.feasibility = value
```
Midnighter: Some of these will raise `AttributeError` if they are not implemented on the specific solver. This could be more nicely handled on optlang but that's where it is at the moment. So each line individually should probably be wrapped in a `try` ... `except`.

cobra/exceptions.py
```diff
@@ -32,6 +32,12 @@ class SolverNotFound(Exception):
     pass
 
 
+class BoundsError(Exception):
```
Midnighter: I'm not a huge fan of introducing too specific error classes. Maybe we need a base `COBRAError` class and a more descriptive message per error. What do others think? @opencobra/cobrapy-core

cobra/core/configuration.py
```diff
@@ -50,8 +52,10 @@ class BaseConfiguration(object):
 
     def __init__(self):
         self._solver = None
+        self.tolerance = 1E-09
```
Midnighter: As @cdiener suggested, I will check feasibility and blocked reactions for all models that I have available using 1E-09 and 1E-06 as tolerance values.

cobra/core/model.py
synchon: Would you want that to be a dictionary?

cobra/core/model.py
synchon: Ah it completely slipped my mind. Will fix that.

cobra/exceptions.py
synchon: I dont't see any problem with having a base class for cobrapy related errors. Then, we could roughly divide the errors into errors from solver optimization and other basic config errors.

synchon:
> One thing that's missing is that all functions which use a tolerance in order to threshold flux results should now by default us the tolerance of the model, I think. Or is that a change that will break too much existing code?

I'll dig a bit more into this and let you know my observations. But, in general, I think it shouldn't be much difficult.

cobra/exceptions.py
cdiener: Aren't there solver specific errors already? There is an OptimizationError I think. If any we might want to have a FeasibilityError which includes bounds as well as equalities, but I think a generic SolverError or SolutionError would also work.

cdiener:
That behavior should also be changed in sampling since it also currently uses its own I think.

cobra/exceptions.py
kvikshaug: +1 for specific error classes that also inherit from a cobrapy base error. It could also make sense that this specific error inherits from `ValueError` as well.
I would add that the exception thrown here should be consistent with the [similar check when setting reaction bounds](https://github.com/opencobra/cobrapy/blob/4481472c90ae653f92aacc6bdb4210dbb715e849/cobra/core/reaction.py#L190).

matthiaskoenig:
Can we change the default bounds from
```
self.bounds = -1000.0, 1000.0
```
to 
```
self.bounds = -float("Inf"), float("Inf")
```
in the configuration. Otherwise arbitrary constraints are created on all reactions without the user noticing. Cobrapy works perfectly with [-Inf, Inf] bounds. If users want to have arbitrary bounds they can set the upper and lower bound to whatever they want, but the correct behavior for the LP problem is to not constraint reactions by default. The only constraints added by a reaction to the LP problem are the stoichiometric constraint. Adding some additional arbitrary constraints is in my opinion incorrect behavior.
All tests pass with `[-float("Inf"), float("Inf")]`, so no reason not to change it to the correct behavior.

Midnighter:
> Cobrapy works perfectly with [-Inf, Inf] bounds.

Are you sure about this? I seem to remember some solvers (CPLEX) supporting infinity and others not (GLPK, Gurobi).

The optlang way would be to set `None` instead of a value. We would just have to adjust bounds checks in order to reflect potential `None` values otherwise we get `TypeError`s all over the place.

matthiaskoenig:
As I understand it the [-Inf, Inf] should not add any constraint on the fluxes, so basically all solvers should support it, i.e. the following should happen for the LP problem
* for `[lb <= v1 <= ub]` the constraint `lb <= v1 <= ub` should be added
* for `[lb <= v1 <= INF]` the constraint `lb <= v1` should be added
* for `[-INF <= v1 <= ub]` the constraint `v1 <= ub` should be added
* for `[-INF <= v1 <= INF]` no constraint is added
Nowhere is there support for INF, -INF in the solvers required.

Midnighter:
We just need to build test cases. I'm quite sure that the symbol `float("inf")` is simply not recognized. I know for sure that using `None` works, so the behavior that you describe can definitely be achieved with that, i.e., `optlang.Variable('v1', lb=0, ub=None)` means 0 <= v1.

synchon:
@Midnighter using `Model.tolerance` should be fairly easy for functions using tolerance as flux threshold. Will opening another PR for this make sense?
@cdiener while going through the `sampling.py`, I was thinking if having a separate submodule for `sampling` make sense. Currently, I feel the code there is a bit of a pain to maintain and splitting the code would definitely reduce it. How do you feel about it?

Midnighter:
> @Midnighter using Model.tolerance should be fairly easy for functions using tolerance as flux threshold. Will opening another PR for this make sense?

Yes, a separate PR is a better idea. Using `Model.tolerance` is easy but doing it in a way that maintains the old function interfaces as much as possible will need a careful look.

cdiener:
I'm all for refactoring the sampling code into a module.

Midnighter:
Okay, let's divide and conquer this issue. It seems to me that there is a consensus for adding a tolerance attribute on the configuration and new behavior for the model class. My suggestion is to merge that code now in this PR and tackle the remaining issues separately. That means:

* Please revert the introduction of the `BoundsError` and the change in logic of the `bounds` property.
* We can hijack https://github.com/opencobra/cobrapy/issues/469 in order to discuss a new exception hierarchy.
* I have opened an issue #815 for discussing the implementation of unbounded (infinite) variables.
* I have opened #816 in order to discuss a proposed change to the default bounds.

While I think that having a new default tolerance, an exact value still has to be determined. This requires action by me to investigate the large model collections that we have assembled.

cobra/core/model.py
Midnighter: No, it was just an idea to shorten the remaining lines.

synchon: It should work. I will push the change. But, I would prefer `solver_tolerances` instead of `tolerances` so that it's a bit more explicit.

Midnighter: Perfect.

Midnighter:
@synchon I will set the value to the less strict 1E-07 for now. I want to merge the code but can't make time to fully investigate right now. We can always set it to a stricter default in a following PR.

PR #809, opened 2019-02-22
Fix appveyor

cdiener:
This is a bit of a hack which should make appveyor work again. 

I looked around a bit in other larger packages (pandas, matplotlib, etc.) and all of them have dropped Python 2 from the tests...


cdiener:
Yes and we should also include Python 3.7. Should I merge this anyways?

Midnighter:
Please do. No point in waiting.

Issue #817, opened 2019-03-10
Refactor sampling.py and separate the base classes from samplers 


synchon:
Refactoring `sampling.py` will simplify future maintenance. I think creating a new directory named `flux_sampling` inside the `flux_analysis`, can be used to keep all the sampling related code.

cdiener:
I think just `sampling` is fine for the name. Also could imagine it to be a folder next to `flux_analysis` to keep hierarchy flat (might make sense since all analysis methods are optimization-based but sampling is not).

PR #820, opened 2019-03-16
refactor: sampling

synchon:
This PR address #817 .
- Uses `Configuration.tolerance` to set `bounds_tol` and `feasibility_tol`. Should it use `model.tolerance` instead?
- Should it replace the `processes` in `OptGPSampler` with `Configuration.processes`?


synchon:
The tests will pass once #805 is merged.

synchon:
Reading the tolerance from the model will definitely be a better decision and I had that in my mind. That will also be a significant change to the way sampling was handled till now.

Another point is that there are two tolerances: feasibility tolerance and bounds tolerance, and both used the same values till now. Should that be the way as it is and be set to the model tolerance?

The ability to change the tolerance will also allow us to add a few more tests. Is that something we want to do?

cobra/test/test_sampling/test_sampling.py
```diff
@@ -9,145 +9,75 @@
...
-        sample(model, 1, method="schwupdiwupp")
 
-
-def test_validate_wrong_sample(achr, model):
-    """Test sample correctness."""
-    s = achr.sample(10)
-    s["hello"] = 1
     with pytest.raises(ValueError):
-        achr.validate(s)
+        sample(model, 1, method="schwupdiwupp")
```
Midnighter: :rofl:

synchon:
Did a recursive isort on `sampling/` and `test_sampling/`. Not sure why the test is still failing. :(

Midnighter:
You may have to rebase or merge in the latest config changes from devel.

synchon:
Should be good to go.

cdiener:
Nice, looks great.

Issue #823, opened 2019-03-21
Using model tolerance in functions which set threshold in flux results


synchon:
This issue is meant to decide how we handle the functions which typically use arguments like `zero_cutoff` or similar. Here is a list of them (might have missed a few):

- [x] cobra.flux_analysis.loopless.add_loopless
- [x] cobra.flux_analysis.loopless.add_fva_iter
- [x] cobra.flux_analysis.fastcc.fastcc
- [x] cobra.flux_analysis.variability.find_blocked_reactions
- [x] cobra.flux_analysis.phenotype_phase_plane.production_envelope

For now, I see two ways:

1. Set the `zero_cutoff = None` in the function definition and set it to `model.tolerance` inside the function (which would still preserve the definition), or,
2. Replace the `zero_cutoff` with `**kwargs` and let the user know in the docstring.

In any of the cases, there's always the provision for the user to specify the tolerance during function call. Do we keep this or make it only possible to access the tolerance by setting the `model.tolerance` (in which case there would no `zero_cutoff` or `**kwargs`)?

Midnighter:
I vote for the first option that you mention.

cdiener:
So I would even prefer if tolerances can only be set for the model and get used by the functions (meaning removing the zero cutoff parameter). For instance having having a zero cutoff smaller than the tolerance does not make sense since that would give arbitrary results (having cutoff > tolerance does though).

synchon:
In that case, we can just issue a warning and take the `max` of `zero_cutoff` and `tolerance`.

Midnighter:
> In that case, we can just issue a warning and take the max of zero_cutoff and tolerance.

Sounds good. I would not like to change the function signatures too much. So I prefer keeping the argument.

taylo5jm:
I understand that changing the function signatures is undesirable, but it seems to me that removing the `zero_cutoff` argument from these functions would decrease code complexity and increase readability. As it stands now, the logic to check whether or not `zero_cutoff` > `tolerance` has to appear in every function where `zero_cutoff` is used. If that logic were moved to the `Configuration` class, it only has to be in one place.

Midnighter:
[...]

PR #825, opened 2019-03-24
update: use model tolerance for functions setting flux thresholds

synchon:
This PR address #823.

### DIFF
--- a/cobra/flux_analysis/fastcc.py
+++ b/cobra/flux_analysis/fastcc.py
@@ -8,6 +8,9 @@ import logging
 
 from optlang.symbolics import Zero
 
+from cobra.flux_analysis.helpers import normalize_cutoff
+
+
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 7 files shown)


cobra/flux_analysis/fastcc.py
```diff
@@ -50,6 +52,17 @@ def fastcc(model, flux_threshold=1.0, zero_cutoff=1E-09):
            PLoS Comput Biol 10(1): e1003424. doi:10.1371/journal.pcbi.1003424
 
     """
+
+    if zero_cutoff is None:
+        zero_cutoff = model.tolerance
+    else:
+        if zero_cutoff < model.tolerance:
+            warn("zero_cutoff can't be lesser than model.tolerance; "
```
Midnighter:
```suggestion
            warn("zero_cutoff can't be less than model.tolerance; "
```

Midnighter: A `logger.warning` would also be fine, I think.

Midnighter:
@synchon I will add the utility function and also fix isort so that Travis passes. Thank you for the work :smiley:

PR #827, opened 2019-04-01
update: make fastcc iterative

### LINKED ISSUE #806: FastCC implementation lacks parts
#### Problem description

The FastCC method implemented right now is missing some parts. It only implements LP-7 from the paper but does not perform the flipping and the additional iterations to include further reactions. The test passes because it uses a toy model from the paper that resolves in one iteration. But as stated in the publication this is not generally true (for instance with branched pathways). Maybe a more stringent test would be to compare with the output of FVA which can be used for the same purpose. 

#### Code Sample

```python
from cobra.test import create_test_model
from cobra.flux_analysis import flux_variability_analysis, fastcc

mod = create_test_model("ecoli")
fva = flux_variability_analysis(mod, fraction_of_optimum=0.0)
nonblocked = fva.index[(fva.minimum.abs() > 1e-9) | (fva.maximum.abs() > 1e-9)]

fastcc_mod = fastcc(mod)

len(nonblocked) - len(fastcc_mod.reactions)
```

#### Actual Output

965

#### Expected Output

0

#### Dependency Information

Please paste the output of `python -c "from cobra import show_versions;
show_versions()"` in the details block below.

<details>
System Information
==================
OS         Darwin
OS-release 17.7.0
Python      3.6.7

Package Versions
================
cobra        0.14.2
depinfo       1.4.0
future       0.17.1
numpy        1.16.1
optlang       1.4.3
pandas       0.24.1
pip          19.0.2
ruamel.yaml 0.15.88
setuptools   40.8.0
six          1.12.0
swiglpk       1.5.1
tabulate      0.8.3
wheel        0.33.1

</details>

synchon:
This PR fixes #806. For now, it still needs some refinement to be in proper shape.

### DIFF
--- a/cobra/flux_analysis/fastcc.py
+++ b/cobra/flux_analysis/fastcc.py
@@ -7,6 +7,43 @@ from __future__ import absolute_import
 import logging
 
 from optlang.symbolics import Zero
+from six import iteritems
+
+
+def add_fastcc_cons_and_vars(model, flux_threshold):
+    """Add constraints and variables for FASTCC."""
+
+    obj_vars = []
+    vars_and_cons = []
+    prob = model.problem
+
+    for rxn in model.reactions:
+        var = prob.Variable("auxiliary_{}".format(rxn.id),
+                            lb=0.0, ub=flux_threshold)
+        const = prob.Constraint(rxn.forward_variable +
+                                rxn.reverse_variable -
+                                var, name="constraint_{}".format(rxn.id),
+                                lb=0.0)
+        vars_and_cons.extend([var, const])
+        obj_vars.append(var)
+
+    model.add_cons_vars(vars_and_cons)
+    model.objective = prob.Objective(Zero, sloppy=True)
+    model.objective.set_linear_coefficients({v: 1.0 for v in obj_vars})
+
+
+def flip_coefficients(model):
+    """Flip the coefficients for optimizing in reverse direction."""
+
+    for rxn in model.reactions:
+        const = model.constraints.get("constraint_{}".format(rxn.id))
+        var = model.variables.get("auxiliary_{}".format(rxn.id))
+        coefs = const.get_linear_coefficients(const.variables)
+        const.set_linear_coefficients({k: -v for k, v in iteritems(coefs) if k is not var})
+
+    objective = model.objective
+    objective_coefs = objective.get_linear_coefficients(objective.variables)
+    objective.set_linear_coefficients({k: -v for k, v in iteritems(objective_coefs)})
 
 from cobra.flux_analysis.helpers import normalize_cutoff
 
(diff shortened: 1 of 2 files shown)


synchon:
@cdiener @Midnighter I'm still not able to figure out the solution. Could you please help me out?

cobra/flux_analysis/fastcc.py
```diff
@@ -59,28 +100,26 @@ def fastcc(model, flux_threshold=1.0, zero_cutoff=None):
...
-        model.objective = prob.Objective(Zero, sloppy=True, direction="max")
-        model.objective.set_linear_coefficients({v: 1.0 for v in obj_vars})
-
-        sol = model.optimize()
-
-    rxns_to_remove = sol.fluxes[sol.fluxes.abs() < zero_cutoff].index
+        add_fastcc_cons_and_vars(model, flux_threshold)
+
+        # TODO: randomly set to 3; should decide on a default
+        for _ in range(3):
```
cdiener: This should run until no more reactions can be made to carry flux. Also this is missing the main part of FastCC which is setting objective coefficients for only those reactions that have not yet been marked as non-blocked. So basically you would need an additional function that updates the objective and takes a list of reactions to either include or exclude in the objective. Then pass in the the set of reactions that have not been flagged yet in each iteration.

cobra/flux_analysis/fastcc.py
```diff
@@ -59,28 +100,26 @@ def fastcc(model, flux_threshold=1.0, zero_cutoff=None):
...
+            sol = model.optimize(objective_sense="max")
+            rxns_to_remove.extend(sol.fluxes[sol.fluxes.abs() <
```
cdiener: This seems odd. Shouln't it be the other way around, meaning you add reactions to a set of non-blocked reactions in each step? So basically something like

```Python
rxns_to_keep.extend(sol.fluxes.index[sol.fluxes.abs() > zero_cutoff])
```

cobra/flux_analysis/fastcc.py
```diff
@@ -59,28 +100,26 @@ def fastcc(model, flux_threshold=1.0, zero_cutoff=None):
...
+                                             zero_cutoff].index)
+
+            flip_coefficients(model)
+
+            sol = model.optimize(objective_sense="min")
```
cdiener: Maybe put that into `flip_coefficients` as well.

cobra/flux_analysis/fastcc.py
synchon: Ah, I see the point I have been missing. I'll give the paper another read.

synchon:
@cdiener Do the changes look good to you?

cdiener:
Could you add a test like the one in the original issue? Would help to avoid regressions.

Issue #580, opened 2017-08-21
Writing universal model fails


Midnighter:
#### Problem description

I don't have time right now to dig deeper but loading a universal model via cameo and trying to write it to SBML currently fails.

#### Code Sample

```python
import cameo
uni = cameo.models.universal.metanetx_universal_model_bigg
from cobra.io import write_sbml_model
write_sbml_model(uni, "universal_mnx.xml.gz")
```

#### Actual Output

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-db62af2b3ca0> in <module>()
----> 1 write_sbml_model(uni, "universal_mnx.xml.gz")

~/.virtualenvs/fumarate/lib/python3.5/site-packages/cobra/io/sbml3.py in write_sbml_model(cobra_model, filename, use_fbc_package, **kwargs)
    715         return
    716     # create xml
--> 717     xml = model_to_xml(cobra_model, **kwargs)
    718     write_args = {"encoding": "UTF-8", "xml_declaration": True}
    719     if _with_lxml:

~/.virtualenvs/fumarate/lib/python3.5/site-packages/cobra/io/sbml3.py in model_to_xml(cobra_model, units)
    482                              hasOnlySubstanceUnits="false")
    483         set_attrib(species, "name", met.name)
--> 484         annotate_sbml_from_cobra(species, met)
    485         set_attrib(species, "compartment", met.compartment)
    486         set_attrib(species, "fbc:charge", met.charge)

~/.virtualenvs/fumarate/lib/python3.5/site-packages/cobra/io/sbml3.py in annotate_sbml_from_cobra(sbml_element, cobra_element)
    248         if isinstance(identifiers, string_types):
    249             identifiers = (identifiers,)
--> 250         for identifier in identifiers:
    251             li = SubElement(bag, ns("rdf:li"))
    252             set_attrib(li, "rdf:resource", "http://identifiers.org/%s/%s" %

TypeError: 'float' object is not iterable
```

#### Expected Output

Should just work.

#### Output of `cobra.show_versions()`

<details>

System Information
==================
OS                     Linux
OS-release  4.4.0-92-generic
Python                 3.5.2

Package Versions
================
pip                    9.0.1
setuptools            36.2.7
cobra                  0.8.2
future                0.16.0
swiglpk                1.4.4
optlang                1.2.1
ruamel.yaml          0.14.12
pandas                0.20.3
numpy                 1.13.1
[...]

Issue #834, opened 2019-04-12
cobra.io.read_sbml_model model loading errors (objective function)


matthiaskoenig:
There was a bug reported in reading SBML models from
 http://www.plantphysiol.org/content/suppl/2016/07/02/pp.16.00593.DC1
Details in
https://github.com/opencobra/schema/issues/13
Opening this here to be able to track and close the issue.

> Hi everybody,
> I'am new in the CBM using Cobrapy. So I tried to read an SBML model of chlorella available in open source, however when I run this commande on a python script I have plenty of errors that I don't understand.
> CODE
> import numpy as np
> import cobra
> import os
> from os.path import join
> Auto_Model = cobra.io.read_sbml_model(join("iCZ843_Model_Heterotrophy.xml"))
> 
> ERRORS
> WARNING:cobra.io.sbml:Model does not contain SBML fbc package information.
> Academic license - for non-commercial use only
> WARNING:cobra.io.sbml:SBML package 'layout' not supported by cobrapy,information is not parsed
> WARNING:cobra.io.sbml:SBML package 'render' not supported by cobrapy,information is not parsed
> WARNING:cobra.io.sbml:Use of FORMULA in the notes element is discouraged, use fbc:chemicalFormula instead: <Species M_10fthf_c "10-Formyltetrahydrofolate">
> WARNING:cobra.io.sbml:Use of FORMULA in the notes element is discouraged, use fbc:chemicalFormula instead: <Species M_10fthf_h "10-Formyltetrahydrofolate">

Issue #830, opened 2019-04-10
SBOterms are duplicated when stored in SBML annotation and sbo attribute


matthiaskoenig:
#### Problem description
In SBML SBO terms can be defined either via the sboTerm attribute on SBase or using RDF annotations. Some model use both methods. The SBML parser creates duplicate entries for such sboTerms.

E.g. see the following metabolites which have duplicate SBO.
```
 <species metaid="meta_glc" sboTerm="SBO:0000247" id="glc" name="glucose" compartment="c" initialConcentration="5" substanceUnits="mmole" hasOnlySubstanceUnits="false" boundaryCondition="false" constant="false" fbc:charge="0" fbc:chemicalFormula="C6H12O6">
        <annotation>
          <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:vCard="http://www.w3.org/2001/vcard-rdf/3.0#" xmlns:vCard4="http://www.w3.org/2006/vcard/ns#" xmlns:bqbiol="http://biomodels.net/biology-qualifiers/" xmlns:bqmodel="http://biomodels.net/model-qualifiers/">
            <rdf:Description rdf:about="#meta_glc">
              <bqbiol:is>
                <rdf:Bag>
                  <rdf:li rdf:resource="http://identifiers.org/sbo/SBO:0000247"/>
                  <rdf:li rdf:resource="http://identifiers.org/chebi/CHEBI:4167"/>
                  <rdf:li rdf:resource="http://identifiers.org/kegg.compound/C00031"/>
                </rdf:Bag>
              </bqbiol:is>
            </rdf:Description>
          </rdf:RDF>
        </annotation>
      </species>
```
Resulting in 
```
glc (c) : C6H12O6, 0, {'sbo': ['SBO:0000247', 'SBO:0000247'], 'kegg.compound': 'C00031', 'chebi': 'CHEBI:4167'}
```
This will be fixed as part of #684 as soon as there is a proper storage of annotation in cobrapy. Reporting this as separate issue to not forget about it.

Issue #826, opened 2019-04-01
write_sbml_model change cobrapy model's gene reaction rule


hvdinh16:
Hi,

When I used cobra.io.write_sbml_model for a specific reaction with this complicated gene reaction rule:
(rtm274 or rtm278) and rtm249 and (rtm199 or rtm194) and (((rt3401 or rt3160) and rt0797 and rt3417 and rt4138 and (rt2984 or rt4907) and rt2949 and rt4639 and rt0434 and rt5260 and rt2317) or (rt3401 and rt3794 and rt2577 and rt1326 and rt5312 and rt0353))

Reloading the .xml file will give me this gene reaction rule:
rtm249

Best,
Hoang

matthiaskoenig:
This is most likely a bug. Could you provide the model which shows the issue so I can debug the problem.

hvdinh16:
Here are the JSON model and the XML written by cobrapy provided below:
https://github.com/maranasgroup/iRhto_memote/blob/master/model_files/iRhtoC.json
https://github.com/maranasgroup/iRhto_memote/blob/master/model_files/iRhtoC.xml

matthiaskoenig:
@hvdinh16 Sorry for the late reply. just back from holidays and will fix the issue within the next few days.

hvdinh16:
Hi,

As indicated in #835 , this issue #826 is fixed. However, I tested them and there are still bug in the write_sbml_model. Below are the scripts that I used to test and here's the model ([iRhto1108C.zip](https://github.com/opencobra/cobrapy/files/3336073/iRhto1108C.zip))



```python
import cobra
import pandas as pd
import numpy as np
from tqdm import tqdm

from input.custom_functions import *
```


```python
config = cobra.Configuration()
config.solver = 'cplex'
```


```python
model = cobra.io.load_json_model('./iRhto1108C.json')
```


```python
len(model.genes)
```




    1111




```python
cobra.io.write_sbml_model(model, './iRhto1108C.xml')
```


```python
model_sbml = cobra.io.read_sbml_model('./iRhto1108C.xml')
```


```python
len(model_sbml.genes)
```




    1111




```python
rxns_bug = []
for rxn in model.reactions:
    rxn_sbml = model_sbml.reactions.get_by_id(rxn.id)
    if rxn.gene_reaction_rule != rxn_sbml.gene_reaction_rule:
        rxns_bug.append(rxn.id)
        print rxn.id
        print rxn.gene_reaction_rule
[...]

PR #835, opened 2019-04-12
Sbml parser bugfixes

matthiaskoenig:
Ton of SBML bugfixes. Things are fixed, but still need to write some additional tests for the issues to avoid regression.

### DIFF
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -1336,6 +1336,7 @@ def _sbase_annotations(sbase, annotation):
     # standardize annotations
     annotation_data = deepcopy(annotation)
     for key, value in annotation_data.items():
+        # handling of non-string annotations (e.g. integers)
         if isinstance(value, (float, int)):
             value = str(value)
         if isinstance(value, string_types):
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -148,7 +148,8 @@ F_REPLACE = {
 # -----------------------------------------------------------------------------
 # Read SBML
 # -----------------------------------------------------------------------------
-def read_sbml_model(filename, number=float, f_replace=F_REPLACE, **kwargs):
+def read_sbml_model(filename, number=float, f_replace=F_REPLACE,
+                    set_missing_bounds=False, **kwargs):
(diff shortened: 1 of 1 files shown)


matthiaskoenig:
This is ready for merging. Just a list of bugfixes and added unittests for the sbml parser.
Fixes #580, #648, #724, #826, #830, #834, #836, #837 
@Midnighter and @cdiener Could you review this and merge in devel?

matthiaskoenig:
@cdiener or @Midnighter Could you review/merge this? Would be great to bring this the users who submitted the issue reports. This are all only minor fixes with a few lines of code handling some edge cases or additional cases. Not much to review, nothing big or breaking in here.

cobra/io/sbml.py
```diff
@@ -1238,8 +1254,8 @@ def _sbase_notes_dict(sbase, notes):
 
 In the current stage the new annotation format is not completely supported yet.
 """
-URL_IDENTIFIERS_PATTERN = r"^http[s]{0,1}://identifiers.org/(.+)/(.+)"
-URL_IDENTIFIERS_PREFIX = r"https://identifiers.org"
+URL_IDENTIFIERS_PATTERN = re.compile(r"^http[s]{0,1}://identifiers.org/(.+?)/(.+)")  # noqa: E501
```
Midnighter: I was wondering, do we really need a regex here? How about [`urlsplit`](https://docs.python.org/3/library/urllib.parse.html?highlight=urlsplit#urllib.parse.urlsplit)? (Careful, it has a different location in Python 2.) If you do feel that the regex is easier I suggest below change.

```suggestion
URL_IDENTIFIERS_PATTERN = re.compile(r"^https?://identifiers.org/(.+?)/(.+)")  # noqa: E501
```

matthiaskoenig:
@Midnighter thanks. I will update the regex with the next round of sbml bugfixes (which will come for sure). 
I was worried about speed with regex and did some speed testing on the regex compared to `string.split` and did not see any differences. It simplified the code and did not want to introduce some py2 and py3 code with py2 end of lifetime around the corner.

PR #841, opened 2019-04-19
Add find_active_reactions.py to find active reactions quicker

shjchan:
Add `find_active_reactions.py` and a `test_find_active_reactions.py`.
Had some discussion with @ChristianLieven and @Midnighter before. Said it might be nice to add these functions for finding active reactions (primarily for finding reactions in loops) which solve an MILP or a few LPs instead of doing FVA. From tests on a few different models, comparing to using `cobra.flux_analysis.find_blocked_reactions.py`, it can be 3 - 5x faster. Would be good to suggest ways to streamline the code and improve speed. I have been using `cobrapy` from time to time but not very into its infrastructure. The test may need changes too.

### DIFF
--- a/cobra/flux_analysis/find_active_reactions.py
+++ b/cobra/flux_analysis/find_active_reactions.py
@@ -5,7 +5,7 @@ or a (small) number of LP problems
 
 """
 import logging
-from cobra.flux_analysis.variability import flux_variability_analysis
+from cobra.flux_analysis.loopless import fastSNP
 from cobra.flux_analysis.helpers import normalize_cutoff
 from optlang import Model, Variable, Constraint, Objective
 from optlang.symbolics import Zero
@@ -13,32 +13,6 @@ from scipy.linalg import orth
 import numpy as np
 
 LOGGER = logging.getLogger(__name__)
-
-
-def relax_model_bounds(model, bigM=1e4):
-    """
-    Relax all upper and lower bounds in the model.
-    All positive upper bounds will become bigM.
-    All negative lower bounds will become -bigM.
-    All positive lower bounds and negative upper bounds will become zero.
-
-    Parameters
-    ----------
-    model: cobra.Model
-        cobra model. It will *not* be modified.
-    bigM: float, optional
-        a large constant for relaxing the model bounds, default 1e4.
-
-    Returns
-    -------
-    cobra.Model
-        cobra model with relaxed bounds
-
-    """
-
-    for r in model.reactions:
-        r.upper_bound = bigM if r.upper_bound > 0 else 0
-        r.lower_bound = -bigM if r.lower_bound < 0 else 0
 
 
 def find_active_reactions(model, bigM=10000, zero_cutoff=None,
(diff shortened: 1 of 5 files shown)


cdiener:
It seems that you describe a full loop-removal procedure in the paper. If that does not have the same drawbacks as CycleFreeFlux it may be worthwhile to switch to yours. Could you use it for loopless FVA as well?

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,583 @@
...
+    """
+
+    for r in model.reactions:
+        r.upper_bound = bigM if r.upper_bound > 0 else 0
+        r.lower_bound = -bigM if r.lower_bound < 0 else 0
+
+
+def find_active_reactions(model, bigM=10000, zero_cutoff=None,
+                          relax_bounds=True, solve="lp", verbose=False):
+    """
```
cdiener: Rather than using a verbose flag and print, use the logging module as done in other modules and output the additional messages with INFO or DEBUG. That allows user to set the verbosity globally rather than individually for each function.

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,583 @@
...
+                     for r in model.reactions])
+    if max_bound < float("inf"):
+        bigM = max(bigM, max_bound)
+
+    if relax_bounds and zero_cutoff is None:
+        eps = 1e-5
+    else:
+        eps = normalize_cutoff(model, zero_cutoff)
+
+    eps = max(eps, model.solver.configuration.tolerances.feasibility * 100)
```
cdiener: I think it may still be good to use `model.tolerance' here. Not all tolerances (integrality, feasibility, optimality) are implemented for all solvers (for instance GLPK has no optimality tolerance, they use different measures).

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,583 @@
...
+
+                if verbose:
+                    print("Solve LP #%d:  " % iter +
+                          "%d new active rxns found" % n_new)
+
+    return active_rxns
+
+
+def find_reactions_in_cycles(model, bigM=10000, zero_cutoff=1e-1,
+                             relax_bounds=True, solve="lp", verbose=False):
```
cdiener: Since it is exported it should get a docstring.

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,583 @@
...
+                    rxns_to_check.remove(r)
+
+                if verbose:
+                    print("Solve LP #%d:  " % iter +
+                          "%d new active rxns found" % n_new)
+
+    return active_rxns
+
+
+def find_reactions_in_cycles(model, bigM=10000, zero_cutoff=1e-1,
```
cdiener: Why does that zero_cutoff not default to model.tolerance as before?

cobra/test/test_flux_analysis/test_find_active_reactions.py
```diff
@@ -0,0 +1,94 @@
...
+    # assert set(active_rxns_fastsnp) == set(active_rxns)
+
+
+def test_find_reactions_in_cycles(large_model, all_solvers):
+    """Test find_reactions_in_cycles."""
+
+    large_model.solver = all_solvers
+    # solve LPs
+    rxns_in_cycles_lp = find_reactions_in_cycles(large_model)
+    # solving MILP or Fast-SNP may not be feasible for some solvers
```
cdiener: You can make tests conditional on the available solvers. We can help you with that or you could look through some of the other tests for some examples.

cdiener (changes requested):
Good style, some minor initial comments for now :smile:

shjchan:
> It seems that you describe a full loop-removal procedure in the paper. If that does not have the same drawbacks as CycleFreeFlux it may be worthwhile to switch to yours. Could you use it for loopless FVA as well?

From what I understand, CycleFreeFlux is a post-optimization loop-removal procedure, which does not guarantee optimality. I guess it is particularly true for the reactions in loops. The method in my paper guarantee optimality as the original loop-law constraints and usually with much less binary variables, thus faster.

cobra/test/test_flux_analysis/test_find_active_reactions.py
shjchan: Could you refer me to one of the examples? Have made other changes.

shjchan:
Any idea why the test is stuck at `test_fastcc.py`? I remember that when I just made the pull request, it ran through it.

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,602 @@
...
+    """
+    Relax all upper and lower bounds in the model.
+    All positive upper bounds will become bigM.
+    All negative lower bounds will become -bigM.
+    All positive lower bounds and negative upper bounds will become zero.
+
+    Parameters
+    ----------
+    model: cobra.Model
+        cobra model. It will *not* be modified.
```
Midnighter: You say the model is not modified but this function changes the bounds on every reaction so that seems a little misleading.

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,602 @@
...
+        feasbility, optimality and integrality
+
+    Returns
+    -------
+    list
+        List of reaction IDs which can carry flux.
+
+    Notes
+    -----
+    The optmization problem solved is as follow:
```
Midnighter: It'd be really neat to use the Sphinx directives so that this is rendered as proper math. See [here](https://stackoverflow.com/a/11226398).

Midnighter (commented):
I didn't have the time yet to go through the entire code. Out of curiosity, how is your algorithm affected by (how does it treat) reactions that are unbounded in one direction? Are they then also intended to be bounded by big M?

shjchan:
> I didn't have the time yet to go through the entire code. Out of curiosity, how is your algorithm affected by (how does it treat) reactions that are unbounded in one direction? Are they then also intended to be bounded by big M?

They are bounded by either big M (if relax_bounds=True) or the original bounds (if relax_bounds=False). The additional constraints for irreversible reactions are actually similar to fastcc while for the irreversible reactions are a little different.

cobra/flux_analysis/find_active_reactions.py
shjchan: Good catch. Just copied and pasted. Fixed.

shjchan:
> > It seems that you describe a full loop-removal procedure in the paper. If that does not have the same drawbacks as CycleFreeFlux it may be worthwhile to switch to yours. Could you use it for loopless FVA as well?
> 
> From what I understand, CycleFreeFlux is a post-optimization loop-removal procedure, which does not guarantee optimality. I guess it is particularly true for the reactions in loops. The method in my paper guarantee optimality as the original loop-law constraints and usually with much less binary variables, thus faster. The functions here are only for the first step of the whole localized-loopless-FVA procedure, i.e., to find a minimal feasible null space. The rest of the implementation will need some significant coding, which I hope I can do some time later.

Good news: the Fast-SNP which I also implemented in this pull request, can be readily used to largely reduce the time for loopless FVA already. I reorganized the newly added functions and added a `method` options to `flux_variability_analysis` (just a few lines). Tested on E. coli core model, results exactly the same. Tested on some reactions in iJO1366, around 40x time reduction (16xx sec --> 40 sec).

cdiener:
> Good news: the Fast-SNP which I also implemented in this pull request, can be readily used to largely reduce the time for loopless FVA already. I reorganized the newly added functions and added a method options to flux_variability_analysis (just a few lines). Tested on E. coli core model, results exactly the same. Tested on some reactions in iJO1366, around 40x time reduction (16xx sec --> 40 sec).

This sounds great. If I find the time I will give it a go with the problematic example from #698. If it handles that well I think the way to go would be to drop CycleFreeFlux in favor of your method.

shjchan:
@cdiener great. I am pretty sure that it will work fine. By the way, if you have time, could you advise me what goes wrong in the test? It just froze after test_fastcc. I couldn't figure why. I don't think I have changed anything related to that. And I could finish the run if I ran pytest locally.

Midnighter (changes requested):
Still need to read your paper in detail but made some comments on the code already. Looks promising :smiley:

cobra/flux_analysis/find_active_reactions.py
```diff
@@ -0,0 +1,434 @@
...
+        feas_tol = model.solver.configuration.tolerances.feasibility
+        eps = max(eps, feas_tol * 100)
+    except:
+        feas_tol = eps
+        LOGGER.debug("Feasibility tolerance not defined")
+
+    if solve == "milp":
+        try:
+            # ensure bigM*z << eps at integrality tolerance limit
+            eps = max(eps, model.solver.configuration.tolerances.integrality *
```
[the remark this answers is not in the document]
shjchan: Thanks. But here it is important for `eps` to be at least 10x larger then the integrality tolerance if solving MILP. Otherwise there could be integrality violation to satisfy the constraint so that some reactions do not necessarily have flux.
And I was not sure if it is always defined for every solver (i.e. `model.configurationas.tolerances.integrality` is always an attribute), so I added `try`/`except` here hoping to see if that might help pass the test. But it was probably not the issue and is not needed, correct? Is `model.tolerance` not less than integrality tolerance if it is used? Any suggestion? Just want to make sure it is large enough when solving MILP and when integrality is used.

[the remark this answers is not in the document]
Midnighter: Ah, I see. We have simplified the tolerance interface in cobrapy. So setting the `model.tolerance` attempts to set all available tolerances given by the solver. Thus it is also expected that the `model.tolerance` value is the same one for integrality, feasiblity, etc. So yes, using `model.tolerance` _should_ do the right thing and also not raise any exceptions.

shjchan:
I think I have addressed all comments and have restructured the code so that it is more pythonic.

cdiener:
Sorry for the delay, just saw it and will go through it again. Will probably take me a little since it is a lot of code changes. From a quick look I don't see any changes to flux_variability_analysis as mentioned above and we will need more tests.

cdiener (changes requested):
Sorry for the long delay. Most of it being compliant with pep8 and numpy docstring format as well as rebasing on the latest devel branch. However, the method also does not seem to pass its own tests. Also the tests should also compare your implementation to the reference implementation (method="original" in add_loopless) for a smaller model.

Some functionality seems to overlap with the already existing fastcc function. In particular the `find_active_reactions` function. It seems to be very similar things.

However, `in_cycles` version looks very useful and it would be great to add it to FVA. Is it guranteed to remove all cycles from the solution?

I can help out with fixing the the style issues if you want but would leave the methods to you if that is all right.

PR #852, opened 2019-06-03
SBML bugfixes

### LINKED ISSUE #845: Read models with invalid SBML identifiers
Many models in the field are just invalid SBML. One major reason for this are invalid SBML identifiers (which will create a multitude of problems). Currently, models with invalid SBML ids are not read in cobrapy.
The parser could be more robust and give such invalid models a chance, by fixing the ids on import.

### LINKED ISSUE #851: Invalid GeneProduct (and probably Gene ids) don't give Error on read_sbml_model
Currently, only species and reactions provide information on invalid SBML as Error on reading. This must be extended to all SBase objects, otherwise invalid ids are are dropped which affects the model. 

See more detailed discussion in #850.

matthiaskoenig:
Fix #845 and Fix #851 
Correct reporting of invalid SBML ids in error log and reading of models with invalid SBML ids.

### DIFF
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -1,5 +1,5 @@
 """
-SBML import and export using python-libsbml(-experimental).
+SBML import and export using python-libsbml.
 
 - The SBML importer supports all versions of SBML and the fbc package.
 - The SBML exporter writes SBML L3 models.
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -57,7 +57,6 @@ class CobraSBMLError(Exception):
     """ SBML error class. """
     pass
 
-
 LOGGER = logging.getLogger(__name__)
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


cobra/io/sbml.py
```diff
@@ -312,12 +312,15 @@ def _sbml_to_model(doc, number=float, f_replace=F_REPLACE,
                 raise Exception("Conversion of SBML fbc v1 to fbc v2 failed")
 
     # Model
-    cobra_model = Model(model.getId())
+    model_id = model.getIdAttribute()
+    if not libsbml.SyntaxChecker.isValidSBMLSId(model_id):
+        logging.error("Model id '%s' is no valid SId." % model_id)
```
cdiener: Change to `logging.error("'%s' is not a valid SBML 'SId'." % value)` as used below.

matthiaskoenig:
Included the proposed change. Thanks for code review. As soon as tests have passed this is ready to merge in develop.

matthiaskoenig:
Not sure what is going on with appveyor. Looks like some issues with scipy on windows and py27, but nothing related to this pull request as far as I can tell.
Does anyone have a clue how to fix this without limiting scipy versions?

cdiener:
Looks like it tries to install a scipy version that does not support Python 2. This is odd since we don't depend on scipy and it should not be installed. Seems to be an issue with tox or the tox config @Midnighter

Midnighter:
I just fixed AppVeyor on memote, I'll take a look at it.

cdiener:
@matthiaskoenig did you want to add something else here? Otherwise I'll merge.

Midnighter:
I think only AppVeyor failing was stopping this PR.

PR #855, opened 2019-06-16
refactor: introduce Summary class

### LINKED ISSUE #586: Create a Summary class
I suggest creating a `Summary` class that is instantiated and returned by `Model.summary`, `Reaction.summary`, and `Metabolite.summary`.

Primary benefits:
* A class instance can have multiple representation methods, e.g., `__repr__` and `_repr_html_`.
* The information of the summary is not merely displayed but can be made accessible via a `to_frame` method, for example.

I don't see any downside other than having to fiddle with preferably the same class being returned by all three summary methods. If that doesn't work out they should be subclasses.

synchon:
This PR is supposed to fix #586.
- [x] introduce Summary class
- [x] introduce required subclasses
- [x] improve ~~`__repr__` and~~ `_repr_html_`
- [x] add facility to work with the summary data


cobra/flux_analysis/summary.py
```diff
@@ -20,16 +22,145 @@
...
-    This method requires the model for which this metabolite is a part
-    to be solved.
+    def __init__(self, solution, threshold, fva, names, floatfmt):
+        self.solution = solution
+        self.threshold = threshold
+        self.fva = fva
+        self.names = names
+        self.floatfmt = floatfmt
+
+    def _generate():
```
matthiaskoenig: - `self` argument missing, i.e., `_generate(self):`
- perhaps mention that this is abstract and is implemented by the respective summaries

cobra/flux_analysis/summary.py
```diff
@@ -1,12 +1,14 @@
 # -*- coding: utf-8 -*-
 
+"""Define the Summary class."""
+
 from __future__ import absolute_import, division
```
matthiaskoenig: Not sure this should be located in `flux_analysis`. The summary class provides information on the core objects, not necessary part of the flux_analysis.

matthiaskoenig:
Not a full code review, but just some basic checks.

Not sure what the expected behavior of the summary class should be, but I expect it should print the summary as stated for instance on reaction:

```
    def summary(self, names=False):
        """
        Print a summary of the reaction.

        Parameters
        ----------
        names : bool, optional
            Emit reaction and metabolite names rather than identifiers (default
            False).

        """

        from cobra.flux_analysis.summary import ReactionSummary
        return ReactionSummary(rxn=self, names=names)
```
At the moment the `summary()` does not print the summary, but returns a ReactionSummary object.
I expect to see all the summary information printed if I do something like
```
import cobra
model = read_sbml_model("models/e_coli_core.xml")
df = model.summary()
model.summary()
for r in model.reactions:
    r.summary()
for m in model.metabolites:
    m.summary()
```
but nothing is printed at the moment (same for the other summaries). Not sure what the correct behavior is here, but not printing things at the moment.

In addition,  there should be an argument in the `summary` methods which lets choose between `table` or `frame`.

cobra/flux_analysis/summary.py
synchon: Yes will add the `self` argument and also update the docstrings as well.

cobra/flux_analysis/summary.py
synchon: I wasn't sure on this as well so I just let it be there. I was hoping we could discuss and come up a proper place for this. I also have another concern: shouldn't be the subclasses implemented in separate modules? I think that would increase the maintainability.

synchon:
> Not sure what the correct behavior is here, but not printing things at the moment.

In my opinion, you would generally use `summary` methods when you are in a notebook-like environment to do some interactive analyses. So, `model.summary()` would give you something like this (in Jupyter notebook): 
<table border="1" class="dataframe">  <thead>    <tr>      <th></th>      <th colspan="2" halign="left">IN_FLUXES</th>      <th colspan="2" halign="left">OUT_FLUXES</th>      <th colspan="2" halign="left">OBJECTIVES</th>    </tr>    <tr>      <th></th>      <th>ID</th>      <th>FLUX</th>      <th>ID</th>      <th>FLUX</th>      <th>ID</th>      <th>FLUX</th>    </tr>  </thead>  <tbody>    <tr>      <th>0</th>      <td>o2_e</td>      <td>21.7995</td>      <td>h2o_e</td>      <td>29.1758</td>      <td>Biomass_Ecol...</td>      <td>0.873922</td>    </tr>    <tr>      <th>1</th>      <td>glc__D_e</td>      <td>10</td>      <td>co2_e</td>      <td>22.8098</td>      <td></td>      <td></td>    </tr>    <tr>      <th>2</th>      <td>nh4_e</td>      <td>4.76532</td>      <td>h_e</td>      <td>17.5309</td>      <td></td>      <td></td>    </tr>    <tr>      <th>3</th>      <td>pi_e</td>      <td>3.2149</td>      <td></td>      <td></td>      <td></td>      <td></td>    </tr>  </tbody></table>

Also, I don't know if you would want a table output in `__repr__`. For that you can use `model.summary().to_table()`. The thing that I was debating in my head was whether `to_frame` and `to_table` should be property or not.

> In addition, there should be an argument in the summary methods which lets choose between table or frame.

You can use `to_table` and `to_frame` to get the required output.

Also, I should probably add a new section on `Summary` in the docs.

cobra/flux_analysis/summary.py
```diff
@@ -51,106 +197,208 @@ def metabolite_summary(met, solution=None, threshold=0.01, fva=False,
...
+
+        flux_summary.loc[~flux_summary.is_input, 'percent'] = \
+            flux_summary.loc[~flux_summary.is_input, 'flux'] / total_flux
+
+        flux_summary['percent'] = flux_summary.percent.\
+            apply(lambda x: '{:.0%}'.format(x))
+
+        return flux_summary
+
+    def to_table(self):
```
cdiener: I feel like that should be implemented via `object.__str__` and return a string and not print anything. This would extend to the `print` method automatically that way.

synchon: I agree on the point of functions not having print statements as side effects. The print statements are remnants of the old code and I kept it as it is. I feel the only downsides of implementing it in `__str__` is that it would bloat the method. We can probably return a string from `to_table` though.

Midnighter:
I agree with @cdiener that the class should implement the `__repr__` and `__str__` methods.

synchon:
I do agree that `__repr__` and `__str__` makes sense for the class. You would want them to give you a better human-readable representation. But, if we make them give the actual summary, wouldn't it be too much for a representation?

Midnighter:
I don't think so. In fact, it might be very easy to fall back to pandas representation both in the HTML and the normal string case since it already produces a fairly nice output in both cases.

synchon:
That seems the best option. If we go with this, it will totally break the previous string summary format and would be a fair amount of change. But, on the other hand, it would be much easier to maintain. Are we okay with the change though?

cobra/flux_analysis/summary.py
synchon: @Midnighter @cdiener Shall we shift this to core as @matthiaskoenig suggests? Also, any thoughts on the splitting of subclasses?

cdiener:
This has been discussed before https://stackoverflow.com/questions/1436703/difference-between-str-and-repr. I would go with the consensus from there an overwrite the `__str__` method but not the `__repr__`.

matthiaskoenig:
> That seems the best option. If we go with this, it will totally break the previous string summary format and would be a fair amount of change. But, on the other hand, it would be much easier to maintain. Are we okay with the change though?

Yes, please break it. This is the chance to make a proper update of the summary behavior. Especially, because breaking just means it looks different (nobody, at least I hope so, will use the former printed summary for any kind of downstream analysis). So it should not break any code, but only look different.

If one prints a summary one expects to see the human-readable summary. I am for overwriting `__str__`.

Not sure where the module belongs. It would put it in `core` where also other `core` helper modules are located, namely `dictlist.py`, `singleton.py`, `formula.py`. In my opinion a much better location than `flux_analysis`. And it is summary for core objects. Main thing is to find things easily, I would look in `core` or perhaps `util` (which is just a collection of misc stuff) for the summary so why not put it in `core` (no harm done).

Midnighter:
> nobody, at least I hope so, will use the former printed summary for any kind of downstream analysis

Just our test :laughing:

I also vote for it going into `cobra.core.summary`.

cobra/core/metabolite.py
```diff
@@ -237,9 +237,11 @@ def summary(self, solution=None, threshold=0.01, fva=None, names=False,
             Format string for floats (default '.3g').
 
         """
-        from cobra.flux_analysis.summary import metabolite_summary
-        return metabolite_summary(self, solution=solution, threshold=threshold,
-                                  fva=fva, names=names, floatfmt=floatfmt)
+
+        from cobra.core.summary import MetaboliteSummary
```
Midnighter: Are you getting cyclic dependencies or why do you import inside the method?

synchon: I was getting cyclic import when it was in `flux_analysis`. I'll check if it is still facing the same issue in `core`, if not I'll place the import at the top.

synchon: The issue persists, so keeping it as it is.

matthiaskoenig:
Just tried to check this, but got an error `ValueError wrong number of arguments passed`:
With this model
http://bigg.ucsd.edu/static/models/e_coli_core.xml
```
import cobra
from IPython.display import display
model = read_sbml_model("./models/e_coli_core.xml")
display(model.summary())
for r in model.reactions:
    display(r.summary())
```

Also it would be great if the summary could be a property. It would save one writing the brackets all the time and allow the easier:
```
display(model.summary)
display(reaction.summary)
display(metabolite.summary)
```
which is also much more readable.

cobra/core/summary.py
```diff
@@ -0,0 +1,621 @@
...
+        flux_summary['percent'] = 0
+        total_flux = flux_summary.loc[flux_summary.is_input, "flux"].sum()
+
+        flux_summary.loc[flux_summary.is_input, 'percent'] = \
+            flux_summary.loc[flux_summary.is_input, 'flux'] / total_flux
+
+        flux_summary.loc[~flux_summary.is_input, 'percent'] = \
+            flux_summary.loc[~flux_summary.is_input, 'flux'] / total_flux
+
+        flux_summary['percent'] = flux_summary.percent.\
```
Midnighter: Again, I think formatting should be part of the output methods.

cobra/core/summary.py
```diff
@@ -0,0 +1,621 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
Midnighter: If you like, I can do this after you have incorporated the other changes.

Midnighter:
> Also it would be great if the summary could be a property. It would save one writing the brackets all the time

I disagree. Properties should be reserved for attribute access with caching or similar side effects. Overall I think cobrapy abuses properties and should use methods much more so that it is clear it is a method call and will have an effect. If anything it should be `.summarize()` to clarify that it is an action.

synchon:
> Just tried to check this, but got an error `ValueError wrong number of arguments passed`:
> With this model
> http://bigg.ucsd.edu/static/models/e_coli_core.xml

I can't reproduce the error since I get proper behavior. Not sure what is wrong in your case.

> Also it would be great if the summary could be a property.

I agree with @Midnighter and I also feel that properties should be used for attribute access only.

cobra/core/metabolite.py
```diff
@@ -216,30 +216,40 @@ def summary(self, solution=None, threshold=0.01, fva=None, names=False,
 
         Parameters
         ----------
-        solution : cobra.Solution, optional
+        solution: cobra.Solution or None, optional (default None)
```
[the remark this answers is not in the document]
synchon: > Also, the default should be provided in the description.

The documentation also states that: _Optional keyword parameters have default values, which are displayed as part of the function signature._ . This is what led me to write it in that fashion. Also, the style is not consistent across cobrapy itself. I feel we should make it consistent to avoid future issues. So, do we stick to what you propose? I feel stating the default in the function signature makes it a bit concise.

cobra/core/summary.py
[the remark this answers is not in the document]
synchon: > Please create a subpackage structure cobra/core/summary/ and split the class definitions into different modules.

Yeah sure.

> If you like, I can do this after you have incorporated the other changes.

I'll do this before moving to the other changes. I think that would ease the process.

cobra/core/summary.py
```diff
@@ -0,0 +1,621 @@
...
+        fraction of the optimum objective to be searched.
+    names : bool
+        Emit reaction and metabolite names rather than identifiers.
+    float_format : one-parameter function
+        Format string for floats.
+
+    """
+
+    def __init__(self, solution, threshold, fva, names, float_format):
+        self.solution = solution
```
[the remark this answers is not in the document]
synchon: Yeah fair point.

cobra/core/summary.py
```diff
@@ -0,0 +1,621 @@
...
+        rxn_name = list()
+        flux = list()
+        reaction = list()
+
+        for rxn in rxns:
+            rxn_id.append(rxn.id)
+            rxn_name.append(format_long_string(emit(rxn), 10))
+            flux.append(self.solution[rxn.id] * rxn.metabolites[self.met])
+            txt = rxn.build_reaction_string(use_metabolite_names=self.names)
+            reaction.append(format_long_string(txt, 40
```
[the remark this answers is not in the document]
synchon: I'll do a quick refactoring and try to rewrite parts which seem weird.

cobra/core/metabolite.py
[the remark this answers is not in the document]
Midnighter: Yes, cobrapy is a patchwork :) Since we decided to use numpydoc, I want like to closely follow it. So whenever we update or otherwise edit a part of cobrapy, we should improve the docstrings where necessary.

matthiaskoenig:
I agree, no property.

I just reproduced the error on a second computer (both linux, py36) in a clean virtualenv
```
mkvirtualenv cobrapy --python=python3.6
(cobrapy) mkoenig@prime1:~/git/cobrapy$ pip install -e .
(cobrapy) mkoenig@prime1:~/git/cobrapy$ pip install ipython
(cobrapy) python summary.py 
```
With `summary.py` being:
```
import cobra
from IPython.display import display
model = read_sbml_model("models/e_coli_core.xml")

r = model.reactions.get_by_id("ACt2r")
display(r)
display(r.summary())
```
Results in
```
ValueError                                Traceback (most recent call last)
~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/managers.py in create_block_manager_from_blocks(blocks, axes)
   1650                 blocks = [make_block(values=blocks[0],
-> 1651                                      placement=slice(0, len(axes[0])))]
   1652 

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/blocks.py in make_block(values, placement, klass, ndim, dtype, fastpath)
   3094 
-> 3095     return klass(values, ndim=ndim, placement=placement)
   3096 

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/blocks.py in __init__(self, values, placement, ndim)
   2630         super(ObjectBlock, self).__init__(values, ndim=ndim,
-> 2631                                           placement=placement)
   2632 

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/blocks.py in __init__(self, values, placement, ndim)
     86                 'Wrong number of items passed {val}, placement implies '
---> 87                 '{mgr}'.format(val=len(self.values), mgr=len(self.mgr_locs)))
     88 

ValueError: Wrong number of items passed 3, placement implies 4

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
~/envs/cobrapy/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/git/cobrapy/cobra/core/summary.py in _repr_html_(self)
    166 
    167     def _repr_html_(self):
--> 168         return self.to_frame()._repr_html_()
    169 
    170 

~/git/cobrapy/cobra/core/summary.py in to_frame(self)
    609 
    610         """
--> 611         return self._generate()
    612 
    613     def _to_table(self):

~/git/cobrapy/cobra/core/summary.py in _generate(self)
    595         rxn_summary = pd.DataFrame(
    596             data=data,
--> 597             columns=columns
    598         )
    599 

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/frame.py in __init__(self, data, index, columns, dtype, copy)
    422             else:
    423                 mgr = init_ndarray(data, index, columns, dtype=dtype,
--> 424                                    copy=copy)
    425 
    426         # For data is list-like, or Iterable (will consume into list)

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/construction.py in init_ndarray(values, index, columns, dtype, copy)
    165         values = maybe_infer_to_datetimelike(values)
    166 
--> 167     return create_block_manager_from_blocks([values], [columns, index])
    168 
    169 

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/managers.py in create_block_manager_from_blocks(blocks, axes)
   1658         blocks = [getattr(b, 'values', b) for b in blocks]
   1659         tot_items = sum(b.shape[0] for b in blocks)
-> 1660         construction_error(tot_items, blocks[0].shape[1:], axes, e)
   1661 
   1662 

~/envs/cobrapy/lib/python3.6/site-packages/pandas/core/internals/managers.py in construction_error(tot_items, block_shape, axes, e)
   1689         raise ValueError("Empty data passed with indices specified.")
   1690     raise ValueError("Shape of passed values is {0}, indices imply {1}".format(
-> 1691         passed, implied))
   1692 
   1693 

ValueError: Shape of passed values is (4, 3), indices imply (4, 4)
```

Dependencies are
```
Package                     Version Location                 
--------------------------- ------- -------------------------
backcall                    0.1.0   
cobra                       0.15.3  /home/mkoenig/git/cobrapy
decorator                   4.4.0   
depinfo                     1.5.1   
future                      0.17.1  
ipython                     7.6.1   
ipython-genutils            0.2.0   
jedi                        0.14.0  
mpmath                      1.1.0   
numpy                       1.16.4  
optlang                     1.4.4   
pandas                      0.24.2  
parso                       0.5.0   
pexpect                     4.7.0   
pickleshare                 0.7.5   
pip                         19.1.1  
pipdeptree                  0.13.2  
prompt-toolkit              2.0.9   
ptyprocess                  0.6.0   
Pygments                    2.4.2   
python-dateutil             2.8.0   
pytz                        2019.1  
ruamel.yaml                 0.15.98 
setuptools                  41.0.1  
six                         1.12.0  
sympy                       1.4     
traitlets                   4.3.2   
wcwidth                     0.1.7   
wheel                       0.33.4 
```

cobra/core/metabolite.py
synchon: `MetaboliteSummary` and `ModelSummary` require some imports from `cobra.core` so that is again in turn becoming cyclic.

synchon:
@matthiaskoenig Okay I checked and the error is showing up for me for the reaction `ACt2r`. If you try some other reaction like `ACALD` it won't show up. I think it's arising due to the way I wrote the internals. I'll fix it ASAP.

synchon:
@Midnighter does the last commit address the basic cosmetic changes that you asked for?

synchon:
@matthiaskoenig I think commit [f2acb27](https://github.com/opencobra/cobrapy/pull/855/commits/f2acb27bb08f38f34d0aa211c7b22e39f2441006) will fix your issue. Please let me know if the issue still persists.

matthiaskoenig:
@synchon Error is fixed with latest release. 
I played a bit with the summary and like it a lot. This is a great improvement in functionality.
From my side this is good.

Small thing I noticed when playing with the summary: Perhaps we should also add the summary on the gene object in addition. The gene summary would than provide the table of all the reactions where the gene is in the GPRs which would give a nice overview of which reactions are affected on gene knockout. Thereby we would have a consistent summary on reaction, metabolite and gene all providing different views on the network/flux solution.

I.e. it would be nice to do the following
```
model = read_sbml_model("models/e_coli_core.xml")
for r in model.reactions:
    print(r)
    display(r.summary())
    
for m in model.metabolites:
    print(m)
    display(m.summary())
    
for g in model.genes:
    print(g)
    print(g.summary())
```

synchon:
> Perhaps we should also add the summary on the gene object in addition.

In my opinion, you can get this same functionality through `model.genes.<gene_id>.reactions`, so will it still make sense to pack it in an interface like `summary`?

Midnighter:
A `summary` method on a gene could be a nice convenience addition but I also think that we should maybe finish up this PR and make a new one for the genes.

synchon:
Working on finishing it ASAP. Will take a few more commits.

Midnighter:
> Working on finishing it ASAP. Will take a few more commits.

I didn't mean to pressure you. We took long enough with reviewing your code in the first place :wink:

synchon:
Still, it's been around a month now. :P

synchon:
This should be good to go now. I made the requested changes and also updated a few internal stuff.

matthiaskoenig:
It would be great to have this in the next release. From my point of view this is ready.

pstjohn:
This code throws an error for me:
```python
>>> import cobra.test
>>> model = cobra.test.create_test_model()
>>> model.metabolites.pyr_c.summary()
...
AttributeError: 'Model' object has no attribute '_tolerance'

```

I think in general this functionality is nice, but one complaint is that by putting everything together in a single frame it can indicate the rows are related. (when in truth they are separate tables)
i.e, the model summaries are OK, but as a user I'd probably rather have 3 objects: 1 for 'in fluxes', 1 for 'out fluxes', and 1 for 'objectives', shown side by side. Especially if later I wanted to be able to manipulate those dataframes.
image

For reactions in particular, the connection between genes and metabolites seems to be intentional, but I'm guessing these are just lists of the genes involved in the reaction:
image
Tbh I never really knew what would go in a reaction summary, so I didn't include them in the original code.

Not sure about the metabolite summaries (I couldn't get them to work) but I'm guessing there you'll probably want to print 2 dataframes (in and out fluxes), but maybe make them available as a single frame with positive input fluxes and negative output fluxes.

Midnighter:
I agree with you @pstjohn we should break the tables up into sensible sub units. I prefer to move ahead with this improvement, though, before making further changes.

PR #866, opened 2019-06-30
Set correct flux bounds 

### LINKED ISSUE #858: `reaction.build_reaction_from_string` sets incorrect flux bounds
Functions which set flux bounds in the model must use the default bounds defined in the configuration object. This is not the case for `reaction.build_reaction_from_string` which sets -1000, 1000.
```
       # reversible case
        arrow_match = reversible_arrow_finder.search(reaction_str)
        if arrow_match is not None:
            self.lower_bound = -1000
            self.upper_bound = 1000
        else:  # irreversible
            # try forward
            arrow_match = forward_arrow_finder.search(reaction_str)
            if arrow_match is not None:
                self.upper_bound = 1000
                self.lower_bound = 0
            else:
                # must be reverse
                arrow_match = reverse_arrow_finder.search(reaction_str)
                if arrow_match is None:
                    raise ValueError("no suitable arrow found in '%s'" %
                                     reaction_str)
                else:
                    self.upper_bound = 0
                    self.lower_bound = -1000
        reactant_str = reaction_str[:arrow_match.start()].strip()
        product_str = reaction_str[arrow_match.end():].strip()
```

MaxGreil:
This PR is supposed to fix #858.

### DIFF
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -25,7 +25,7 @@ from cobra.util.solver import (
 from cobra.util.util import format_long_string
 
 
-CONFIGURATION = Configuration()
+config = Configuration()
 
 # precompiled regular expressions
 # Matches and/or in a gene reaction rule
@@ -88,9 +88,9 @@ class Reaction(Object):
 
         # from cameo ...
         self._lower_bound = lower_bound if lower_bound is not None else \
-            CONFIGURATION.lower_bound
+            config.lower_bound
         self._upper_bound = upper_bound if upper_bound is not None else \
-            CONFIGURATION.upper_bound
+            config.upper_bound
 
     def _set_id_with_model(self, value):
         if value in self.model.reactions:
@@ -1050,12 +1050,12 @@ class Reaction(Object):
         # reversible case
         arrow_match = reversible_arrow_finder.search(reaction_str)
         if arrow_match is not None:
-            self.bounds = (CONFIGURATION.lower_bound, CONFIGURATION.upper_bound)
+            self.bounds = config.lower_bound, config.upper_bound
         else:  # irreversible
             # try forward
             arrow_match = forward_arrow_finder.search(reaction_str)
             if arrow_match is not None:
-                self.bounds = (0, CONFIGURATION.upper_bound)
+                self.bounds = 0, config.upper_bound
             else:
                 # must be reverse
                 arrow_match = reverse_arrow_finder.search(reaction_str)
@@ -1063,7 +1063,7 @@ class Reaction(Object):
                     raise ValueError("no suitable arrow found in '%s'" %
                                      reaction_str)
                 else:
-                    self.bounds = (CONFIGURATION.lower_bound, 0)
+                    self.bounds = config.lower_bound, 0
         reactant_str = reaction_str[:arrow_match.start()].strip()
         product_str = reaction_str[arrow_match.end():].strip()
 
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -1049,17 +1049,13 @@ class Reaction(Object):
 
         # reversible case
         arrow_match = reversible_arrow_finder.search(reaction_str)
-        lower_bound = CONFIGURATION.lower_bound
-        upper_bound = CONFIGURATION.upper_bound
         if arrow_match is not None:
-            self.lower_bound = lower_bound
-            self.upper_bound = upper_bound
+            self.bounds = (CONFIGURATION.lower_bound, CONFIGURATION.upper_bound)
         else:  # irreversible
             # try forward
             arrow_match = forward_arrow_finder.search(reaction_str)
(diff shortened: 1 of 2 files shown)


matthiaskoenig:
Hi @MaxGreil,
this looks good.

Some smaller things I would change
1. use CONFIGURATION.lower_bound directly instead of defining local variables for them.
2. set upper and lower bounds at the same time (not part of issue, but if you touch the code you should update this)

The problem with setting the upper and lower bounds individually, is that there are 2 checks performed (both time it is checked if the bounds fit, whereas setting both at once requires only one check). In addition as far as I remember the constraints in optlang are generated at once, instead after each other (which results in some speedup for many reactions).

I.e. use
```
    @bounds.setter
    @resettable
    def bounds(self, value):
        lower, upper = value
        # Validate bounds before setting them.
        self._check_bounds(lower, upper)
        self._lower_bound = lower
        self._upper_bound = upper
        self.update_variable_bounds()
```
to set the upper and lower bound at the same time directly from configuration, i.e. instead of
```
self.lower_bound = lower_bound
self.upper_bound = upper_bound
```
better
```
self.bounds(CONFIGURATION.lower_bound, CONFIGURATION.upper_bound)
```
Same at the other two locations.

MaxGreil:
Hi @cdiener ,
how should this test look like?

cdiener:
> Hi @cdiener ,
> how should this test look like?

Sets default bounds via configuration, adds a reaction with `build_reaction_from_string` and checks that the set bounds were applied. Could be added below https://github.com/opencobra/cobrapy/blob/374ebfed7345eef6b32d8b7e6311676cb6232b17/cobra/test/test_core/test_core_reaction.py#L238.

cobra/core/reaction.py
```diff
@@ -1050,23 +1050,21 @@ def build_reaction_from_string(self, reaction_str, verbose=True,
...
-            self.upper_bound = 1000
+            self.bounds = (CONFIGURATION.lower_bound,
+                           CONFIGURATION.upper_bound)
         else:  # irreversible
             # try forward
             arrow_match = forward_arrow_finder.search(reaction_str)
             if arrow_match is not None:
-                self.upper_bound = 1000
-                self.lower_bound = 0
+                self.bounds = (0, CONFIGURATION.upper_bound)
```
Midnighter:
```suggestion
                self.bounds = 0, config.upper_bound
```

cobra/core/reaction.py
```diff
@@ -1050,23 +1050,21 @@ def build_reaction_from_string(self, reaction_str, verbose=True,
...
             else:
                 # must be reverse
                 arrow_match = reverse_arrow_finder.search(reaction_str)
                 if arrow_match is None:
                     raise ValueError("no suitable arrow found in '%s'" %
                                      reaction_str)
                 else:
-                    self.upper_bound = 0
-                    self.lower_bound = -1000
+                    self.bounds = (CONFIGURATION.lower_bound, 0)
```
Midnighter:
```suggestion
                    self.bounds = config.lower_bound, 0
```

cobra/core/reaction.py
```diff
@@ -1050,23 +1050,21 @@ def build_reaction_from_string(self, reaction_str, verbose=True,
         # reversible case
         arrow_match = reversible_arrow_finder.search(reaction_str)
         if arrow_match is not None:
-            self.lower_bound = -1000
-            self.upper_bound = 1000
+            self.bounds = (CONFIGURATION.lower_bound,
```
Midnighter:
```suggestion
            self.bounds = config.lower_bound, config.upper_bound
```

cobra/core/reaction.py
```diff
@@ -1050,23 +1050,21 @@ def build_reaction_from_string(self, reaction_str, verbose=True,
...
+                           CONFIGURATION.upper_bound)
```
Midnighter:
```suggestion
```

matthiaskoenig:
@MaxGreil you are almost there. Just write the test and do the minor changes proposed by @Midnighter then this pull request can be merged.

cobra/test/test_core/test_core_reaction.py
```diff
@@ -269,6 +269,13 @@ def test_build_from_string(model):
         assert model.metabolites.foo
         assert len(model.metabolites) == m
 
+    with model:
+        config = Configuration()
```
Midnighter: Since you anyway have to fix the test. I'd write it slightly differently.

1. Make `config` global because it is a global singleton. (Bad for testing but this is the design choice.)
2. Then test with different bounds.
```py
old_bounds = config.bounds
config.bounds = -5, 5
pgi.build_reaction_from_string("g6p_c = f6p_c + new", verbose=False)
assert pgi.bounds == (-5, 5)
config.bounds = old_bounds
```

Midnighter:
Thanks a lot for your contribution @MaxGreil :smiley:

PR #873, opened 2019-07-11
Gap-filling multi-strain workflow

yseif:
This function is a modified version of the existing gapfilling algorithm adapted to the multi-strain workflow.
In such workflows, strain-specific models are built through a reductionist approach with comparative genomics identifying absent metabolic genes and metabolic modeling identifying the corresponding absent metabolic reactions and pathways. A common step in this workflow is a resulting non-functional model which needs to be gapfilled. As such, this function allows the user to pass the list of absent genes and the base model, and returns the reactions to be gapfilled.

### DIFF
--- a/cobra/flux_analysis/gapfill_multi.py
+++ b/cobra/flux_analysis/gapfill_multi.py
@@ -1,0 +1,65 @@
+def gapfill_multi(model, missing_genes, **kwargs):
+    
+    """
+    
+    Generate a list of gapfilling reactions from a list of missing genes for a strain-specific model.
+    
+    :param model: COBRA model for the base strain with the objective coefficient for the reaction of interest (e.g. biomass reaction) set to 1.
+    
+    :param missing_genes: list of genes with no homologs in the strain of interest.
+    
+    :param lower_bound: minimum allowable yield of gapfilled model.
+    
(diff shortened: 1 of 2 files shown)


gregmedlock:
Not sure how others think here, but I think the application is a bit too narrow given how straightforward the process is to warrant a new function to maintain. There are certainly more than a couple users I am aware of that are working on comparative/phylogenetic modeling analyses along the lines you're proposing, but the workflows used to generate reconstructions across those studies aren't very uniform. The functions we add would need to be pretty comprehensive to help anyone, and we'd be helping relatively few people in that case.

What do others think?

Midnighter:
@yseif first of all thank you for the contribution!

I agree with Greg that this is another perfect example for a "cookbook" kind of contribution. Just that we don't have one yet. If there is a need for certain recipes in the community, then we could integrate it in the wiki or the documentation. @yseif I am thinking of something like the [scipy cookbook](https://scipy-cookbook.readthedocs.io/). If you can see yourself starting something like that, I'm sure we find many more things to add.

On a technical note: Your PR seems to contain the same function twice? Also, please note that we use the [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html) for docstrings.

PR #876, opened 2019-07-16
Pinning dependency to fix wheel issues

matthiaskoenig:
This is fixing the SBML issues for now. 
Main issue is that there is a delay between python releases and the availability of wheels.
Pinning the dependencies fixes this. I will update the version when wheels are available.


cdiener:
Is there a difference between `python-libsbml 5.18.0` and `python-libsbml-experimental 5.18.0`. Otherwise we could just depend on `python-libsbml` which always seems to have wheels available...

Issue #880, opened 2019-07-23
Inconsistent bounds in core.Reaction constructor


matthiaskoenig:
The reaction constructor looks as follows:
```
class Reaction(Object):

    def __init__(self, id=None, name='', subsystem='', lower_bound=0.0,
                 upper_bound=None):
       ...        

        # from cameo ...
        self._lower_bound = lower_bound if lower_bound is not None else \
            config.lower_bound
        self._upper_bound = upper_bound if upper_bound is not None else \
            config.upper_bound
```
Because of `lower_bound=0.0` and `upper_bound=None` the bounds are set by default to 
`0.0, config.upper_bound`.
This is probably not what was intended. The constructor should probably be `lower_bound=None, upper_bound=None`.

Not sure what the intended behavior was, but most likely: set default bounds in config if no bounds are provided, not set `0.0, default upper bound`.

matthiaskoenig:
This basically adds new reactions as irreversible when not specifying bounds (which is probably not what users expect when adding a reaction to a model). A lot of tests rely on this assumption, i.e. changing `0.0` to `None` fails many things.
@Midnighter @cdiener could you comment on this?

Midnighter:
I don't know about the original reasoning but irreversible has been the default for quite some time.

Edit: I just checked in the git history and this has been the case since first commits by Daniel Hyduke in 2012.

synchon:
@matthiaskoenig As far as I remember, we did this change in #758.

> This basically adds new reactions as irreversible when not specifying bounds 

I think that is exactly what we thought of. So, if you set `lower_bound` as `None` when initiating, that would make it reversible from the start.

I don't think we documented this caveat after merging it.

matthiaskoenig:
It would be great to add 1-2 sentences of documentation to the Reaction class about this behavior along the lines of 
> Reactions are by default irreversible with bounds `(0.0, config.upper_bound)` if no bounds are provided on creation. To create an irreversible reaction use `lower_bound=None`, resulting in reaction bounds of `(config.lower_bound, config.upper_bound)`.

Issue #879, opened 2019-07-22
Silent drop of complex fbc:geneProductAssociation logic on SBML export


bdelepine:
#### Problem description

Hello guys,

I was playing around with iJO1366 and COBRApy import/export features when I noticed that GPR having a mix of AND and OR logic are silently dropped on SBML export. I would have expected the logic to be exported without trouble as it is a valid SBML+fbc format.

HTH, best,

#### Code Sample

Taking [ACGAptspp as an example, you can see on BiGG](http://bigg.ucsd.edu/models/iJO1366/reactions/ACGAptspp) that its GPR is `(b2417 and b1101 and b2415 and b2416) or (b0679 and b2415 and b2416)`.

Now, if we import ACGAptspp from the iJO1366 shipped with COBRApy:

```python
import cobra
import cobra.test

model = cobra.test.create_test_model("ecoli")  # iJO1366
cobra.io.write_sbml_model(model, "cobra_iJO1366.xml")

print(model.reactions.ACGAptspp.gene_reaction_rule)
```

#### Actual Output

In Python, we do have `'( b2417 and b1101 and b2415 and b2416 ) or ( b0679 and b2415 and b2416 )'`.

In the SBML ("cobra_iJO1366.xml"), however:

```xml
      <reaction id="R_ACGAptspp" name="N-Acetyl-D-glucosamine transport via PEP:Pyr PTS  (periplasm)" reversible="false" fast="false" fbc:lowerFluxBound="cobra_0_bound" fbc:upperFluxBound="cobra_default_ub">
        <listOfReactants>
          <speciesReference species="M_pep_c" stoichiometry="1" constant="true"/>
          <speciesReference species="M_acgam_p" stoichiometry="1" constant="true"/>
        </listOfReactants>
        <listOfProducts>
          <speciesReference species="M_pyr_c" stoichiometry="1" constant="true"/>
          <speciesReference species="M_acgam6p_c" stoichiometry="1" constant="true"/>
        </listOfProducts>
      </reaction>
```

#### Expected Output

I would have expected something along:

```xml
      <reaction id="R_ACGAptspp" name="N-Acetyl-D-glucosamine transport via PEP:Pyr PTS  (periplasm)" reversible="false" fast="false" fbc:lowerFluxBound="cobra_0_bound" fbc:upperFluxBound="cobra_default_ub">
        <listOfReactants>
          <speciesReference species="M_pep_c" stoichiometry="1" constant="true"/>
[...]

Issue #878, opened 2019-07-19
Identifiers.org new identifiers pattern not working


matthiaskoenig:
Due to recent changes in identifiers.org identifiers pattern, the new pattern annotations will not be read correctly (both forms will occur in models)
I.e. the following must be supported when handling annotations

- http://identifiers.org/chebi/CHEBI:17234
- https://identifiers.org/chebi/CHEBI:17234
- https://identifiers.org/CHEBI:17234

For more infomration see:
https://registry.identifiers.org/
https://docs.identifiers.org/articles/api.html#getdataset

Issue #857, opened 2019-06-17
`set_missing_flux_bounds` not respected when reading SBML models


matthiaskoenig:
The SBML parser has a setting to set missing flux bounds to the values specified in the `cobra.configuration.Configuration` object.

In the default setting `set_missing_bounds` no bounds should be injected, but some inf/-inf are written due to incorrect check.

PR #881, opened 2019-07-23
SBML bugfixes

### LINKED ISSUE #850: Error encountered trying to `set model history` when writing SBML
There is a known bug when writing the model history. This results in an error in the log 
```
Error encountered trying to `set model history` 
LibSBML error code -5: The object passed as an argument to the method is not of a type that is valid for the operation or kind of object involved. For example, handing an invalidly-constructed ASTNode to a method expecting an ASTNode will result in this error.
```
This does not affect the model in any way, but is only about writing provenance which is currently only experimental. This error can be savely ignored.

This depends on #810, i.e., how and what meta information to store on the model. As long as there is no decision on #810 it is not clear what the correct behavior is here, so just giving an error for now.

matthiaskoenig:
Another round of bugfixes. One major issue with GPR writing which should go into a release asap (added much better tests for the GPR and proper error reporting if GPRs).
This pull request fixes #850, #879, #878, #857


cobra/io/sbml.py
```diff
@@ -1288,7 +1289,11 @@ def _sbase_notes_dict(sbase, notes):
 
 In the current stage the new annotation format is not completely supported yet.
 """
-URL_IDENTIFIERS_PATTERN = re.compile(r"^https?://identifiers.org/(.+?)/(.+)")  # noqa: E501
+
+URL_IDENTIFIERS_PATTERN1 = re.compile(r"^https?://identifiers.org/(.+?)/(.+)")  # noqa: E501
```
Midnighter: Might be wrapped into one pattern if that helps?

```suggestion
URL_IDENTIFIERS_PATTERN1 = re.compile(r"^https?://identifiers.org/(.+?)[:/](.+)")  # noqa: E501
```

cobra/io/sbml.py
```diff
@@ -1057,14 +1052,20 @@ def _model_to_sbml(cobra_model, f_replace=None, units=True):
             if f_replace and F_GENE_REV in f_replace:
                 gpr = gpr.replace('(', '( ')
                 gpr = gpr.replace(')', ' )')
+                gpr = gpr.replace('  ', ' ')
```
Midnighter: Since the next line anyway splits one a single space I suggest to drop this line and rather replace the next one with splitting on any whitespace:
```py
tokens = gpr.split()
```

cobra/io/sbml.py
```diff
@@ -1057,14 +1052,20 @@ def _model_to_sbml(cobra_model, f_replace=None, units=True):
...
                     if tokens[k] not in [' ', 'and', 'or', '(', ')']:
                         tokens[k] = f_replace[F_GENE_REV](tokens[k])
                 gpr_new = " ".join(tokens)
+            else:
+                gpr_new = gpr
 
             gpa = r_fbc.createGeneProductAssociation()  # noqa: E501 type: libsbml.GeneProductAssociation
-            gpa.setAssociation(gpr_new)
+            # uses ids for GP and does not create GPs
+            _check(gpa.setAssociation(gpr_new, True, False),
```
Midnighter: What do the new arguments change here?

cobra/io/sbml.py
Midnighter: You can also add a `re.I` flag for ignoring case since URLs are not case sensitive.

cobra/test/test_io/test_sbml.py
```diff
@@ -260,15 +261,16 @@ def test_missing_flux_bounds1(data_directory):
         assert r1.upper_bound == config.upper_bound
 
 
+# FIXME
```
Midnighter: What needs to be fixed here?

cobra/io/sbml.py
matthiaskoenig: Did not know about that, changed.

cobra/io/sbml.py
matthiaskoenig: Added better documentation:
```
# uses ids to identify GeneProducts (True),
# does not create GeneProducts (False)
```
The first True decides if `label` or `id` should be used to identify the GeneProduct. To be more specific and avoid problems in case of gene labels I specified to always use the ids. The second flag allows to create the GeneProducts from a GPR if these are not existing in the model. We add all GeneProducts before, so no need to add these.
This could fix some bugs if people have labels on genes, so added the flags while fixing things.

cobra/io/sbml.py
matthiaskoenig: Will make the change. Regular expressions are my weakness, always a trial and error for me.

cobra/test/test_io/test_sbml.py
matthiaskoenig: Was a failing test, just forgot to remove comment after fixing things.

matthiaskoenig:
Proposed changes have been implemented. In addition added the 2 sentences to Fix #880.

cobra/io/sbml.py
```diff
@@ -1385,21 +1382,19 @@ def _parse_annotation_info(uri):
     -------
     (provider, identifier) if resolvable, None otherwise
     """
-    match = URL_IDENTIFIERS_PATTERN1.match(uri)
+    match = URL_IDENTIFIERS_PATTERN.match(uri)
     if match:
         provider, identifier = match.group(1), match.group(2)
+        print(provider, identifier)
```
Midnighter: Please remove the print statement here.

cobra/core/reaction.py
```diff
@@ -46,6 +46,13 @@ class Reaction(Object):
     """Reaction is a class for holding information regarding
     a biochemical reaction in a cobra.Model object.
 
+    Reactions are by default irreversible with bounds
+    (0.0, config.upper_bound)
```
Midnighter: I would replace `config` with `cobra.Configuration()`.

matthiaskoenig:
Remaining changes have been implemented.

Midnighter:
Thank you :blush:

PR #882, opened 2019-07-24
refactor: abstract summary

Midnighter:
I have made a number of refactorings to your branch @synchon. Starting with commit https://github.com/opencobra/cobrapy/commit/c549df7b8643042e0285965a7d1f0b9518548c5b, I did some heavy refactoring to the summary itself. So far only for the metabolite summary. Please take a look at the changes and tell me what you think. Still need to work on reaction and model summary, as well as how exactly information is displayed. I don't think we need to exactly reproduce the old output but some of it certainly makes sense.

### DIFF
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -1099,17 +1099,17 @@ class Reaction(Object):
 
         Returns
         -------
-        cobra.core.summary.ReactionSummary
+        cobra.ReactionSummary
 
         See Also
         --------
-        cobra.core.metabolite.Metabolite.summary
-        cobra.core.model.Model.summary
+        Metabolite.summary
+        Model.summary
 
         """
         from cobra.core.summary import ReactionSummary
 
-        return ReactionSummary(model=self._model, rxn=self, names=names)
+        return ReactionSummary(model=self._model, reaction=self, names=names)
 
     def __str__(self):
         return "{id}: {stoichiometry}".format(
--- a/cobra/core/reaction.py
+++ b/cobra/core/reaction.py
@@ -1087,15 +1087,38 @@ class Reaction(Object):
                     met = Metabolite(met_id)
                 self.add_metabolites({met: num})
 
-    def summary(self, names=False):
+    def summary(
+        self,
+        solution=None,
+        threshold=0.01,
+        fva=None,
+        names=False,
+        float_format="{:.3g}".format
+    ):
         """
         Create a summary of the producing and consuming fluxes of the reaction.
 
         Parameters
         ----------
+        solution : cobra.Solution, optional
+            A previous model solution to use for generating the summary. If
+            None, the summary method will resolve the model.  Note that the
+            solution object must match the model, i.e., changes to the model
+            such as changed bounds, added or removed reactions are not taken
+            into account by this method (default None).
+        threshold : float, optional
+            Threshold below which fluxes are not reported. May not be smaller
+            than the model tolerance (default 0.01).
+        fva : pandas.DataFrame or float, optional
+            Whether or not to include flux variability analysis in the output.
+            If given, fva should either be a previous FVA solution matching the
+            model or a float between 0 and 1 representing the fraction of the
+            optimum objective to be searched (default None).
         names : bool, optional
             Emit reaction and metabolite names rather than identifiers (default
             False).
+        float_format : callable, optional
+            Format string for floats (default ``'{:3G}'.format``).
 
         Returns
         -------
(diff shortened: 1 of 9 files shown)


cobra/core/reaction.py
```diff
@@ -1087,29 +1087,60 @@ def build_reaction_from_string(self, reaction_str, verbose=True,
                     met = Metabolite(met_id)
                 self.add_metabolites({met: num})
 
-    def summary(self, names=False):
+    def summary(
```
synchon: Do you intend to add some functionality which might require the other parameters to be passed?

cobra/core/summary/metabolite_summary.py
```diff
@@ -55,106 +76,50 @@ def _generate(self):
...
-                 rxn) for rxn in rxns]
-
-        flux_summary = pd.DataFrame.from_records(
+        rxns = sorted(self.metabolite.reactions, key=attrgetter('id'))
+        data = [
+            (r.id, self.solution[r.id] * r.metabolites[self.metabolite],
+             r.metabolites[self.metabolite])
+            for r in rxns
+        ]
+        flux_summary = DataFrame(
```
synchon: I think it's better to use `DataFrame.from_records()` since the `data` is a list of tuples.

cobra/core/summary/metabolite_summary.py
```diff
@@ -55,106 +76,50 @@ def _generate(self):
...
             data=data,
-            index=[rxn.id for rxn in rxns],
```
synchon: You could possibly declare the `index` here since you anyway set the `reaction` as index later. Also, then you could remove the `r.id` from `data`.

synchon:
The `Summary` looks really good and clean. I just have a couple of comments for the `MetaboliteSummary`. Also, it would be good to have some more comments in subclass logic, so that it's easier to maintain in the future.

matthiaskoenig:
Just wanted to have a look at the MetaboliteSummary, but getting the following errors:

```
import cobra
model = read_sbml_model("models/e_coli_core.xml")
# print(model.summary())
# for r in model.reactions:
#    r.summary()
for m in model.metabolites:
    m.summary()
```

```
AttributeError                            Traceback (most recent call last)
<ipython-input-6-125b81ebc167> in <module>
      6 #    r.summary()
      7 for m in model.metabolites:
----> 8     m.summary()

~/git/cobrapy/cobra/core/metabolite.py in summary(self, solution, threshold, fva, names, float_format)
    264             fva=fva,
    265             names=names,
--> 266             float_format=float_format
    267         )
    268 

~/git/cobrapy/cobra/core/summary/metabolite_summary.py in __init__(self, metabolite, model, **kwargs)
     53 
     54         """
---> 55         super(MetaboliteSummary, self).__init__(model=model, **kwargs)
     56         self.metabolite = metabolite
     57         self._generate()

~/git/cobrapy/cobra/core/summary/summary.py in __init__(self, model, solution, threshold, fva, names, float_format, **kwargs)
     86         self.float_format = float_format
     87         self.data_frame = None
---> 88         self._generate()
     89 
     90     def __repr__(self):

~/git/cobrapy/cobra/core/summary/metabolite_summary.py in _generate(self)
     77 
     78         """
---> 79         rxns = sorted(self.metabolite.reactions, key=attrgetter('id'))
     80         data = [
     81             (r.id, self.solution[r.id] * r.metabolites[self.metabolite],

AttributeError: 'MetaboliteSummary' object has no attribute 'metabolite'
```

Midnighter:
My bad, this is in progress work. I basically wanted your opinion if I should take it further or just abandon everything including and after commit https://github.com/opencobra/cobrapy/commit/c549df7b8643042e0285965a7d1f0b9518548c5b.

matthiaskoenig:
For me this looks good. Providing a more comprehensive information definitely makes sense. I don't think we have to be backwards compatible in the summary, so making it as useful/informative as possible should be the priority.

cdiener:
Sorry, didn't get what's the major difference between this PR and the other one...

Midnighter:
In order to move ahead with this, I will close this PR and make a new one that includes everything here before https://github.com/opencobra/cobrapy/commit/c549df7b8643042e0285965a7d1f0b9518548c5b. At some point I can make another PR that does this refactoring. The benefit of attempting to do a further abstraction is that the summary would maintain an internal data frame structure which seems the easiest to deal with. It would then only create different representations (text, HTML) from that table.

cobra/core/reaction.py
Midnighter: Not really. It's mostly for interface consistency.

PR #885, opened 2019-07-30
Some improvements to SBML reading

akaviaLab:
These changes improve the reading of notes field, and processing gid working with RECON 2.2

### DIFF
--- a/cobra/__init__.py
+++ b/cobra/__init__.py
@@ -41,4 +41,4 @@ from cobra import io
 from cobra import sampling
 from cobra.util import show_versions
 
-__version__ = "0.15.4"
+__version__ = "0.16.0"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.15.4
+current_version = 0.16.0
 commit = True
 tag = True
 parse = (?P<major>\d+)
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ except IOError:
 if __name__ == "__main__":
     setup(
         name="cobra",
-        version="0.15.4",
+        version="0.16.0",
         packages=find_packages(),
         setup_requires=setup_requirements,
         install_requires=[


matthiaskoenig:
@akaviaLab Your test is breaking some of the GPR rules. Make sure all the test pass.
- Can you add a test/example for the notes which are not parsed and which you try to support. Just a minimal model with a standard note and your new note.
- Can you add a test/example for the GPR which is not supported right now and you try to fix

cobra/io/sbml.py
```diff
@@ -770,7 +772,7 @@ def process_association(ass):
     else:
         # parse deprecated subsystems on reactions
         groups_dict = {}
-        for cobra_reaction in cobra_model.reactions:
+        for cobra_reaction in cobra_model.reactions: # TODO - reaction subsystems are empty, so perhaps they should be filled in or reactions should have groups
```
matthiaskoenig: The SUBSYSTEMS are parsed in groups. Not sure what the TODO means here. Could you explain?

matthiaskoenig:
I opened issues for the things this pull request tries to address.
See 
#887 and #888 
Like mentioned above #887 requires a more comprehensive solution, #888 can be addressed here, but we need an example and tests for this.

cobra/io/sbml.py
```diff
@@ -697,7 +699,7 @@ def process_association(ass):
                     LOGGER.warning("Encoding OBJECTIVE_COEFFICIENT in "
                                    "KineticLaw is discouraged, "
                                    "use fbc:fluxObjective "
-                                   "instead: %s", cobra_reaction)
+                                   "instead: %s", reaction)
```
akaviaLab: This is necessary, because the loop is on reaction, so if it is not corrected, the output stays with the last reaction in a different loop.

cobra/io/sbml.py
akaviaLab: Each reaction has the field subsystem.
After adding subsystems to groups, the ".subsystem" field in each reaction remains empty. I think the subsystem field should link to the groups and/or each reaction should have a ".groups" field.

akaviaLab:
RECON 2.2 is an example (https://www.ebi.ac.uk/biomodels/MODEL1603150001), and I've loaded a small version of it. However, this can be fixed in other issues (@matthiaskoenig said he would fix the notes), so this pull request could be closed if everything is fixed in different pull requests.
[exampleModel.xml.gz](https://github.com/opencobra/cobrapy/files/3458792/exampleModel.xml.gz)

cobra/io/sbml.py
Midnighter: There is some work in progress on adding compatibility for SBML groups indeed.

cobra/io/sbml.py
```diff
@@ -1241,7 +1243,7 @@ def _parse_notes_dict(sbase):
     """
     notes = sbase.getNotesString()
     if notes and len(notes) > 0:
-        pattern = r"<p>\s*(\w+\s*\w*)\s*:\s*([\w|\s]+)<"
+        pattern = r"p>\s*(\w+\s*\w*)\s*:\s*(.*?)</[^<>p]+p>"
```
Midnighter: Can you please comment what this regex is supposed to match? I think making it non-greedy (`(.*?)`) is the right choice but the `</[^<>p]+p>` is unclear to me. You want to match until a non-paragraph element? The original regex also kinda smells so maybe we find a better solution? Also, this should probably be a compiled regex and match case-insensitively since HTML has no case requirement.

Midnighter: I see now the relation to https://github.com/opencobra/cobrapy/issues/888 sorry.

PR #890, opened 2019-08-03
Abstract summary

Midnighter:
Okay, re-opening these commits which mainly modify the docstrings and some of the argument passing of the other PR.


Midnighter:
If you agree with this part @synchon, I will merge this into your branch and then merge that. In a separate PR I can then change the summary further and address other concerns.

PR #891, opened 2019-08-04
chore: remove sbml extras

Midnighter:
Since libsbml (experimental) is now a permanent dependency, we don't need the sbml extras any longer.


Midnighter:
Went ahead with this in order to make a release.

Issue #893, opened 2019-08-07
ImportError: cannot import name 'YAML'


LordVoldemort28:
#### Problem description

Python Version: 3.6.8
CobraPy Version: 0.15.4
Ruamel.yaml Version: 0.16.0

When attempting to import `cobrapy` I get the actual output below. A quick fix for this would be to simply

```python3
from ruamel.yaml.main import YAML
```

instead

```python3
from ruamel.yaml import YAML
```

within `yaml.py` of the `cobra.io` module.

#### Code Sample

Create a [minimal, complete, verifiable example](https://stackoverflow.com/help/mcve).

```python
# Paste your code here.
import cobra
```

#### Actual Output

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/cobra/__init__.py", line 15, in <module>
    from cobra import io
  File "/usr/lib/python3.6/site-packages/cobra/io/__init__.py", line 10, in <modu
le>
    from cobra.io.yaml import from_yaml, load_yaml_model, save_yaml_model, to_yam
l
  File "/usr/lib/python3.6/site-packages/cobra/io/yaml.py", line 7, in <module>
    from ruamel.yaml import YAML
ImportError: cannot import name 'YAML'
```

#### Expected Output

Successful `import`

#### Dependency Information

Please paste the output of `python -c "from cobra import show_versions;
show_versions()"` in the details block below.

<details>

Above code fails due to import error.

</details>

Midnighter:
Can you post the version of ruamel.yaml, please? `pip freeze | grep yaml`

LordVoldemort28:
ruamel.yaml==0.16.0

Midnighter:
Via #894

PR #894, opened 2019-08-07
Fix import of YAML for ruamel

LordVoldemort28:
In reference to issue #893

### DIFF
--- a/cobra/io/yaml.py
+++ b/cobra/io/yaml.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 
 import io
 
-from ruamel.yaml import YAML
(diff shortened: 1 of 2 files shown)


cobra/io/yaml.py
```diff
@@ -4,7 +4,7 @@
 
 import io
 
-from ruamel.yaml import YAML
+from ruamel.yaml.main import YAML
```
cdiener: If you exchange that line with the following one the CI will pass (imports need to be sorted alphabetically).

Midnighter:
Thanks for the fix @LordVoldemort28. I took the liberty to push some more changes to your branch because I want to try making a release tonight.

PR #895, opened 2019-08-14
Chain the original SBML exception

kvikshaug:
Hi all, in our application we saw cobrapy logging an error with message "None", in addition to a stacktrace for a _handled_ exception being printed to console. The cause is that `traceback.print_exc()` is assumed to return a string, but it only _prints_ an exception to console and returns nothing.

In this fix, I suggest to use [exception chaining](https://legacy.python.org/dev/peps/pep-3134/) instead of logging. This gives the calling context access to everything needed to investigate the root cause if wanted, and decide whether or not to log events and/or print anything to console.

### DIFF
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -39,7 +39,7 @@ from copy import deepcopy
 from sys import platform
 
 import libsbml
-from six import iteritems, string_types
+from six import iteritems, raise_from, string_types
 
 import cobra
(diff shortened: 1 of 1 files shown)


matthiaskoenig:
Hi @kvikshaug, thanks for this. I am back from holidays now and will fix the open SBML issues this week. As part of this I would include your proposed changes.

Issue #887, opened 2019-07-31
GPR of the form (A and B) or (A and C) are currently not supported


matthiaskoenig:
Some reactions have (A and B) or (A and C), and removing the parenthesis breaks adding the gene rules, during GPR parsing.
This is due to some processing necessary for id replacements during writing.

@akaviaLab Could you provide an example model for this and add further description if I did not catch the full extent of the issue?

gregmedlock:
I just happened across this issue with a model today (paper here: https://www.nature.com/articles/ncomms14250#supplementary-information but SBML only available here in "Ratcon files": http://bme.virginia.edu/csbl/Downloads1.html). In the GPRs with parentheses I'm coming across, the reactions within the cobrapy model end up with nothing in the GPR.

I can take a crack at it if you are working on other issues @matthiaskoenig. I need to fix it before moving forward with this work so can probably complete very soon.

akaviaLab:
Hi
One example is RECON 2.2 (https://www.ncbi.nlm.nih.gov/pubmed/27358602), available on https://www.ebi.ac.uk/biomodels/MODEL1603150001
RECON 2.2 is valid SBML according to the online validator. Can you explain what purpose does the parenthesis removal serve? If this is fixed, it would help to figure out what the purpose of the original function was, in order not to remove functionality.
I'm working on truncated versions of RECON 2.2 that could be useful as quick test cases

matthiaskoenig:
@gregmedlock and @akaviaLab Thanks for the examples. I can fix this issue tomorrow.

akaviaLab:
This has not been fixed. I'm still not clear why removing the parenthesis is necessary at all.
For example, if the GPR is (A and B), it should work fine with parenthesis. Can you please explain why removing the parenthesis is necessary?

akaviaLab:
Sorry, I was looking at an older version of the code. It seems to process reactions correctly. I think it can be closed.

gregmedlock:
Issues with iHsa GPRs are resolved--thanks!

Midnighter:
:tada:

Issue #883, opened 2019-07-25
error writing to sbml without prefixing identifiers


Karrenbelt:
Trying to prevent cobra from prefixing my identifiers when writing SBML, I got this (using the textbook model, model history error is unrelated):

`cobra.io.write_sbml_model(model, 'test.xml', f_replace={})`

```
Error encountered trying to <set model history>.
LibSBML error code -5: The object passed as an argument to the method is not of a type that is valid for the operation or kind of object involved. For example, handing an invalidly-constructed ASTNode to a method expecting an ASTNode will result in this error.
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Users/Zarathustra/ETH/repositories/sbml_tools/venv/lib/python3.6/site-packages/cobra/io/sbml.py", line 831, in write_sbml_model
    doc = _model_to_sbml(cobra_model, f_replace=f_replace, **kwargs)
  File "/Users/Zarathustra/ETH/repositories/sbml_tools/venv/lib/python3.6/site-packages/cobra/io/sbml.py", line 1059, in _model_to_sbml
    gpa.setAssociation(gpr_new)
UnboundLocalError: local variable 'gpr_new' referenced before assignment
```

replacing gpr_new with gpr in these [lines](https://github.com/opencobra/cobrapy/blob/master/cobra/io/sbml.py#L1064-L1067 ) should fix this

matthiaskoenig:
I basically fixed this yesterday in the `devel` branch as part of other bugfixes ;). Please try the devel version. 
Do in your virtualenv
```
git clone https://github.com/opencobra/cobrapy.git
git checkout devel
pip install -e .
```
This will be resolved with the next release.

Karrenbelt:
it writes, but reading back it should to be erroneous, with libsbml shows errors

`cobra.io.read_sbml_model('test.xml')`

```
Traceback (most recent call last):
  File "/Users/Zarathustra/ETH/repositories/sbml_tools/cobrapy/cobra/io/sbml.py", line 210, in read_sbml_model
    **kwargs)
  File "/Users/Zarathustra/ETH/repositories/sbml_tools/cobrapy/cobra/io/sbml.py", line 381, in _sbml_to_model
    sid = _check_required(specie, specie.getIdAttribute(), "id")
  File "/Users/Zarathustra/ETH/repositories/sbml_tools/cobrapy/cobra/io/sbml.py", line 1194, in _check_required
    raise CobraSBMLError(msg)
[...]

PR #901, opened 2019-08-21
Sbml improvements - notes, gene IDs, group IDs, and test

akaviaLab:
As requested in #885, the commits are atomic.
This pull request covers multiple SBML improvements
1) Removes parenthesis, unless their removal results in a parsing error #887 
2) Parses SBML notes of various formats #888 
3) Converts gene, reaction, metabolite ID back and forth, including dealing with non-ascii characters that are not allowed in SBML IDs, should tackle #883 and has some progress towards https://github.com/opencobra/schema/issues/6 and https://github.com/opencobra/schema/issues/8
4) Added conversion of group IDs. Currently this uses the prefix G_ but it can be changed. One option is P_ for grouP
5) Added test of notes and GPR


cobra/test/test_io/test_sbml.py
```diff
@@ -388,3 +388,62 @@ def test_identifiers_annotation():
...
+                            '41-21)6-39-44(36,37)42-43(34,35)38-5-10-13(29)15(31)20(40-10)27-3-1-2-9(4-27)18('
+                            '23)33/h1,3-4,7-8,10-11,13-16,20-21,29-32H,2,5-6H2,(H2,23,33)(H,34,35)(H,36,37)(H2,22,24,'
+                            '25)/p-2/t10-,11-,13-,14-,15-,16-,20-,21-/m1/s1',
+                   'chebi': 'CHEBI:57945'}
+    }
+    rxnNotesValue = {'CONFIDENCE_LEVEL': '4', 'NOTES': 'NCD', 'SUBSYSTEM': 'Propanoate metabolism',
+                     'GENE_ASSOCIATION': '(HGNC:8546 and HGNC:8548) or (HGNC:8547 and HGNC:8548)'}
+    rxnAnnotValue = {'sbo': 'SBO:0000176', 'ec-code': '1.1.1.27', 'pubmed': ['10108', '21765']}
+
+    for metID in metNoteValues.keys():
```
Midnighter: You don't need the `.keys()` method. A `for` loop automatically iterates over the keys in a dictionary.

```suggestion
    for met_id in metNoteValues:
```

Midnighter:
If you like I can fix these last little issues. I feel bad making you work so much on style problems when we should be using a tool like [black](https://pypi.org/project/black/).

akaviaLab:
Current status - I can't figure out why 3.5 is complaining on import order. Suggestions would be appreciated. Can we just ignore this? You can also remove the kludge fix (6ec605b0dc27f38b73b932a962c010a0845fc0e3), since it didn't fix 3.5 complaining.

2.7 fails. I can't manage to run 2.7 tests on my system. I'm getting an error importing sympy.

If you can fix 3.5 and 2.7, I would appreciate it.

cdiener:
I think 3.5 is complaining about import order since isort is only run there (doesn't make sense to re-run style checks on all versions).

Midnighter:
Looks really good now. Thank you for the fixes and your persistence!

akaviaLab:
Question - how do these functions relate to cobra.manipulation.escape_ID(model)?
Should I have modified escape_ID instead of modifying it in place? Is escape_ID obselete and shoudl be erased?

PR #908, opened 2019-09-07
WIP: Add a code of conduct

Midnighter:
@opencobra/cobrapy-core we need to decide on a contact e-mail address where concerns can be reported.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,0 +1,139 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
+or harmful.
+
+Community leaders have the right and responsibility to remove, edit, or reject
+comments, commits, code, wiki edits, issues, and other contributions that are
+not aligned to this Code of Conduct, and will communicate reasons for moderation
+decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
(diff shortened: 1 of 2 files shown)


cdiener:
I would suggest a committee of 3 people.

synchon:
Or maybe a separate core group e-mail address?

Midnighter:
Yes, I was also thinking a single person is not a good idea. Another problematic situation is that our core team at the moment consists entirely of men. Both for the development core team, as well as for a committee dealing with misconduct more diversity would be a great improvement. I was thinking earlier that we could, at least on the development side, offer closer mentorship for one or two devs from minority groups?

Actually, I should probably bounce this into the mailing list. Especially people in the community who are maybe not completely comfortable making code changes but still want to contribute could very well help with the committee.

synchon:
>  I was thinking earlier that we could, at least on the development side, offer closer mentorship for one or two devs from minority groups?

Definitely a good idea.

cdiener:
I would have suggested @sgalkina. I agree that we should have more diverse contributor/maintainer populations.

gregmedlock:
Myself and @maureencarey can offer to serve on the committee and as points of contact. GLM5UH@virginia.edu and MAC9JC@virginia.edu are, respectively, good contact addresses. We will both be in academia working in the COBRA world long-term so are happy to serve as contacts for the foreseeable future.

Midnighter:
Excellent, I'm glad to see some movement on this. Ideally, there would be one contact email address and we can set up forwarding to whoever is on the committee at the time.

gregmedlock:
I can create a UVA-based listserv that would forward to all committee members. Messages would be archived and private (unless someone filed a Freedom of Information Act [FOIA] request, which all government-affiliated organizations in the USA have to comply with). How does cobrapy_conduct_committee@virginia.edu sound?

Midnighter:
I would love to have an opencobra.org domain but since that's a long shot, setting up such a list sounds good to me. Can we have hyphens instead of underscores, though? :slightly_smiling_face:

matthiaskoenig:
@Midnighter Don't think this is a longshot. I just got the domain `opencobra.org`. Should be activated in a few hours. We can then setup an email service for that such as zoho and should be good to go in a few days.

cdiener:
Either way we should also allow contacting only a subset of the committee in case the report involves any of the sitting members. Everything else would expose the identity of the victim to a potential perpetrator. The hybrid solutions from pytest makes sense https://github.com/pytest-dev/pytest/blob/master/CODE_OF_CONDUCT.md to me.

Midnighter:
> @Midnighter Don't think this is a longshot. I just got the domain `opencobra.org`. Should be activated in a few hours. We can then setup an email service for that such as zoho and should be good to go in a few days.

That is the pro-active way :slightly_smiling_face: I said long shot because opencobra is more than just cobrapy, it also involves the Matlab and Julia community. Of course, it'd be great to have one code of conduct committee for the whole organisation but the tricky part is that there _is_ no formal organisation. So maybe what we should do here is take the lead but ask other communities if they want in on it?

@cdiener that makes a lot of sense to me :+1:

matthiaskoenig:
Everything activated. For now I just forwarded http://opencobra.org to http://opencobra.github.io/. 
I setup an email address cobrapy@opencobra.org for now which works from now on (just mail me privately that I can share the password and access for imap/smtp privately with the core team). Something like contact@opencobra.org could work for a bigger comittee in the future.

I used migadu (https://www.migadu.com/en/index.html) for now which limits the outgoing emails in the free plan, so don't go wild with sending mails from the account for now.

Best Matthias

Midnighter:
Sounds great. I think either conduct@opencobra.org or coc@opencobra.org would be nice addresses to use for this purpose.

matthiaskoenig:
@Midnighter The great part about migadu is that there is no limitation on addresses. So we can just decide on what is the proper address and who is on the board so that they get access to the email. Just let me know when there is consensus then this can be done in 5 minutes. Just tested if all of this works with the `opencobra.org` domain and the respective emails. Looks basically very simple. There would be an email `conduct@opencobra.org` which would be delegated to multiple other addresses (the code of conduct team, e.g. beber@opencobra.org, diener@opencobra.org, ...). These users would get all the `conduct@opencobra.org` information. The code of conduct team would then just add their respective email to their email program. When the team changes we just add/remove addresses from the delegation (works like a team email). There is full SMTP/IMAP integration so no need for web frontends, but just add the address to thunderbird or any other email client.

Midnighter:
@sgalkina are you still up for being a committee member, too, or not so much any longer? Let me know and I will update this PR with the relevant addresses.

Midnighter:
Svetlana has confirmed her position with me. @matthiaskoenig can you set up conduct@opencobra.org to forward to the included virginia.edu and biosustain.dtu.dk addresses, please?

Midnighter:
@maureencarey @sgalkina @gregmedlock many thanks for stepping up and taking on this responsibility. I hope it will be uneventful :wink:

Issue #865, opened 2019-06-27
Error in loopless FVA


Midnighter:
#### Problem description

It seems like the loopless FVA code is not robust against infeasible solutions when it comes to the Gurobi solver. It does not happen with the GLPK solver although many solutions are infeasible or the status is undefined. CPLEX seems also fine.

#### Code Sample

Download the [yeast-GEM 8.3.3](https://github.com/SysBioChalmers/yeast-GEM/releases/tag/v8.3.3).

```python
import cobra
from cobra.io import read_sbml_model
from cobra.flux_analysis import flux_variability_analysis


config = cobra.Configuration()
config.solver = "gurobi"
model = read_sbml_model("yeastGEM.xml")
fva_result = flux_variability_analysis(model, loopless=True)
```

#### Actual Output

```
cobra/util/solver.py:416 UserWarning: solver status is 'infeasible'

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<timed exec> in <module>

~/.virtualenvs/carrot/lib/python3.6/site-packages/cobra/flux_analysis/variability.py in flux_variability_analysis(model, reaction_list, loopless, fraction_of_optimum, pfba_factor, processes)
    195             else:
    196                 _init_worker(model, loopless, what[:3])
--> 197                 for rxn_id, value in map(_fva_step, reaction_ids):
    198                     fva_result.at[rxn_id, what] = value
    199 

~/.virtualenvs/carrot/lib/python3.6/site-packages/cobra/flux_analysis/variability.py in _fva_step(reaction_id)
     47     sutil.check_solver_status(_model.solver.status)
     48     if _loopless:
---> 49         value = loopless_fva_iter(_model, rxn)
     50     else:
     51         value = _model.solver.objective.value

~/.virtualenvs/carrot/lib/python3.6/site-packages/cobra/flux_analysis/loopless.py in loopless_fva_iter(model, reaction, solution, zero_cutoff)
    221         # If the previous optimum is maintained in the loopless solution it was
    222         # loopless and we are done
--> 223         if abs(reaction.flux - current) < zero_cutoff:
    224             if solution:
    225                 return sol

TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```

#### Expected Output

I expect to receive a valid FVA result.

#### Dependency Information

<details>

System Information
==================
OS                     Linux
OS-release 4.15.0-52-generic
Python                 3.6.7

Package Versions
================
cobra                        0.15.3
[...]

PR #917, opened 2019-10-18
Handle numerical problems gracefully in FVA

cdiener:
Addresses #865.

For some models there may exist an optimum for the reference FVA solution but consecutive optimizations may not find the optimum again due to numerical issues, leading to infeasible solutions in FVA loops. This adds code to handle this gracefully. 

I can't really add a test for that since none of our test models are this unstable.

### DIFF
--- a/cobra/flux_analysis/variability.py
+++ b/cobra/flux_analysis/variability.py
@@ -54,6 +54,7 @@ def _fva_step(reaction_id):
         value = _model.solver.objective.value
     # handle infeasible case
     if value is None:
+        value = float("nan")
         LOGGER.warning(
             "Could not get flux for reaction %s,"
             " setting it to NaN. This is usually due to numerical instability.",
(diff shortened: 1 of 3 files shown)


cobra/flux_analysis/variability.py
```diff
@@ -42,21 +44,35 @@ def _fva_step(reaction_id):
...
+        {rxn.forward_variable: 1, rxn.reverse_variable: -1}
+    )
     _model.slim_optimize()
     sutil.check_solver_status(_model.solver.status)
     if _loopless:
         value = loopless_fva_iter(_model, rxn)
     else:
         value = _model.solver.objective.value
+    # handle infeasible case
+    if value is None:
```
Midnighter: As far as I can tell the value is not set in this condition, so nothing actually changed?

cdiener: Oh yeah, sorry for that. Will fix.

PR #918, opened 2019-10-21
Small fixes to flux sampling

### LINKED ISSUE #903: Documentation possible mistake in Flux Sampling
Hello, everybody,

I have installed the stable version 0.16.0 of Cobra. For do the sampling, the documentation specifies that it is necessary to import _sampling_ from _cobra.flux_analysis_. However, this module does not exist and, instead, it should be imported from _cobra.sampling._

Before
`from cobra.flux_analysis.sampling import OptGPSampler, ACHRSampler
`
Now: 
`from cobra.sampling import OptGPSampler, ACHRSampler
`
Is this correct?

Thanks in advance!

### LINKED ISSUE #911: OptGPSampler raises ValueError when additional constraints are added
#### Problem description

When using a model with additional constraints added, the OptGPSampler raises a ValueError when constructing the center from the warmup points:

In `cobra.sampling.optgp.py` (line 172):
`ValueError: `data` must have the same shapeas the created array.`

This is likely due to the parent class copying and updating the model in preparation for the sampling, causing the number of variables to change.

#### Suggested fix
In `cobra.sampling.optgp.py` (line 172):
replace `len(model.variables)` with `len(self.model.variables)`

#### Code Sample

```python
import cobra
import cobra.test

cobra.test.create_test_model("textbook")

expression = model.reactions.TPI.forward_variable
constraint = model.problem.Constraint(expression=expression, name="test", lb=0, ub=1)
model.add_cons_vars([constraint])

sampler = cobra.sampling.OptGPSampler(model)
sampler.sample(1)
```

<details>

System Information
==================
OS                     Linux
OS-release 4.15.0-64-generic
Python                 3.7.4

Package Versions
================
cobra                       0.16.0
depinfo                      1.5.1
future                      0.17.1
numpy                       1.17.2
optlang                      1.4.4
pandas                      0.25.1
pip                         19.2.3
ruamel.yaml                 0.16.5
setuptools                  41.2.0
six                         1.12.0
wheel                       0.33.6

</details>

cdiener:
This fixes two small issues that have popped up.

1. Optlang creates helper variables for Gurobi constraints. When copying the solver this can create extra dangling variables. Those do not seem to affect the results so the provided workaround should be okay for now. In the long run we should fix the actual issue though. Fixes #911.

2. Documentation did not yet reflect the new import paths for flux sampling. Fixes #903.


cobra/sampling/optgp.py
```diff
@@ -169,16 +169,17 @@ def __init__(self, model, processes=None, thinning=100, nproj=None,
 
         # This maps our saved center into shared memory,
         # meaning they are synchronized across processes
-        self.center = shared_np_array((len(model.variables), ),
-                                      self.warmup.mean(axis=0))
+        self.center = shared_np_array(
```
cdiener: This is the actual change.

cdiener:
Sorry for that. I created a comment that shows the actual change. Basically teh number of variables is not taken from from the copied model and not the one passed to the constructor (since Gurobi will add variables on copy). I will turn off the autoformatting for the next PRs.

PR #925, opened 2020-01-15
[feat] add allow_nan to to_json

achillesrasquinha:
By default, `cobra.io.to_json` dumps json with `allow_nan` as `False`. Allow overriding the same. This behaviour happens to be available within the `save_json_model` function.

### DIFF
--- a/cobra/io/json.py
+++ b/cobra/io/json.py
@@ -42,7 +42,11 @@ def to_json(model, sort=False, **kwargs):
     """
     obj = model_to_dict(model, sort=sort)
     obj[u"version"] = JSON_SPEC
-    return json.dumps(obj, allow_nan=False, **kwargs)
+
+    dump_opts = { "allow_nan": False }
+    dump_opts.update(**kwargs)
+
+    return json.dumps(obj, **kwargs)
 
 
 def from_json(document):


Midnighter:
Hi @achillesrasquinha, thank you for your contribution! At the time, I decided against allowing nan because it is not part of the official JSON specification and is Python specific. Since the JSON format is most often used in web communication, we don't know what kind of tool will handle the data and thus I assumed it safest to stick to the official spec.

Thus I would rather warn against using the feature in `save_json_model` than introduce it here.

achillesrasquinha:
Ok, do you then recommend to have `save_json_model` raise a warning in this case?

Midnighter:
Yes, please insert a condition that raises a warning (`UserWarning`) along the lines of what I described above. Could you also adjust the `to_json` function to have arguments just like `save_json_model` and handle keyword arguments plus the warning there, too, please?

If you feel like going the whole nine yards, you could also adjust the json module tests to ensure that the warning is raised the `allow_nan=True` is passed as a keyword argument.

Midnighter:
Hey @achillesrasquinha, I think this was handled more thoroughly in #957 so I'm closing this PR. Thank you for the suggestion, though.

Issue #928, opened 2020-01-29
minimal_medium IndexError: list index out of range


PkiwiBird:
#### Problem description
I applied FASTCORE algorithm to a recon3D model for different samples. I got sample specific models. Then I put bounds of drains associated with import of medium components from -1000 to 1000, while putting remaining drains from 0 to 1000 in each sample model. Then tested FBA, but models were infeasible.
To find other drains that have to be opened so that models are feasible, I first opened all drains of each sample model to use the function 'minimal_medium': `minimal_medium(OpenedSampleModel, 0.1, minimize_components=True) ` For three of the samples it worked fine and when opening their minimal medium drains they were feasible. For one sample however it gave the error:
```
Traceback (most recent call last):
  File "C:\Users\Tania\.conda\envs\models\lib\site-packages\IPython\core\interactiveshell.py", line 3296, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-56-e34d9e64d654>", line 2, in <module>
    minMed_01 = list(minimal_medium(OpenedSampleModel, 0.1, minimize_components=True).index.values) # most cell lines have a growth rate of 0.1
  File "C:\Users\Tania\.conda\envs\models\lib\site-packages\cobra\medium\minimal_medium.py", line 222, in minimal_medium
    medium = media[0]
IndexError: list index out of range
```
Below is a pickle of `OpenedSampleModel`
Could you please tell what is wrong?
[OpenedSampleModel.zip](https://github.com/opencobra/cobrapy/files/4127685/OpenedSampleModel.zip)

cdiener:
Thanks for providing the model I'll investigate and get back to you.

cdiener:
Hi, looks like your model was pickled using gurobi as the solver which means I can't unpickle it on my machine. Would you mind switching the solver before pickling?

```Python
model.solver = "glpk"
```

Thanks!

tanialbb:
> Hi, looks like your model was pickled using gurobi as the solver which means I can't unpickle it on my machine. Would you mind switching the solver before pickling?
> 
> ```python
> model.solver = "glpk"
> ```
> 
> Thanks! What I can already say that the problem is that the optimization does not converge (can't find an optimal solution)...

[OpenedSampleModel.zip](https://github.com/opencobra/cobrapy/files/4129869/OpenedSampleModel.zip)
goes

cdiener:
Hi,

[...]

PR #929, opened 2020-01-29
Handle repeated infeasible MIP optimization in minimal_medium

cdiener:
Similar to FVA this adds a warning and `None` return for cases where the repeated optimization of the same objective with the same model suddenly leads to a non-optimal solution. This is only the case for numerically unstable models so it is hard to add a test for this case since none of the test models included in cobrapy will ever run into this case.

Related to #928

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,7 @@ install:
   - set PATH=%PYTHON%;%PYTHON%\Scripts;%PYTHON%\Library\bin;%PATH%"
   - "ECHO %PYTHON%"
   - python --version
-  - python -m pip install --upgrade --disable-pip-version-check pip setuptools wheel tox
+  - python -m pip install --upgrade --disable-pip-version-check pip setuptools wheel tox six>=1.14.0
 
 test_script:
   - tox
--- a/cobra/medium/minimal_medium.py
+++ b/cobra/medium/minimal_medium.py
@@ -161,9 +161,8 @@ def minimal_medium(model, min_objective_value=0.1, exports=False,
     minimize the number of "large" import fluxes. Specifically, the detection
     limit is given by ``integrality_tolerance * max_bound`` where ``max_bound``
     is the largest bound on an import reaction. Thus, if you are interested
-    in small import fluxes as well you may have to adjust the integrality
-    tolerance at first with
-    `model.solver.configuration.tolerances.integrality = 1e-7` for instance.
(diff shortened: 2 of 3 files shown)


cdiener:
Good point. Right now this is already mentioned in the documentation for `minimal_medium`. 

```
Notes
-----
Due to numerical issues the `minimize_components` option will usually only
minimize the number of "large" import fluxes. Specifically, the detection
limit is given by ``integrality_tolerance * max_bound`` where ``max_bound``
is the largest bound on an import reaction. Thus, if you are interested
in small import fluxes as well you may have to adjust the integrality
tolerance at first with
`model.solver.configuration.tolerances.integrality = 1e-7` for instance.
However, this will be *very* slow for large models especially with GLPK.
```

Midnighter:
Okay, two follow up points then:

1. How about the warning says something like, "please carefully read the note on numerical instability in the function documentation".
2. In the note itself, do you think it would be a bad recommendation to suggest changing `model.tolerance` (thus potentially also setting feasibility and optimality tolerance) or do you think it would be okay to do so?

cdiener:
Hmmm, AFAICT I did not touch the parts that are failing now (summaries).

Midnighter:
Odd indeed, your changes shouldn't have any effect at all. I'm only with phone until next week. Can't really help until then, sorry!

cdiener:
No worries, I'll investigate.

cdiener:
Okay this is not super clean but I fixed the summary tests here. If we don't squash the commits it will still show up as separate ones... The summary tests relied on a particular maximum column width in pandas output so not sure how that ever worked. Appveyor was due to some issues with tox and pip which require fixing certain package versions.

Midnighter:
That's super weird but I'm glad you found a fix. Please merge when it passes. I'll see if I can update some of the open PRs once this issue should be fixed.

synchon:
> The summary tests relied on a particular maximum column width in pandas output so not sure how that ever worked.

I remember having to add the extra ellipsis for the tests to pass. My initial refactored version didn't include ellipsis and the tests failed. But, I guess the string outputs don't require it anymore (not sure why?!). Thanks for the fix though!

cdiener:
Yeah I also shortened the strings so they don't get cut off in the output. I think we don't need to be too strict with the string output since it is generated from the pandas DataFrame directly. We can just assume that pandas `DataFrame.to_string` works and mostly test for the correct entries in the dataframe version.

Issue #736, opened 2018-07-16
Cannot export SBML when annotation is a list of lists


zakandrewking:
#### Problem description

The JSON format allows annotations as lists of lists, but trying to export such a model as SBML causes an error. 

This is the case for the BiGG Models universal model (http://bigg.ucsd.edu/data_access). If that model is incorrectly formatted, I can fix it, but it might be good to also add some warning in COBRApy.

Might be related to #580. Originally reported as https://github.com/SBRG/bigg_models/issues/299.

(Thanks!!!!)

#### Code Sample

Model:

```json
{
  "metabolites":[
    {
      "id":"4crsol_c",
      "name":"",
      "compartment":"",
      "annotation":[
        [
          "KEGG Compound",
          "http://identifiers.org/kegg.compound/C01468"
        ],
        [
          "CHEBI",
          "http://identifiers.org/chebi/CHEBI:11981"
        ]
      ]
    }
  ],
  "reactions":[],
  "genes":[],
  "id":"test",
  "compartments":{},
  "version":"1"
}
```

Script:

```python
import cobra
m = cobra.io.load_json_model('test.json')
cobra.io.write_sbml_model(m, 'test.xml')
```

#### Actual Output

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-13-a2868fcb5a70> in <module>()
----> 1 cobra.io.write_sbml_model(m, 'test.xml')

/usr/local/lib/python3.6/site-packages/cobra/io/sbml3.py in write_sbml_model(cobra_model, filename, use_fbc_package, **kwargs)
    715         return
    716     # create xml
--> 717     xml = model_to_xml(cobra_model, **kwargs)
    718     write_args = {"encoding": "UTF-8", "xml_declaration": True}
    719     if _with_lxml:

/usr/local/lib/python3.6/site-packages/cobra/io/sbml3.py in model_to_xml(cobra_model, units)
    482                              hasOnlySubstanceUnits="false")
    483         set_attrib(species, "name", met.name)
--> 484         annotate_sbml_from_cobra(species, met)
    485         set_attrib(species, "compartment", met.compartment)
    486         set_attrib(species, "fbc:charge", met.charge)

/usr/local/lib/python3.6/site-packages/cobra/io/sbml3.py in annotate_sbml_from_cobra(sbml_element, cobra_element)
    242     bag = SubElement(SubElement(rdf_desc, ns("bqbiol:is")),
[...]

PR #930, opened 2020-02-01
Check annotation to be of type dict while reading it from the json

Hemant27031999:
## What does it fixes: ##
The current version of code throws an error at the time of writing model in SBML format if annotations are of type list of lists. The JSON schema wants annotations to be of dictionary type, but the validation check of JSON is not performed using the schema because it will make the parsing slow. So I have _updated cobra/io/object.py_ class by adding setter for annotation which checks if the annotation is of type dictionary or not at the time reading model from the JSON and throws a TypeError if it is not.
## Files changed/added: ##
* Updated cobra/io/object.py.
* Added _cobra/test/test_io/test_annotation_format.py_ test which tests the JSON format from the files _cobra/test/data/valid_annotation_format.json_ and _cobra/test/data/invalid_annotation_format.json_ of the annotation. If the annotation is in the right format, parsing is done and SBML of the model is written in a file called _cobra/test/data/valid_annotation_output.xml_.


Hemant27031999:
@cdiener , I have implemented the needs for annotations format that were discussed on [issue:736](https://github.com/opencobra/cobrapy/issues/736). I have also added the test cases to test the parsing. However, a few other tests are now producing error. One of them is: [_cobra/test/test_io/test_sbml.py_](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/test_io/test_sbml.py). It has a _test_read_2_ test which compares models from two files. For eg: it compares models [_mini.pickle_ ](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/data/mini.pickle)and [_mini_fbc2.xml_](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/data/mini_fbc2.xml)  using the function _extra_comparisons()_ of same file. However, while comparing data from two files, the following error occurs : 

>     @classmethod
>     def extra_comparisons(cls, name, model1, model2):
>         assert model1.compartments == model2.compartments
>     
>         # FIXME: problems of duplicate annotations in test data
>         #  ('cas': ['56-65-5', '56-65-5'])
>         # assert dict(model1.metabolites[4].annotation) == dict(
>         #    model2.metabolites[4].annotation)
>         d1 = model1.reactions[4].annotation
>         d2 = model2.reactions[4].annotation
> >       assert list(d1.keys()) == list(d2.keys())
> E       AssertionError: assert [] == ['sbo', 'bigg.reaction']
> E       Right contains 2 more items, first extra item: 'sbo'

Here as you can see, the two annotation lists are different (one is empty and other is ['sbo', 'bigg.reaction']) from the two models and hence a false assert is produced which fails the test. I tried to look into the corresponding models ([_mini.pickle_](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/data/mini.pickle) and [_mini_fbc2.xml_](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/data/mini_fbc2.xml)) but one of them is a pickle file and I am a bit confused how I compare two of them to dig more deeper. Can you help me a little to visualize them?

cdiener:
Sure, I'll have a look during the week.

cobra/core/object.py
```diff
@@ -40,6 +40,19 @@ def id(self, value):
     def _set_id_with_model(self, value):
         self._id = value
 
+    @property
+    def annotation(self):
+        return getattr(self, "_annotation", None)
```
Midnighter: Since the attribute `_annotation` is created on line 23, why do you expect it to not exist here?

```suggestion
        return self._annotation
```

Seems to be sufficient?

Hemant27031999: Oh yaah, sorry about that. I'll change it. Thanks

Hemant27031999:
Thanks Sir for looking into the matter.

The error occurring here is due to the changes I made in Object class for annotation. I haven't repickled the models after modifying the Object class, that's why it is giving the annotation list as an empty list. So I tried to repickle the models using the modified Object class using the [update_pickles.py](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/data/update_pickles.py) file, but I think this file is outdated, and it throws the following error:
* it is using 'write_sbml2' method of 'cobra.io.sbml3' module in line 13, but neither this module nor this method (after checking in cobra.io.sbml) is present anywhere.
* it is using 'iJO1366.xml' model in line 29 but this model is present in compressed form i.e 'iJO1366.xml.gz', so it also throws an error.
* in line 63, 'D__LACt2pp' reaction is used from 'ecoli_model' (this model is made using iJO1366.xml), but model [iJO1366](http://bigg.ucsd.edu/models/iJO1366) doesn't have any such reaction, so it throw an error here also.

Am I missing something or is it really outdated?

Midnighter:
> Am I missing something or is it really outdated?

Yeah, most likely outdated. It only really gets updated when someone needs to run it, like you do now :wink:

Hemant27031999:
Ok. Let me guess, what we will have to change in it is somewhat like this :
* Regarding the first error where 'write_sbml2' method is used to write the model in 'mini_fbc1.xml', is it still required anywhere or should it be removed? I looked for its usage and found one in ['test_sbml.py'](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/test_io/test_sbml.py) file in line 41 i.e the fourth test case 

                  IOTrial('fbc1', 'mini.pickle', 'mini_fbc1.xml',
                  read_sbml_model, write_sbml_model, None),

 But even here also, it is not in any actual use. In the first test i.e 

            def test_validate(trial, data_directory):
                  """ Test validation function. """
                  if trial.validation_function is None:
                  pytest.skip('not implemented')
                  test_file = join(data_directory, trial.test_file)
                  trial.validation_function(test_file)

it gets skipped due to no provided validation_function for it. And similarly for further test cases also, it is either marked as 'pytest.xfail('not supported')' or skipped. So it is not actually used anywhere.
* Regarding the second error i.e using 'iJO1366.xml', I think replacing it with 'iJO1366.xml.gz' will do the required work.
* Regarding the third error i.e using reaction 'D__LACt2pp' from model [iJO1366](http://bigg.ucsd.edu/models/iJO1366), I have some doubts. On the BiGG Model database, I can't find this reaction under model [iJO1366](http://bigg.ucsd.edu/models/iJO1366). So should it be removed here also?
Thanks.

Hemant27031999:
Sir, I have updated update_pickle.py file as per the changes discussed above. About the third error which I mentioned, I have removed the reaction 'D__LACt2pp' from model iJO1366 and have updated the pickle and other resource files using the updated Object class. Changes suggested by @Midnighter have also been done. There was one test case that was failing i.e test_validate inside [test_sbml.py](https://github.com/opencobra/cobrapy/blob/devel/cobra/test/test_io/test_sbml.py) (line 277) file. It was asking for 23 errors when validation is performed over mini_fbc3.xml resource model, but there were no errors generated after I have modified the Object class. 
Please review my changes and let me know if some changes are required. 
Thanks.

Hemant27031999:
@cdiener , @Midnighter do you suggest some changes or is it fine?

cobra/test/data/update_pickles.py
```diff
@@ -60,8 +60,8 @@
 mini.add_reaction(ecoli_model.reactions.EX_lac__D_e.copy())
 r = cobra.Reaction("D_LACt2")
 mini.add_reaction(r)
-r.gene_reaction_rule = ecoli_model.reactions.D__LACt2pp.gene_reaction_rule
-r.reaction = ecoli_model.reactions.D__LACt2pp.reaction.replace("_p", "_e")
+# r.gene_reaction_rule = ecoli_model.reactions.D__LACt2pp.gene_reaction_rule
+# r.reaction = ecoli_model.reactions.D__LACt2pp.reaction.replace("_p", "_e")
```
Midnighter: If these replacements are no longer required to satisfy the old existing tests, please just delete the lines entirely.

cobra/test/data/update_pickles.py
```diff
@@ -86,10 +86,11 @@
     dump(mini, outfile, protocol=2)
 save_matlab_model(mini, "mini.mat")
 save_json_model(mini, "mini.json", pretty=True)
+save_yaml_model(mini, "mini.yml")
 write_sbml_model(mini, "mini_fbc2.xml")
 write_sbml_model(mini, "mini_fbc2.xml.bz2")
 write_sbml_model(mini, "mini_fbc2.xml.gz")
-write_sbml2(mini, "mini_fbc1.xml", use_fbc_package=True)
+# write_sbml2(mini, "mini_fbc1.xml", use_fbc_package=True)
```
Midnighter: Either delete the line or leave it in, please.

cobra/test/test_io/test_annotation_format.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+from os.path import join
+from cobra.io import load_json_model, write_sbml_model
+
+
+def test_load_json_model_valid(data_directory):
+    """Testing valid annotation format"""
+    path_to_file = join(data_directory, "valid_annotation_format.json")
+    model = load_json_model(path_to_file)
+    dict = {
```
Midnighter: `dict` is a Python builtin type so a different name is better style.

```suggestion
    expected = {
```

cobra/test/test_io/test_annotation_format.py
```diff
@@ -0,0 +1,25 @@
...
+        'bigg.reaction': [['is', 'PFK26']],
+        'kegg.reaction': [['is', 'R02732']],
+        'rhea': [['is', '15656']]
+    }
+    for x in model.metabolites:
```
Midnighter: Should prefer at least three letter variable names.

```suggestion
    for metabolite in model.metabolites:
```

cobra/test/test_io/test_annotation_format.py
```diff
@@ -0,0 +1,25 @@
...
+        assert x.annotation == dict, "Dictionary is not equal"
```
Midnighter: The `pytest` suite already provides deep inspection of the assertion so the message is not necessary.

```suggestion
        assert metabolite.annotation == expected
```

cobra/test/test_io/test_annotation_format.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+from os.path import join
+from cobra.io import load_json_model, write_sbml_model
+
+
+def test_load_json_model_valid(data_directory):
+    """Testing valid annotation format"""
```
Midnighter: The first line of a docstring should be in imperative mood.

```suggestion
    """Test loading a valid annotation from JSON."""
```

cobra/test/test_io/test_annotation_format.py
```diff
@@ -0,0 +1,25 @@
...
+        'rhea': [['is', '15656']]
+    }
+    for x in model.metabolites:
+        assert x.annotation == dict, "Dictionary is not equal"
+    path_to_output = join(data_directory, 'valid_annotation_output.xml')
+    write_sbml_model(model, path_to_output)
+
+
+def test_load_json_model_invalid(data_directory):
+    """Testing invalid annotation format"""
```
Midnighter:
```suggestion
    """Test that loading an invalid annotation from JSON raises."""
```

Hemant27031999:
@Midnighter, I have made the changes. Please review.

Midnighter:
Let me know @cdiener when you're happy, too. I can merge it then and make a new release.

cdiener:
I'm on board if it looks good to you. Seems like CI is failing again due to unsorted imports.

PR #932, opened 2020-02-11
feat: set interface to solver timeout in configuration

carrascomj:
### Description

Allow the user to set the timeout of the solver through an attribute of `cobrapy.Model`, as for the tolerance. The default value in the configuration of this timeout would be `None`.

### DIFF
--- a/cobra/core/configuration.py
+++ b/cobra/core/configuration.py
@@ -7,7 +7,6 @@ from __future__ import absolute_import
 import logging
 import types
 from multiprocessing import cpu_count
-from warnings import warn
 
 from six import string_types, with_metaclass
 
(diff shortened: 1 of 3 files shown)


cdiener:
IMO it's enough to have it only in the configuration. We do not want to expose every optlang config parameter in cobrapy. This is already possible by using `model.solver`. `model.tolerance` was a convenience thing since that is used very frequently and there are some inconsistencies with handling tolerances in optlang (not all solver interpret tolerances the same). Also `model.tolerance` is used in other algorithms to compare values to zero and is thus not completely specific to the solver. Also, `model.timeout` would be to ambiguous (what does the timeout refer to? Solving the model? Getting the primal values?).

carrascomj:
OK, I didn't notice the reasons behind having `model.tolerance`, but now I see your points.

Midnighter:
I agree with your reasoning @cdiener. I did think that it could be nice to have a timeout on the global configuration object but I'm not sure of the exact behavior yet and whether there are any real advantages. Apologies for suggesting to add it here @carrascomj.

PR #933, opened 2020-02-12
Switch to non-experimental python-libsbml

ecederstrand:
From the same maintainers as python-libsbml-experimental. Available at https://pypi.org/project/python-libsbml/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ if __name__ == "__main__":
             "pandas>=0.17.0",
             "optlang>=1.4.2",
             "depinfo",
-            "python-libsbml-experimental==5.18.0",
+            "python-libsbml==5.18.0",
         ],
         tests_require=[
             "jsonschema > 2.5",


ecederstrand:
Appvayor on Python 3.6 is failing due to incompatible `six` version: `virtualenv 20.0.2 has requirement six<2,>=1.12.0, but you'll have six 1.11.0 which is incompatible.`

I'm not sure how to fix it, and I don't think it's related to the changes in this PR.

Midnighter:
Hi, Thank you for the suggested change. Can you say a little bit more about your motivation for this change? As far as I know from @matthiaskoenig the experimental branch is actually stable but contains newer features.

ecederstrand:
Ah, I just assumed by the name that python-libsbml-experimental is deprecated and superseded by python-libsbml.

I ended up here because I'm attempting to upgrade our analysis platform to Python 3.8 (that has `cobrapy` installed) and neither python-libsbml nor python-libsbml-experimental have binary packages for Python 3.8. I expected that requesting binary packages for python-libsbml would have a better chance of being accepted than for the experimental package.

If `python-libsbml-experimental` is preferred for ``cobrapy` than I'll just close this PR.

Midnighter:
Have you created issues on the python-libsbml repos? Otherwise I'm sure @matthiaskoenig can push that agenda.

Midnighter:
I just checked and there is a bioconda package available. So if you can use conda temporarily it might be available for Python 3.8. The bioconda cobra package actually uses python-libsbml and not the 'experimental' version. Something we should consider as well.

cdiener:
Matthias should know before but I mentioned this at some points my and he said that he uses some features that are only available in the experimental version. I can confirm that cobrapy also works with the non-experimental version. This is what the conda version uses right now.

ecederstrand:
Yes, I have filed https://github.com/sbmlteam/python-libsbml/issues/10 to have 3.8 wheels for python-libsbml.

Unfortunately, we can't use conda in our environment.

ecederstrand:
@Midnighter @cdiener Would this change be acceptable, or should I close the PR?

codecov is complaining about lower coverage which seems like a rounding error because I didn't touch any code.

cdiener:
Codecov would not be a blocker to merge here. But we are still waiting for comments from @matthiaskoenig. As a workaround you can install the normal python-libsbml and all other dependencies for cobrapy with pip and then install cobrapy with the `--no-deps` option.

matthiaskoenig:
Sorry for the slow reply.
Yes, `python-libsbml-experimental` is stable and basically `python-libsbml` plus additional packages not yet released. So if people want to use cobrapy with experimental packages they need a `python-libsbml-experimental` installation.
The `experimental` just refers to the packages, but the core and especially also the `fbc` package are the stable ones.

E.g. the `distrib` package is not part of `python-libsbml` but I am using it with cobrapy to encode uncertainties on flux bounds.

ecederstrand:
Thanks for the explanation, @matthiaskoenig. Since my assumption was just wrong, I'll close this PR.

ecederstrand:
@matthiaskoenig Do you know where I should request Python 3.8 whels for  `python-libsbml-experimental`? I got 3.8 whels for `python-libsbml` in https://github.com/sbmlteam/python-libsbml/issues/10 but since cobrapy is not switching, I need them for the other package as well.

Midnighter:
I'm trying to add a conda-forge package for python-libsbml-experimental, in case that helps you (https://github.com/conda-forge/staged-recipes/pull/10905).

Right now it's blocked by the following issue, though: https://github.com/sbmlteam/python-libsbml/issues/15

matthiaskoenig:
@ecederstrand I opened an issue for this (https://github.com/sbmlteam/python-libsbml/issues/17)

PR #940, opened 2020-02-29
Add Python 3.8 and remove EOL version support

ecederstrand:
Python 3.8 was released in late 2019.

Python 2.7 is EOL since January 2020, and 3.4 since March 2019. See https://devguide.python.org/devcycle/#end-of-life-branches. CI test runs on both versions was removed in #939. This also removes them from package metadata.

This change marks removal of support for Python 2. It may deserve a major version bump.


Midnighter:
>  It may deserve a major version bump.

Removal of Python 2 was accepted by the community. There are still a number of targets on the development roadmap before we can make a 1.0 release.

PR #943, opened 2020-03-13
Writing notes for model in SBML file

Hemant27031999:
### What does it fixes?
Currently, if we add some notes to a cobra_model and write that model to SBML, these notes are not written in the file.

Fixes issue #902 

### What was the problem?
Inside the method **_model_to_sbml()** which is used convert the _cobra_model_ to _libsbml.SBMLDocument_ object, the conversion of _notes_ for the _cobra_model_ was not done. Only _annotations_ were converted. So the method for conversion of _notes_ of the model is added here.

### DIFF
--- a/cobra/test/test_io/test_model_notes.py
+++ b/cobra/test/test_io/test_model_notes.py
@@ -1,0 +1,31 @@
+import pytest
+import cobra
+from os.path import join
+from cobra.util.solver import set_objective
+from cobra.io import read_sbml_model, write_sbml_model
+
+
+def test_model_notes(tmp_path):
+    """Testing if model notes are written in SBML"""
+    path_to_file = join(str(tmp_path), "model_notes.xml")
+
+    # making a minimal cobra model to test notes
+    model = cobra.Model("e_coli_core")
+    model.notes["Remark"] = "...Model Notes..."
+    met = cobra.Metabolite("pyr_c", compartment="c")
+    model.add_metabolites([met])
+    met.notes["Remark"] = "I appear."
+    rxn = cobra.Reaction("R_ATPM")
+    model.add_reactions([rxn])
+    rxn.notes["Remark"] = "What about me?"
+    model.objective_direction = "max"
+    coefficients = {}
+    coefficients[rxn] = 1
+    set_objective(model, coefficients)
+    write_sbml_model(model, path_to_file)
+
+    # reading the model back
+    model_after_reading = read_sbml_model(path_to_file)
+
+    # checking if notes are written to model
+    assert model_after_reading.notes["Remark"] == "...Model Notes..."
--- a/cobra/test/test_io/test_notes.py
+++ b/cobra/test/test_io/test_notes.py
@@ -21,13 +21,17 @@ def test_notes(tmp_path):
     model.add_reactions([rxn])
     rxn.notes["Remark"] = "What about me?"
     model.objective_direction = "max"
-    coefficients = {}
-    coefficients[rxn] = 1
-    set_objective(model, coefficients)
+    model.objective = rxn
     write_sbml_model(model, path_to_file)
 
     # reading the model back
     model_after_reading = read_sbml_model(path_to_file)
+    met_after_reading = model_after_reading.metabolites.get_by_id("pyr_c")
+    reaction_after_reading = model_after_reading.reactions.get_by_id("R_ATPM")
 
     # checking if notes are written to model
     assert model_after_reading.notes["Remark"] == "...Model Notes..."
+
+    # checking notes for metabolite and reaction
+    assert met_after_reading.notes["Remark"] == "I appear."
+    assert reaction_after_reading.notes["Remark"] == "What about me?"
--- a/cobra/test/test_io/test_notes.py
+++ b/cobra/test/test_io/test_notes.py
@@ -1,11 +1,13 @@
-import pytest
-import cobra
 from os.path import join
-from cobra.util.solver import set_objective
+
+import pytest
+
+import cobra
 from cobra.io import read_sbml_model, write_sbml_model
+from cobra.util.solver import set_objective
 
 
-def test_model_notes(tmp_path):
+def test_notes(tmp_path):
     """Testing if model notes are written in SBML"""
     path_to_file = join(str(tmp_path), "model_notes.xml")
 
--- a/cobra/test/test_io/test_notes.py
+++ b/cobra/test/test_io/test_notes.py
@@ -4,7 +4,6 @@ import pytest
 
 import cobra
 from cobra.io import read_sbml_model, write_sbml_model
-from cobra.util.solver import set_objective
 
 
 def test_notes(tmp_path):
@@ -16,7 +15,7 @@ def test_notes(tmp_path):
     model.notes["Remark"] = "...Model Notes..."
     met = cobra.Metabolite("pyr_c", compartment="c")
     model.add_metabolites([met])
-    met.notes["Remark"] = "I appear."
+    met.notes["Remark"] = "Note with \n newline"
     rxn = cobra.Reaction("R_ATPM")
     model.add_reactions([rxn])
     rxn.notes["Remark"] = "What about me?"
@@ -33,5 +32,5 @@ def test_notes(tmp_path):
     assert model_after_reading.notes["Remark"] == "...Model Notes..."
 
     # checking notes for metabolite and reaction
-    assert met_after_reading.notes["Remark"] == "I appear."
+    assert met_after_reading.notes["Remark"] == "Note with \n newline"
     assert reaction_after_reading.notes["Remark"] == "What about me?"
--- a/cobra/io/sbml.py
+++ b/cobra/io/sbml.py
@@ -101,7 +101,7 @@ SBML_DOT = "__SBML_DOT__"
 # -----------------------------------------------------------------------------
 pattern_notes = re.compile(
     r"<(?P<prefix>(\w+:)?)p[^>]*>(?P<content>.*?)</(?P=prefix)p>",
-    re.IGNORECASE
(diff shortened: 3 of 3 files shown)


cdiener:
Isn't `_sbase_notes_dict` a legacy reader for SBML 2 models? I think notes should be passed as string since they are meant for human consumption only and should not be parsed.

Hemant27031999:
> I think notes should be passed as string since they are meant for human consumption only and should not be parsed

I am sorry for that. I think it is wrong to use the word "parse" here. I have updated the above comment. Please have a look at it. What I meant was simply "converting/writing" the notes to SBML file as string which is not happening here with the model but working fine for other components,

cdiener:
Ok that would be great. It does look like it's always using ` _sbase_notes_dict` (I may be wrong here) which does parse the notes to a dict. What happens if the notes are not in HTML but in a different format here?

Hemant27031999:
> It does look like it's always using _sbase_notes_dict (I may be wrong here)

Ya that's correct because no restriction is placed before calling this method. But I think we don't need to place any restrictions.

> which does parse the notes to a dict

Actually, `_sbase_notes_dict` works the other way round. It takes in a dictionary and converts it to notes of an sbase object (in HTML format). Here is the method :

```
def _sbase_notes_dict(sbase, notes):
    """Set SBase notes based on dictionary.

    Parameters
    ----------
    sbase : libsbml.SBase
        SBML object to set notes on
    notes : notes object
        notes information from cobra object
    """
    if notes and len(notes) > 0:
        tokens = ['<html xmlns = "http://www.w3.org/1999/xhtml" >'] + \
                 ["<p>{}: {}</p>".format(k, v) for (k, v) in notes.items()] + \
                 ["</html>"]
        _check(
            sbase.setNotes("\n".join(tokens)),
            "Setting notes on sbase: {}".format(sbase)
        )
```

So the notes would be in a dictionary form and we have to attach it to sbase object (model in this case) and don't have to place any restriction.

Hemant27031999:
I have added the test cases for testing model notes, @Midnighter @cdiener can someone please review them.
Thanks

cdiener:
I think I don't full understand what this PR is doing. Since notes should not be interpreted by cobrapy I think the best behavior would be to just have them be strings that are passed on to the model. I am not sure why they need to be dicts... This is what annotations are for. Could you add a behavioral example of the PR? What would happen with the following commands (current output shown for reference):

```Python

In [2]: from cobra.io import read_sbml_model, write_sbml_model                                                                                                                

In [4]: mod.notes = "This should not be modified by the parser :O"                                                                                                            

In [5]: mod.reactions.PFK.notes = "I am another note \n an a new line"                                                                                                        

In [6]: write_sbml_model(mod, "notes_mod.xml")                                                                                                                                
AttributeError                            Traceback (most recent call last)
<ipython-input-6-3290079bd026> in <module>
----> 1 write_sbml_model(mod, "notes_mod.xml")

~/.local/lib/python3.8/site-packages/cobra/io/sbml.py in write_sbml_model(cobra_model, filename, f_replace, **kwargs)
    886     f_replace: dict of replacement functions for id replacement
    887     """
--> 888     doc = _model_to_sbml(cobra_model, f_replace=f_replace, **kwargs)
    889 
    890     if isinstance(filename, string_types):

~/.local/lib/python3.8/site-packages/cobra/io/sbml.py in _model_to_sbml(cobra_model, f_replace, units)
   1073         reaction.setReversible((cobra_reaction.lower_bound < 0))
   1074         _sbase_annotations(reaction, cobra_reaction.annotation)
-> 1075         _sbase_notes_dict(reaction, cobra_reaction.notes)
   1076 
   1077         # stoichiometry

~/.local/lib/python3.8/site-packages/cobra/io/sbml.py in _sbase_notes_dict(sbase, notes)
   1334     if notes and len(notes) > 0:
   1335         tokens = ['<html xmlns = "http://www.w3.org/1999/xhtml" >'] + \
-> 1336                  ["<p>{}: {}</p>".format(k, v) for (k, v) in notes.items()] + \
   1337                  ["</html>"]
   1338         _check(

AttributeError: 'str' object has no attribute 'items'

In [7]: mod.reactions.PFK.notes = {"note": "I am another note \n an a new line"}                                                                                              

In [8]: mod.notes = {"note": "This should not be modified by the parser :O"}                                                                                                  

In [9]: write_sbml_model(mod, "notes_mod.xml")                                                                                                                                

In [10]: mod = read_sbml_model("notes_mod.xml")                                                                                                                               

In [11]: mod.notes                                                                                                                                                            
Out[11]: {}

In [12]: mod.reactions.PFK.notes                                                                                                                                              
Out[12]: {}
```

For the commands I above I would expect that setting via string should work and that notes are conserved but not modified by cobrapy.

Hemant27031999:
This PR is simply referring the issue #902 , where the notes corresponding to reaction and metabolites were being written to the SBML file, but notes corresponding to "model" were not written. To give you a behavioral example, consider the above commands you mentioned, working upon this PR : 

```
>>> import cobra
>>> from cobra.test import create_test_model
>>> from cobra.io import read_sbml_model, write_sbml_model
>>> mod = create_test_model("textbook")
>>> mod.notes = "This should not be modified by the parser :O"
>>> mod.reactions.PFK.notes = "I am another note \n an a new line"
>>> write_sbml_model(mod, "notes_mod.xml")
  File "/home/hemant/Desktop/cobrapy/cobra/io/sbml.py", line 888, in write_sbml_model
    doc = _model_to_sbml(cobra_model, f_replace=f_replace, **kwargs)
  File "/home/hemant/Desktop/cobrapy/cobra/io/sbml.py", line 941, in _model_to_sbml
    _sbase_notes_dict(model, cobra_model.notes)
  File "/home/hemant/Desktop/cobrapy/cobra/io/sbml.py", line 1339, in _sbase_notes_dict
    ["<p>{}: {}</p>".format(k, v) for (k, v) in notes.items()] + \
AttributeError: 'str' object has no attribute 'items'
```
This PR is not handling this issue. It is thrown because the current version of cobrapy expects notes in the form of dictionary having key-value pairs, but here you have passed a string directly to notes field, which is expecting the notes to be of dictionary type.
```
>>> mod.reactions.PFK.notes = {"note": "I am another note \n an a new line"}
>>> mod.notes = {"note": "This should not be modified by the parser :O"}
>>> write_sbml_model(mod, "notes_mod.xml")
>>> mod = read_sbml_model("notes_mod.xml")
>>> mod.notes
{'note': 'This should not be modified by the parser :O'}
```
This is the issue handled in this PR. The model notes are now written to SBML file (and hence are read back to cobra model). Initially, it was not happening. Moving further in the same shell
```
>>> mod.reactions.PFK.notes
{}
```
This PR is not handling this issue of notes not being read when there is a new line character. If newline character is inserted, then the notes string does not match the pattern specified by the pattern_notes

> pattern_notes = re.compile(
>     r"<(?P<prefix>(\w+:)?)p[^>]*>(?P<content>.*?)</(?P=prefix)p>",
>     re.IGNORECASE
> )

and hence it is not read (though written in the SBML file).
Continuing in the same shell but removing new line charachter from reaction notes
```
>>> mod.reactions.PFK.notes = {"note": "I am another note in the same line"}
>>> write_sbml_model(mod, "notes_mod.xml")
>>> mod = read_sbml_model("notes_mod.xml")
>>> mod.reactions.PFK.notes
{'note': 'I am another note in the same line'}
```

So this PR is simply handling the issue of model notes not getting written to SBML file. Here you can see now, the model notes are written and even read again to cobra model.

cdiener:
Cool thanks for the clarification. This makes sense now.

cobra/test/test_io/test_model_notes.py
```diff
@@ -0,0 +1,31 @@
+import pytest
```
cdiener: Those import have to be sorted alphabetically to have the CI pass.

Hemant27031999: @cdiener, I have made the changes. 
Thanks.

cobra/test/test_io/test_notes.py
```diff
@@ -0,0 +1,33 @@
...
+    met = cobra.Metabolite("pyr_c", compartment="c")
+    model.add_metabolites([met])
+    met.notes["Remark"] = "I appear."
+    rxn = cobra.Reaction("R_ATPM")
+    model.add_reactions([rxn])
+    rxn.notes["Remark"] = "What about me?"
+    model.objective_direction = "max"
+    coefficients = {}
+    coefficients[rxn] = 1
+    set_objective(model, coefficients)
```
Midnighter: I think the more straight forward way to set an objective is:

```suggestion
    model.objective = rxn
```

cobra/test/test_io/test_notes.py
```diff
@@ -0,0 +1,33 @@
...
+    write_sbml_model(model, path_to_file)
+
+    # reading the model back
+    model_after_reading = read_sbml_model(path_to_file)
+
+    # checking if notes are written to model
+    assert model_after_reading.notes["Remark"] == "...Model Notes..."
```
Midnighter: You add notes to a metabolite and reaction in the test. I would either assert that the are correct, too, here or don't add them at all.

Hemant27031999:
@Midnighter, I have made the changes. Please review.

Midnighter:
I'm not completely sure right now but maybe adding the `re.MULTILINE` flag to the pattern that you mention above would solve the newline issues?

Hemant27031999:
@Midnighter, [re.MULTILINE](https://docs.python.org/3/library/re.html#re.match) checks only at the end or at the start of each line as specified in the doc. The requirement was [re.DOTALL](https://docs.python.org/3/library/re.html#re.DOTALL). It accepts the newline character also which appears within the text and then verifies the pattern. 
I have modified the "pattern_notes" accordingly and have also added the test for the newline. It is now working fine. Please review.

PR #945, opened 2020-04-07
Update faq.ipynb

schultzdre:
Included SAMMI as an alternative visualization tool.

### DIFF
--- a/documentation_builder/faq.ipynb
+++ b/documentation_builder/faq.ipynb
@@ -335,7 +335,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "cobrapy works well with the [escher](https://escher.github.io/) package, which is well suited to this purpose. Consult the [escher documentation](https://escher.readthedocs.org/en/latest/) for examples."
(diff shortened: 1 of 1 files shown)


cdiener:
You might want to add it to https://github.com/opencobra/cobrapy-website as well so it can be highlighted at https://opencobra.github.io/cobrapy/packages/.

schultzdre:
Just added it to the website. Thanks!

Midnighter:
@cdiener have you tried out SAMMI? I haven't worked with it since seeing it at COBRA '18. It feels to me like we should have some positive experience with SAMMI before we recommend it in the documentation? Or do you see it as a neutral list of existing tools?

cdiener:
No idea, I was under the impression that it was #2 (neutral list). But I think I would actually prefer removing all tool names and linking directly to the visualization tag on the website (https://opencobra.github.io/cobrapy/tags/visualization/). This would be easier to maintain.

Midnighter:
I have implemented @cdiener's suggestion. Closing this PR in favour of that. Thank you for suggesting it, though @schultzdre

PR #949, opened 2020-04-14
fix: SBO terms for infinity bounds

BenjaSanchez:
I think `SBO:0000626` is more appropiate than `SBO:0000625` for infinity bounds. From the SBO documentation:

* [SBO:0000625](http://www.ebi.ac.uk/sbo/main/SBO:0000625): _"A parameter that limits the upper or lower value that a flux may assume. This parameter may be determined experimentally, or may be the result of theoretical investigation."_
* [SBO:0000626](http://www.ebi.ac.uk/sbo/main/SBO:0000626): _"A value used for flux bound in cases where a precise value, supported experimentally or theoretically, is not available."_


matthiaskoenig:
This changes looks good. Best Matthias

Issue #856, opened 2019-06-17
JSON format does not support -inf/inf bounds


matthiaskoenig:
Trying to convert models with -inf/inf bounds to json fails with
```
Traceback (most recent call last):
  File "/home/mkoenig/git/pancreas_model/pypancreas/pancreas/model_factory.py", line 51, in <module>
    cobra.io.save_json_model(model, MODEL_JSON)
  File "/home/mkoenig/envs/pancreas/lib/python3.6/site-packages/cobra/io/json.py", line 110, in save_json_model
    json.dump(obj, file_handle, **dump_opts)
  File "/usr/lib/python3.6/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/usr/lib/python3.6/json/encoder.py", line 430, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.6/json/encoder.py", line 404, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.6/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.6/json/encoder.py", line 396, in _iterencode_dict
    yield _floatstr(value)
  File "/usr/lib/python3.6/json/encoder.py", line 241, in floatstr
    repr(o))
ValueError: Out of range float values are not JSON compliant: -inf
```
-inf/inf values are not JSON serializable, consequently such models can currently not correctly be represented in JSON.

cdiener:
The [JSON spec](https://tools.ietf.org/html/rfc7159#section-6) does not allow infinity values so those should be converted to `null`.

matthiaskoenig:
Sounds like a good solution. 
Just have to make sure that the JSON importer and exporter are handling the conversions. There could still exist some strange corner cases were we loose information, e.g. somebody decided to have bounds of `(-INF, -INF)` or `(INF, INF)` (not sure why, but is a valid LP).
The `cobrapy-json-cobrapy` roundtrip would be `(-INF, -INF) -> (null, null) -> (-INF, INF)`. We should give a warning in such cases, i.e. exporting `(-INF, -INF)` or `(INF, INF)`

Midnighter:
Another option is to transform it to a string since Python is perfectly capable of parsing `float("inf")`, `float("-inf")`, or even `float("nan")`.

matthiaskoenig:
[...]

PR #957, opened 2020-05-26
Handling "Inf" bounds in JSON. 

Hemant27031999:
## What does it fixes?
This PR is fixing issue #856 i.e., reading and writing of "inf" bounds in JSON format.


Hemant27031999:
@matthiaskoenig @draeger @Midnighter @cdiener Please review it.

cobra/io/dict.py
```diff
@@ -121,7 +121,12 @@ def reaction_to_dict(reaction):
...
     for key in _REQUIRED_REACTION_ATTRIBUTES:
         if key != "metabolites":
-            new_reaction[key] = _fix_type(getattr(reaction, key))
+            if key == "lower_bound" and reaction.lower_bound in {float('inf'), float('-inf'), float('nan')}:
+                new_reaction[key] = str(_fix_type(getattr(reaction, key)))
+            elif key == "upper_bound" and reaction.upper_bound in {float('inf'), float('-inf'), float('nan')}:
+                new_reaction[key] = str(_fix_type(getattr(reaction, key)))
+            else:
+                new_reaction[key] = _fix_type(getattr(reaction, key))
             continue
```
draeger: The list `_REQUIRED_REACTION_ATTRIBUTES` is a constant list containing 6 entries at the moment. The loop iterates through all 6 elements but for 3 of them there are special cases. Depending on how often the function `reaction_to_dict(reaction)` is executed, this could have some run-time implications. It would be better not to use a loop at all but to cover the 6 cases directly.

cobra/io/dict.py
```diff
@@ -147,7 +152,10 @@ def reaction_from_dict(reaction, model):
                 (model.metabolites.get_by_id(str(met)), coeff)
                 for met, coeff in iteritems(v)))
         else:
-            setattr(new_reaction, k, v)
+            if k == "lower_bound" or k == "upper_bound":
+                setattr(new_reaction, k, float(v))
+            else:
+                setattr(new_reaction, k, v)
```
draeger, line 158: It is not clear why a loop is needed to cover 6 cases half of which receive a special treatment within the loop (`metabolites`, `lower_bound`, and `upper_bound`). Why not also treating the remaining 3 cases directly?

cobra/io/dict.py
Hemant27031999: > It would be better not to use a loop at all but to cover the 6 cases directly.

 
Well, I can't say much about it. Each component of the COBRA model (reaction, metabolite, gene) has separate lists to define their required and optional attributes. That's how it was initially done, so doing it separately for the reaction would make it a little out of the order.

> Depending on how often the function reaction_to_dict(reaction) is executed, this could have some run-time implications.

That's surely a point to be noted. The extra comparisons introduced to handle `inf` values for lower or upper bound can be handled more efficiently if made outside the `for` loop. It will reduce the number of comparisons and will surely increase the efficiency.

cobra/io/dict.py
Hemant27031999: Regarding the use of `for` loop, again, I would request COBRApy community to provide their review for it.

cobra/io/dict.py
```diff
@@ -121,7 +122,16 @@ def reaction_to_dict(reaction):
     new_reaction = OrderedDict()
     for key in _REQUIRED_REACTION_ATTRIBUTES:
         if key != "metabolites":
-            new_reaction[key] = _fix_type(getattr(reaction, key))
+            if key == "lower_bound" and (
+                    np.isnan(reaction.lower_bound) or
```
matthiaskoenig, line 126: `not np.isfinite(reaction.lower_bound)` is the test you want; https://numpy.org/doc/stable/reference/generated/numpy.isfinite.html

cobra/io/dict.py
```diff
@@ -121,7 +122,16 @@ def reaction_to_dict(reaction):
...
+                    np.isinf(reaction.lower_bound)):
+                new_reaction[key] = str(_fix_type(getattr(reaction, key)))
+            elif key == "upper_bound" and (
+                    np.isnan(reaction.upper_bound) or
+                    np.isinf(reaction.upper_bound)):
```
matthiaskoenig, line 131: use `np.isfinite

cobra/io/dict.py
matthiaskoenig: Loop looks fine. Is standard practise here. There could be additional attributes in the future.

Midnighter:
Merged this yesterday @matthiaskoenig. Feel free to open a new PR with your suggested changes.

PR #958, opened 2020-05-26
FIX: Allow comparment names with two letters

weilandtd:
New yeast models are provided with 2 letter compartment names.  Currently, the import of Yeast models from Matlab will omit the two-letter compartments.

### DIFF
--- a/cobra/io/mat.py
+++ b/cobra/io/mat.py
@@ -25,18 +25,18 @@ except ImportError:
 
 
 # precompiled regular expressions
-_bracket_re = re.compile(r"\[[a-z]\]$")
-_underscore_re = re.compile(r"_[a-z]$")
+_bracket_re = re.compile(r"\[[a-z]{1,2}\]$")
+_underscore_re = re.compile(r"_[a-z]{1,2}$")
 
 
 def _get_id_compartment(id):
     """extract the compartment from the id string"""
     bracket_search = _bracket_re.findall(id)
     if len(bracket_search) == 1:
-        return bracket_search[0][1]
+        return bracket_search[0][1:]
     underscore_search = _underscore_re.findall(id)
     if len(underscore_search) == 1:
-        return underscore_search[0][1]
+        return underscore_search[0][1:]
     return None
 
 
--- a/cobra/io/mat.py
+++ b/cobra/io/mat.py
@@ -33,7 +33,7 @@ def _get_id_compartment(id):
     """extract the compartment from the id string"""
     bracket_search = _bracket_re.findall(id)
     if len(bracket_search) == 1:
-        return bracket_search[0][1:]
+        return bracket_search[0][1:-1]
     underscore_search = _underscore_re.findall(id)
     if len(underscore_search) == 1:
         return underscore_search[0][1:]
--- a/cobra/io/mat.py
+++ b/cobra/io/mat.py
@@ -25,7 +25,7 @@ except ImportError:
 
 
 # precompiled regular expressions
-_bracket_re = re.compile(r"\[[a-z]{1,3}\]$")
+_bracket_re = re.compile(r"\[[a-z]+\]$")
 _underscore_re = re.compile(r"_[a-z]{1,3}$")
 
 
--- a/cobra/io/mat.py
+++ b/cobra/io/mat.py
@@ -25,8 +25,8 @@ except ImportError:
 
 
 # precompiled regular expressions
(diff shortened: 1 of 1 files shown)


cobra/io/mat.py
```diff
@@ -25,18 +25,18 @@
...
+_bracket_re = re.compile(r"\[[a-z]{1,3}\]$")
+_underscore_re = re.compile(r"_[a-z]{1,3}$")
 
 
 def _get_id_compartment(id):
     """extract the compartment from the id string"""
     bracket_search = _bracket_re.findall(id)
     if len(bracket_search) == 1:
-        return bracket_search[0][1]
+        return bracket_search[0][1:]
```
Midnighter, line 36: For a single letter, this was extracting just the compartment letter. So to achieve the same with your new expression, you will have to ignore the closing square bracket.

```suggestion
        return bracket_search[0][1:-1]
```

cobra/io/mat.py
```diff
@@ -25,18 +25,18 @@
...
     underscore_search = _underscore_re.findall(id)
     if len(underscore_search) == 1:
-        return underscore_search[0][1]
+        return underscore_search[0][1:]
```
Midnighter, line 39: Same as above

```suggestion
        return underscore_search[0][1:-1]
```

cobra/io/mat.py
```diff
@@ -25,18 +25,18 @@
 
 
 # precompiled regular expressions
-_bracket_re = re.compile(r"\[[a-z]\]$")
-_underscore_re = re.compile(r"_[a-z]$")
+_bracket_re = re.compile(r"\[[a-z]{1,3}\]$")
```
Midnighter, line 28: Is there any reason not to match everything inside of the square brackets?

```suggestion
_bracket_re = re.compile(r"\[[a-z]+\]$")
```

cobra/io/mat.py
weilandtd: Dear Moritz, 
yes I agree, I overlocked this one I was only testing my changes for a models with underscore notation.
I will test your suggestions for models with the bracket notation. 
For the test it is not necessary to have Matlab licenses, an exported model with the respective notation is sufficient. 

Best wishes,
Daniel

cobra/io/mat.py
weilandtd: No there is not I form the yeast models perspective. 

For the underscore notation I just wanted to play it save so met names without compartment names in the id are not accidentally read as such.

Midnighter:
Hi @weilandtd, can you apply the following patch to your branch, please? (I don't have commit rights to your branch so I couldn't add the commit.) Save the contents to `file.path`, then apply it using `git am < file.path` and test that it works as expected. I think then it is good to be merged.

```
From 91ce894a01f773e8613c36b22eed3ff773504178 Mon Sep 17 00:00:00 2001
From: Midnighter <midnighter@posteo.net>
Date: Tue, 7 Jul 2020 22:54:38 +0200
Subject: [PATCH] refactor: search for named groups

---
 cobra/io/mat.py | 16 ++++++++--------
 1 file changed, 8 insertions(+), 8 deletions(-)

diff --git a/cobra/io/mat.py b/cobra/io/mat.py
index ac38f232..36033f2c 100644
--- a/cobra/io/mat.py
+++ b/cobra/io/mat.py
@@ -25,18 +25,18 @@ except ImportError:
 
 
 # precompiled regular expressions
-_bracket_re = re.compile(r"\[[a-z]+\]$")
-_underscore_re = re.compile(r"_[a-z]{1,3}$")
+_bracket_re = re.compile(r"\[(?P<compartment>[a-z]+)\]$")
+_underscore_re = re.compile(r"_(?P<compartment>[a-z]+)$")
 
 
 def _get_id_compartment(id):
     """extract the compartment from the id string"""
-    bracket_search = _bracket_re.findall(id)
-    if len(bracket_search) == 1:
-        return bracket_search[0][1:-1]
-    underscore_search = _underscore_re.findall(id)
-    if len(underscore_search) == 1:
-        return underscore_search[0][1:]
+    bracket_search = _bracket_re.search(id)
+    if bracket_search:
+        return bracket_search.group("compartment")
+    underscore_search = _underscore_re.search(id)
+    if underscore_search:
+        return underscore_search.group("compartment")
     return None
 
 
-- 
2.17.1

```

Midnighter:
In the interest of moving this along I will merge and add the patch myself. Thank you for contributing @weilandtd :slightly_smiling_face:

PR #956, opened 2020-05-26
modified json schema and put into a separate file

Hemant27031999:
## What does it fixes?
This PR is solving issue #720.

## Additional fixes.
I have also extracted the json schema from `json.py` file and have put it in a separate file.
Modifying the json schema also passed the two test cases inside `test_json.py` which were initially marked as xfail.

PR #969, opened 2020-07-11
Re-organize the package structure and other infrastructure

### LINKED ISSUE #787: Suggestion to re-organize the repository
Hey @opencobra/cobrapy-core 

I'm proposing two changes to the organization of the repository structure:

1. Move the `cobra` package into a `src/` directory.
2. Move the test suite out of the package itself into its own `tests` directory.

The second point would go hand in hand with moving the `load_model` logic from cameo into cobrapy so we have an alternative to the common strategy of using `from cobra.test import create_test_model`.

I'm using this structure on a number of other packages with great success. For a bit more background, please take a look at this blog post http://andrewsforge.com/article/python-new-package-landscape/

Please let me know if you agree or disagree with these changes. If you're not feeling motivated to argue a thumbs up or down is enough.

Midnighter:
Fix #787 

This PR updates a number of configuration files. There are a lot of changes but I tried to be rather atomic so I hope each commit is clear in its intention.

I think these changes should be applied soon and then the following points still need to be tackled:

1. Documentation is due for a big overhaul.
2. Switch from pep8 to flake8 requires a lot of work in fixing source modules.
3. Whenever we work on a module, we should spend some time on updating to Python 3.6 syntax.
4. We should include the cameo `load_model` logic in some way.
5. Possibly contentious issue: I vote for moving the tests out of the source tree and using `load_model` to replace `create_test_model`.


Midnighter:
Actually, @opencobra/cobrapy-core would love to get many insights here. I consider most of these changes just upgrades but they are quite sweeping changes. Please consider the individual commits and do not be put off by the large number of changes overall. I'd feel more confident in tackling the other open PRs when this upgraded infrastructure is in place.

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+    optimization
+    flux-balance analysis
+    reconstruction
+
+[options]
+zip_safe = True
+install_requires =
+    depinfo
+    future
+    numpy ~= 1.13
```
cdiener, line 46: I would prefer minimum versions instead of locked minor versions. We often go a while without releases and that might block security updates for other packages. Also I think we can lift the lock fro python-libsbml-experimental. The recent versions always had wheels.

setup.cfg
```diff
@@ -1,3 +1,78 @@
...
+[options.package_data]
+cobra = test/data/*
+
+[options.extras_require]
+array =
+    scipy
+development =
+    black
+    bumpversion
+    isort
```
cdiener, line 71: Does isort play nice with black yet? A while ago those would conflict (bacl would overwrite isort and vice versa), but I haven't checked in a while.

cdiener (changes requested):
In general it looks great :tada: Love the switch to Github Actions. Will require way less tokens to be passed around. Would prefer not locking minor versions (major is okay) for dependencies.

Also we should be aware that this will break every open PR right now quite badly.

setup.cfg
Midnighter: In our current setup we use `numpy >=1.13`. This notation `numpy ~= 1.13` is a shorthand for `numpy >=1.13,<2`. Since major versions are allowed to contain breaking API changes I thought it's a good idea to be cautious here.

I'm less certain about python-libsbml-experimental, in the past, creation of the wheels has often been delayed by days to weeks, leading to issues for users with libSBML compilation. I would see how this goes in the next two releases before feeling confident in removing the pin.

setup.cfg
Midnighter: I've put the following isort configuration into the `tox.ini`:

```ini
[isort]
skip = __init__.py
line_length = 88
indent = 4
multi_line_output = 3
include_trailing_comma = true
force_grid_wrap = 0
use_parentheses = true
lines_after_imports = 2
```

which makes it compatible with black in my experience.

Midnighter:
> Also we should be aware that this will break every open PR right now quite badly.

This is an important point. Which PRs should we merge before this one then? Some of the PRs are unfortunately quite stale.

setup.cfg
cdiener: Oh yeah my bad, I misunderstood the docs and thought that would enforce numpy 1.13.x. If it only looks major versions that's perfect with me.

cdiener:
All PRs we can merge. I can refactor the OSQP one so that one can stay open.

cdiener (approved):

Midnighter:
I will wait for https://github.com/opencobra/cobrapy/pull/956 to be merged and then rebase this one and I think it's ready.

PR #975, opened 2020-07-25
refactor: make `cobra.util` Python 3.6+

synchon:
This PR addresses a part of #972 and makes `cobra.util` along with its tests compatible with Python 3.6+ only.

### DIFF
--- a/src/cobra/test/test_util/test_array.py
+++ b/src/cobra/test/test_util/test_array.py
@@ -1,8 +1,4 @@
-# -*- coding: utf-8 -*-
-
 """Test functions of array.py"""
-
-from __future__ import absolute_import
 
 import numpy as np
 import pytest
--- a/src/cobra/test/test_util/test_array.py
+++ b/src/cobra/test/test_util/test_array.py
@@ -6,12 +6,6 @@ import pytest
 from cobra.util import create_stoichiometric_matrix
 
 
-try:
-    import scipy
-except ImportError:
-    scipy = None
-
-
 def test_dense_matrix(model):
     S = create_stoichiometric_matrix(model, array_type="dense", dtype=int)
     assert S.dtype == int
@@ -29,8 +23,8 @@ def test_dense_matrix(model):
     assert np.allclose(mass_balance, 0)
 
 
-@pytest.mark.skipif(not scipy, reason="Sparse array methods require scipy")
 def test_sparse_matrix(model):
+    scipy = pytest.importorskip("scipy")
     sparse_types = ["dok", "lil"]
 
     solution = model.optimize()
--- a/src/cobra/test/test_util/test_solver.py
+++ b/src/cobra/test/test_util/test_solver.py
@@ -1,8 +1,4 @@
-# -*- coding: utf-8 -*-
-
 """Test functions of solver.py"""
-
-from __future__ import absolute_import
 
 import numpy as np
 import pytest
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -134,12 +134,9 @@ Before you submit a pull request, check that it meets these guidelines:
    usage in ``documentation_builder`` (documentation is written as
    jupyter notebooks in the ``documentation_builder`` directory, which
    are then converted to rst by the ``autodoc.sh`` script.)
-3. The pull request should work for Python 2.7, 3.4 and 3.5. Check
-   https://travis-ci.org/biosustain/cobrapy/pull_requests
-   and make sure that the tests pass for all supported Python versions.
-4. Assign a reviewer to your pull request. If in doubt, assign Henning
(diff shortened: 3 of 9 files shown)


src/cobra/test/test_util/test_array.py
```diff
@@ -1,9 +1,5 @@
-# -*- coding: utf-8 -*-
-
 """Test functions of array.py"""
 
-from __future__ import absolute_import
-
 import numpy as np
 import pytest
```
matthiaskoenig, line 5: Remove the 
``` 
try:
    import scipy
except ImportError:
    scipy = None
```

Not used here, unnecessary.

src/cobra/test/test_util/test_solver.py
```diff
@@ -1,9 +1,5 @@
-# -*- coding: utf-8 -*-
-
 """Test functions of solver.py"""
```
matthiaskoenig, line 2: Replace string concatenation with fstrings:
``` 
optlang_solvers = ["optlang-" + s for s in stable_optlang if s in su.solvers]
```
```
optlang_solvers = [f"optlang-{s}" for s in stable_optlang if s in su.solvers]
``
Much clearer to read and faster.

**Review by matthiaskoenig — CHANGES_REQUESTED** [review]

**Comment by matthiaskoenig** [comment]
Please add type annotations to the function signatures whenever possible (i.e. where the types are clear). 
This allows automatic checking of the code base and also provides many helpers to users of cobrapy.
Type annotations are one of the best >= python3.6 features and libraries should make use of them.

File: src/cobra/test/test_util/test_array.py (hunk 1 above)
Comment (synchon) [reply src/cobra/test/test_util/test_array.py:5]: It's being used for:
```python
@pytest.mark.skipif(not scipy, reason="Sparse array methods require scipy")
```

synchon:
Yes I agree and strongly support type annotations. I thought we could tackle it in a separate PR but doing it here is not bad either.

matthiaskoenig:
I would do all the python3.6 changes at once. I.e.

- fix imports
- remove object from class definition
- remove six usage
- remove future imports
- remove encodings
- use fstrings when possible
- type annotations

Otherwise we will go over all code 3 times. Let's do this in a single round if possible.

synchon:
I propose to capitalize variable names in module's global scope, for example:
```python
qp_solvers = ["cplex", "gurobi"]
```
to
```python
QP_SOLVERS = ["cplex", "gurobi"]
```
I feel this improves readability and also indicates that it's in the global scope. Any thoughts on this?

Midnighter:
@synchon to me it depends on the intention of the global (module-level) variable. I would use uppercase when it is intended as a constant and lowercase when it is used as an object.

I think I introduced it myself some time ago but nowadays I would, for example, use lowercase for the logger object.

```py
logger = logging.getLogger(__name__)
```

Your example seems like a constant so I think uppercase is the right way to go. Unless a list is required the constant aspect could be made even stronger by:

```py
QP_SOLVERS = frozenset(["cplex", "gurobi"])
```

Midnighter:
@matthiaskoenig I have lifted your comment into #972.

Midnighter:
@synchon I would really go module by module rather than tackling a whole sub-package. There are more things that we can tackle here in addition to what Matthias mentioned. New code should be formatted with `isort` and `black`, and I would like us to slowly move towards passing `flake8`. That especially means that we need to give the docstrings some love. This is currently not part of the pipeline because there are too many errors but locally you can run `tox -e flake8 path/to/module.py` to check. When writing docstrings please follow https://numpydoc.readthedocs.io/en/latest/format.html

I will add this to the issue, too.

src/cobra/test/test_util/test_array.py
Midnighter: I would use https://docs.pytest.org/en/stable/skipping.html#skipping-on-a-missing-import-dependency

synchon:
> I would use uppercase when it is intended as a constant and lowercase when it is used as an object.

Yes exactly. That's what I use for my personal projects as well.

synchon:
@Midnighter For the further commits, I'll go atomic at the module level. I'm using `isort` and `black` currently, I'll also add `flake8` for local development. I'm working on the docstrings as well.

Midnighter:
@synchon if you can fix the scipy import check in the tests I'm actually tempted to pull in these changes now. I think it will  be easier to create separate quick PRs for each module rather than adding to this one.

synchon:
@Midnighter sorry, I was a bit caught up with work through the week, but I'll push the change for scipy import check. I have local changes for the `cobra.util` subpackage and I'll push them in separate PRs.

synchon:
The CI is failing due to #978 but it improves the scipy import check.

Midnighter (approved):
Pulling this in for now. We know that the tests fail for other reasons. Thank you for your work Synchon.

PR #980, opened 2020-08-03
Replace unsupported set indexing for pandas DataFrames

### LINKED ISSUE #978: Deprecation of pandas data frame frozenset indexing breaks our CI
We have some consistently failing tests ~since the pytest 6.0 release~. https://github.com/opencobra/cobrapy/actions/runs/190607600

```
FAILED src/cobra/test/test_flux_analysis/test_deletion.py::test_single_gene_deletion_fba[glpk]
FAILED src/cobra/test/test_flux_analysis/test_deletion.py::test_single_gene_deletion_linear_moma[glpk]
FAILED src/cobra/test/test_flux_analysis/test_deletion.py::test_single_reaction_deletion[glpk]
FAILED src/cobra/test/test_flux_analysis/test_deletion.py::test_double_gene_deletion
FAILED src/cobra/test/test_flux_analysis/test_deletion.py::test_double_reaction_deletion
```

cdiener:
For the deletions results were indexed by frozensets of the knocked genes/reactions. Unfortunately this is broken in recent pandas versions and seems to never have been officially supported (https://github.com/pandas-dev/pandas/issues/7787). The only allowed index types are list/tuple like. 

This PR changes the set indexer to a sorted tuple of unique reactions. This fixes the tests but does not have the prior behavior where `solution[frozenset("X", "Y")]` is the same as `solution[frozenset("Y", "X")]` (indexing with frozenset or set will not work with pandas anymore). The prior behavior is definitely better. If we want to maintain all deletion functions can not return DataFrames or Series but could return dicts which allow indexing with frozenset. 

EDIT: Now also added an accessor for the results DataFrame and columns:

**Usage examples:**

<details>

```Python3
Python 3.8.5 (default, Aug  2 2020, 15:09:07) 
IPython 7.17.0 -- An enhanced Interactive Python. Type '?' for help.

In [3]: from cobra.flux_analysis import *                                                                                                                                            

In [4]: single = single_gene_deletion(mod)                                                                                                                                           

In [5]: double = double_gene_deletion(mod)                                                                                                                                           

In [6]: single.knockout[mod.genes[0]]                                                                                                                                                
Out[6]: 
        ids    growth   status
64  {b1241}  0.873922  optimal

In [7]: single.knockout[mod.genes[0].id]                                                                                                                                             
Out[7]: 
        ids    growth   status
64  {b1241}  0.873922  optimal

In [8]: single.knockout[mod.genes[0], mod.genes[1]]                                                                                                                                  
Out[8]: 
        ids    growth   status
1   {b0351}  0.873922  optimal
64  {b1241}  0.873922  optimal

In [9]: double.knockout[{mod.genes[0], mod.genes[1]}]                                                                                                                                
Out[9]: 
                 ids    growth   status
9131  {b0351, b1241}  0.873922  optimal

In [10]: double.knockout[{mod.genes[0], mod.genes[1]}, {mod.genes[5], mod.genes[3]}]                                                                                                 
Out[10]: 
                 ids    growth   status
8722  {b3115, b2296}  0.873922  optimal
9131  {b0351, b1241}  0.873922  optimal

In [11]: double.knockout[{mod.genes[0].id, mod.genes[1].id}]                                                                                                                         
Out[11]: 
                 ids    growth   status
9131  {b0351, b1241}  0.873922  optimal
```

</details>

---

* [x] fix #978
* [x] tests added/passed
* [x] update docs with new accessor


Midnighter:
Thanks for the quick reaction @cdiener. I'm indeed also not happy to depend on some order now. I'm wondering if we should create a slim class around the solution that would handle this? I'm happy to pull this in first so that we can move on.

cdiener:
Hm, pandas [discourages subclassing](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures) DataFrame and it would be a lot of code since pandas indexing does a lot of things like slicing, boolean indexing, etc. Maybe we could write an [accessor](https://pandas.pydata.org/docs/development/extending.html#registering-custom-accessors). Then we could do something like:

```Python
df = #gene/reaction deletion
df.reactions["rxn1", "rxn2"]
```

This would then sort uniques and return the indices. This could be added to this PR.

Midnighter:
I didn't know about accessors, looks very promising! I didn't mean to subclass the `DataFrame` directly. I had in mind something similar to our `Solution` class whose attributes are pandas objects. We could then turn access to that deletion result with any iterable into a sorted tuple. Something like:

```py
import pandas as pd


class DeletionResult:
    """"""

    def __init__(self, *, result: pd.Series, **kwargs) -> None:
        """"""
        self._result = result

    def __getitem__(self, args) -> float:
        """"""
        return self._result.loc[tuple(sorted(args))]
```

(Not sure yet if the data frame should be public or private.) This then provides a lot of flexbility how data is accessed, for example,

```py
result[frozenset(["A", "B"])]
result[{"A", "B"}]
result["A", "B"]
```

I think the main benefit is that it would clearly communicate that certain functions return a particular type of result. The disadvantage is that it hides away the data frame a bit making it less intuitive to use the powerful indexing functionality. The latter is maybe where an accessor is better, however, the accessor will be available on any data frame with the right column names...

cdiener:
Hmm, the results things is fine as well I think. Just have to add a `to_frame` and I thinks it would be fine.

Midnighter:
I think a separate class is the better long term solution but quite a breaking change and the design should be considered a little more carefully. In the interest of finding a solution quickly we can either pin pandas to a lower version or create an accessor. Even though an accessor is also a breaking change, I think it is the best medium term solution since we were relying on unsupported functionality and it's bad to prevent pandas upgrades in the meantime.

cdiener:
Okay, I'll add the accessor for now. Any preference on the name (the part after the dot)?

Midnighter:
> Okay, I'll add the accessor for now. Any preference on the name (the part after the dot)?

Since it will be used for both gene and reaction deletion results it should be a bit more general. I was shortly thinking about `.deletion` but it's not a great name either.

I was also wondering if we should make it possible to pass in instances that are then auto-converted to identifiers? Or is that overkill?

cdiener:
> I was also wondering if we should make it possible to pass in instances that are then auto-converted to identifiers? Or is that overkill?

Not really, we can just make it pass through `get_by_any`.

cdiener:
One caveat here: `df.knockout` will only work for a single requested index. Otherwise `df.knockout["rxn1", "rxn2"]` would not be clear; do you want the double deletion for rxn1 and rxn2 or the single deletions fro rxn1 and rxn2?

Midnighter:
> One caveat here: `df.knockout` will only work for a single requested index. Otherwise `df.knockout["rxn1", "rxn2"]` would not be clear; do you want the double deletion for rxn1 and rxn2 or the single deletions fro rxn1 and rxn2?

I think a single index is reasonable. If a user wants something fancier, they will have to construct the correct data frame indexing themselves. What do you think?

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+            A result from one of the deletion methods.
+        """
+        self._validate(pandas_obj)
+        self._result = pandas_obj
+
+    @staticmethod
+    def _validate(obj):
+        # verify it is a deletion results
+        if obj.index.name != "ids":
+            raise AttributeError("Must be DataFrame returned by a deletion method.")
```
Midnighter, line 441: Should we verify the column names as well (`"growth"`, `"status"`)?

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+
+        Returns:
+        --------
+        pd.DataFrame
+            The deletion result where the chosen entities have been deleted.
+        """
+        if not any(isinstance(args, t) for t in [list, tuple, set]):
+            args = [args]
+
+        if isinstance(args[0], Reaction) or isinstance(args[0], Gene):
```
Midnighter: I would not do this type checking here. Since iterables can be of very varied type (at least `frozenset` is an obvious missing one). I would try to access the `id` attribute and recover.

> It's easier to ask forgiveness than it is to get permission.

:wink:

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+        -----------
+        args : iterable of str, cobra.Reaction or cobra.Gene
+
+        Returns:
+        --------
+        pd.DataFrame
+            The deletion result where the chosen entities have been deleted.
+        """
+        if not any(isinstance(args, t) for t in [list, tuple, set]):
+            args = [args]
```
Midnighter: Since `args` is supposed to be iterable, maybe we shouldn't do this?

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+
+        if isinstance(args[0], Reaction) or isinstance(args[0], Gene):
+            args = tuple(sorted(set(obj.id for obj in args)))
```
Midnighter: It'd be weird to pass duplicates when the typical use-case is double knock-outs but I guess you're right to be careful.

```suggestion
            args = tuple(sorted({obj.id for obj in args}))
```

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
         processes=processes,
         **kwargs
     )
+
+
+@pd.api.extensions.register_dataframe_accessor("knockout")
+@pd.api.extensions.register_series_accessor("knockout")
```
Midnighter: Is there a use-case for `Series`? I think it's always a `DataFrame`, isn't it?

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+class KnockoutAccessor:
+    """Access unique combinations of reactions in deletion results."""
+
+    def __init__(self, pandas_obj):
```
Midnighter, line 426:
```suggestion
    def __init__(self, pandas_obj: pd.DataFrame) -> None:
```

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+        Parameters:
+        -----------
+        pandas_obj : pd.DataFrame or pd.Series
+            A result from one of the deletion methods.
+        """
+        self._validate(pandas_obj)
+        self._result = pandas_obj
+
+    @staticmethod
+    def _validate(obj):
```
Midnighter, line 438:
```suggestion
    def _validate(obj: pd.DataFrame) -> None:
```

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+        # verify it is a deletion results
+        if obj.index.name != "ids":
+            raise AttributeError("Must be DataFrame returned by a deletion method.")
+
+    def __getitem__(self, args):
```
Midnighter, line 443:
```suggestion
    def __getitem__(self, args: typing.Iterable[typing.Union[str, Gene, Reaction]]) -> pd.Series:
```

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +405,47 @@ def double_gene_deletion(
...
+            The deletion result where the chosen entities have been deleted.
+        """
+        if not any(isinstance(args, t) for t in [list, tuple, set]):
+            args = [args]
+
+        if isinstance(args[0], Reaction) or isinstance(args[0], Gene):
+            args = tuple(sorted(set(obj.id for obj in args)))
+        else:
+            args = tuple(sorted(set(args)))
+        return self._result.loc[[args]]
```
Midnighter: I don't see why this needs to be packed into a list?

```suggestion
        return self._result.loc[args, :]
```

src/cobra/test/test_flux_analysis/test_deletion.py
```diff
@@ -42,7 +42,7 @@ def test_single_gene_deletion_fba(model, all_solvers):
         model=model, gene_list=list(growth_dict), method="fba", processes=1
     )["growth"]
     for gene, value in iteritems(growth_dict):
-        assert np.isclose(result[frozenset([gene])], value, atol=1e-02)
+        assert np.isclose(result[[(gene,)]], value, atol=1e-02)
```
Midnighter: For all of these I don't understand why they need to be wrapped in a list.

Midnighter (changes requested):
I think the general approach looks promising. I have some doubts around the single element use case and there seems to be something funky with list wrapping that I don't get right now.

src/cobra/flux_analysis/deletion.py
cdiener: In the tests it's always used as:

```
growth = results.growth
...
assert growth[rxn1] ...
```

This can only be handled if you have an accessor on the columns as well. Otherwise there is no simple way to index `results.growth`.

src/cobra/flux_analysis/deletion.py
cdiener: If args is a single element like in `res.knockout["rxn1"]` it won't be. In this case it would iterate over the characters in the string which would be wrong.

```Python3
In [13]: class A: 
    ...:     def __getitem__(self, args): 
    ...:         print(args) 
    ...:                                                                                                                                                                      

In [14]: a = A()                                                                                                                                                              

In [15]: a["string"]                                                                                                                                                          
string

In [16]: a[["string"]]                                                                                                                                                        
['string']
```

src/cobra/flux_analysis/deletion.py
cdiener: But this would allow any object with an id attribute which is too broad, isn't it?

src/cobra/flux_analysis/deletion.py
cdiener: May happen if you want the diagonal elements like `rxn1, rxn1` which get indexed as `(rxn1, )` in the deletion module.

src/cobra/flux_analysis/deletion.py
cdiener: Try it. Doesn't work otherwise. Pandas will treat individual args as multiple indices to look up, not as individual tuple indices. See below.

src/cobra/test/test_flux_analysis/test_deletion.py
cdiener: Will get unpacked automatically by pandas otherwise into a string and index. Deletion module indices are always tuples even when being single string indices. But this is an oversight and should use the accessor.

cdiener:
To illustrate how pandas handles tuple indices:

```Python3
In [1]: import pandas as pd                                                                                                                                                   

In [2]: df = pd.DataFrame({"status": "solved", "growth": 0.8}, index=[("rxn1",)])                                                                                             

In [3]: df  # how deletion modlule returns look                                                                                                                               
Out[3]: 
         status  growth
(rxn1,)  solved     0.8

In [7]: df.loc["rxn1"]                                                                                                                                                        
KeyError                                  Traceback (most recent call last)
~/.local/lib/python3.8/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2888             try:
-> 2889                 return self._engine.get_loc(casted_key)
   2890             except KeyError as err:
...

In [8]: df.loc[("rxn1", )]                                                                                                                                                    
KeyError                                  Traceback (most recent call last)
~/.local/lib/python3.8/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2888             try:
-> 2889                 return self._engine.get_loc(casted_key)
   2890             except KeyError as err:
...

In [9]: df.loc[[("rxn1", )]]                                                                                                                                                  
Out[9]: 
         status  growth
(rxn1,)  solved     0.8

```

cdiener:
Multiple elements are more difficult since you have structures like `[(rxn1, rxn2),..]`, so you have to iterate through the entire tree to map ids on Reaction and Gene objects while still allowing structures such as `[rxn1, rxn2]`.

src/cobra/flux_analysis/deletion.py
cdiener: Why should row indexing on a DataFrame return a Series?

src/cobra/flux_analysis/deletion.py
Midnighter: I think that's okay because within this context I would expect a user to only use cobrapy objects. And if a user attempts to pass metabolite IDs into a gene deletion result it seems okay to me that it will raise a `KeyError` or so.

src/cobra/flux_analysis/deletion.py
Midnighter: I wonder if we need to cover this use-case. If the index consists of simple strings, I don't see the need for the accessor. Then one can simply use the normal pandas index directly.

Midnighter:
Thank you for your examples. I was commenting from memory and should have tried it out.

I was just thinking, since we anyway have to break the behaviour here, how about implementing a multi index instead of tuple indeces?

I was also wondering if it would make it easier to have two accessors: one for data frames and the other for series.

src/cobra/flux_analysis/deletion.py
cdiener: Sure, we could do it for the Data frames at least.

src/cobra/flux_analysis/deletion.py
cdiener: Unfortunately won't work either since deletion results indices are always tuples. See below for the caveats with that.

src/cobra/flux_analysis/deletion.py
Midnighter: My bad, I was going off of `.iterrows` behaviour but it indeed returns a data frame with a single row.

Midnighter: That's weird behaviour, I just played with a MultiIndex and there the returned object is indeed a series.

cdiener:
> Thank you for your examples. I was commenting from memory and should have tried it out.

No prob. I was also surprised by all the conversions pandas does on indices. Indices of tuples are not well handled at all.

> I was just thinking, since we anyway have to break the behavior here, how about implementing a multi index instead of tuple indices?

This would be a bit cleaner but would only work for deletion with the same size (double, triple deletions etc.). However, `multi_gene_deletion` permits arbitrary mixtures like `[("rxn1", ), ("rxn1, "rxn2")]`. Also this would still not solver the ordering problem since we need that `results["rxn1"]["rxn2"] == results["rxn2"]["rxn1"]`. An alternative is to have the deletion elements in a column (which does support sets) and not as an index. Then you could do `results[results.elements == set("rxn1", "rxn2")]` or `results[results.elements.isin(mylist)]`, maybe while having a much simpler accessor now. 

> I was also wondering if it would make it easier to have two accessors: one for data frames and the other for series.

Those would do pretty much the same, wouldn't they? At least indexing would be the same (both use .loc), this is why I put the same accessor for both. You could still have different validations for each.

Midnighter:
> An alternative is to have the deletion elements in a column (which does support sets) and not as an index. Then you could do results[results.elements == set("rxn1", "rxn2")] or results[results.elements.isin(mylist)], maybe while having a much simpler accessor now.

That is a very good argument. I like the idea a lot and it would be a pretty simple change. The only downside is that you can't work with the individual series any longer. I never do that anyway but since we do in the tests I suppose it is not uncommon.

cdiener:
I think all strategies will have downsides, we will just have to decide for one. I do like the column one but I am biased since I come from an R (tidyverse) background where one tries not to keep important information in the data frame indices. Having native indexing with `df[df.ids == {"rxn1", "rxn2"}] ` is pretty nice and does look like decent syntax.

Midnighter:
I'm still partial to the frozenset since it's immutable but otherwise I am also in favour of the column.

cdiener:
Yeah it just becomes a bit uglier since there is no `{}` notation for frozenset. I see very little danger here in using a mutable type. This is what panas does internally anyways.

Midnighter:
> Yeah it just becomes a bit uglier since there is no {} notation for frozenset. I see very little danger here in using a mutable type. This is what panas does internally anyways.

Fair enough. Then let's go with a column and normal sets. :+1:

src/cobra/test/test_flux_analysis/test_deletion.py
```diff
@@ -316,16 +317,15 @@ def test_double_gene_deletion(model):
...
         },
     }
-    solution = double_gene_deletion(model, gene_list1=genes, processes=3)["growth"]
-    solution_one_process = double_gene_deletion(model, gene_list1=genes, processes=1)[
-        "growth"
-    ]
-    for (rxn_a, sub) in iteritems(growth_dict):
+    solution = double_gene_deletion(model, gene_list1=genes, processes=3)
+    solution_one_process = double_gene_deletion(model, gene_list1=genes, processes=1)
+    print(solution)
```
Midnighter:
```suggestion

```

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +401,60 @@ def double_gene_deletion(
...
+        # verify it is a deletion results
+        if any(name not in obj.columns for name in ["ids", "growth", "status"]):
+            raise AttributeError("Must be DataFrame returned by a deletion method.")
+
+    def __getitem__(self, args):
+        """Return the deletion result for a particular set of knocked entities.
+
+        Parameters:
+        -----------
+        args : iterable of str, cobra.Reaction or cobra.Gene
```
Midnighter: Could you extend this description, please? It seems like you're covering many more input scenarios now.

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +401,60 @@ def double_gene_deletion(
...
+
+        Returns:
+        --------
+        pd.DataFrame
+            The deletion result where the chosen entities have been deleted.
+        """
+        if not any(isinstance(args, t) for t in [tuple, list]):
+            args = [args]
+
+        if any(isinstance(args[0], t) for t in [Reaction, Gene, str]):
```
Midnighter, line 467: If I'm not mistaken this will turn `knockout[rxn1, rxn2]` into `[{rxn1}, {rxn2}]`. I would have expected this to yield `[{rxn1, rxn2}]` originally. Maybe we should map out the different input -> output scenarios that we expect.

src/cobra/flux_analysis/deletion.py
```diff
@@ -405,3 +401,60 @@ def double_gene_deletion(
...
+                args = [set(elem.id for elem in obj) for obj in args]
+            except AttributeError:
+                args = [set(obj) for obj in args]
+        else:
+            raise ValueError(
+                "Allowed indices are single Reactions or Genes, "
+                "lists of Reactions of Genes, or lists of sets "
+                "of Reactions or Genes."
+            )
+        found = [x in args for x in self._result.ids]
```
Midnighter, line 484: `args` being a list of sets here, I think this will perform rather poorly. Could you not use `self._result.ids.isin(args)` or does that fail because sets are not hashable?

Midnighter (changes requested):
I'm sorry that I'm only bringing this up now but I notice that the arguments -> results scenarios are unclear to me.

I think it's a good idea to always return a data frame. So
:heavy_check_mark: `knockout[rxn] ->[{rxn}] -> DataFrame shape (1, 3)`
:negative_squared_cross_mark:  `knockout[rxn] -> {rxn} -> Series`

I am confused about the case where `knockout[rxn1, rxn2] -> [{rxn1}, {rxn2}] -> DataFrame shape (2, 3)` because my expectation was `knockout[rxn1, rxn2] -> [{rxn1, rxn2}] -> DataFrame shape (1, 3)` but I see that one can argue either way.

I am wondering if we should cover set input and list of sets input at all since then a user can work with the `ids` column directly.

src/cobra/flux_analysis/deletion.py
Midnighter: I see it doesn't work because they are unhashable.

cdiener:
Can you give an example where the accessor returns a Series right now? I definitely wrote it with the intention to always return a frame.

src/cobra/flux_analysis/deletion.py
cdiener: Yeah, so my motivation was to have consistent indexing behavior for single and double deletions. So that passing two elements to the index always returns two rows. I felt it would be confusing otherwise, but we could do it the other way. Would just be a bit more code since you now have to treat single deletions differently from multi-deletions.

Midnighter: Let's keep it that way then.

Midnighter:
> Can you give an example where the accessor returns a Series right now? I definitely wrote it with the intention to always return a frame.

I don't think it does. And as mentioned above, I think that's a good design decision.

Could you comment on my other point, please?

> I am wondering if we should cover set input and list of sets input at all since for those cases a user can work with the ids column directly.

cdiener:
> I am wondering if we should cover set input and list of sets input at all since for those cases a user can work with the ids column directly

That wouldn't be that easy since you already saw that `isin` does not work with lists of sets. So you have to do:

```Python
my_ids = [{rxn1.id, rxn2.id}, {.., ..}]
found = [id in my_ids for id in results.ids]
results[found]
```

This is now wrapped by `results.knockout[my_ids]` which will also accept Reaction or Gene objects in `my_ids` and will traverse through the list elements and remap them to the string IDs. So I think we kind of need it as long as we think a user may want to access a subset of deletions (for instance all deletions involving `rxn1`). Otherwise we could drop it and allow only single access.

Midnighter:
Hmm, that's true about the ID mapping. I thought we might simplify the accessor by not handling sets but you're right that we should keep it.

Midnighter:
In that case, I think there are only simple changes left and this is good to be merged :slightly_smiling_face:

cdiener:
I very much agree that it can probably be improved in the future. I can't come up with anything better right now but somebody else probably will at some point. I'll address the other changes.

Midnighter (approved):

Midnighter:
Cheers, thank you for all the work :slightly_smiling_face:

PR #981, opened 2020-08-09
refactor: upgrade the configuration module

Midnighter:
* Upgrade the module to be in line with our upgrade guide
* Handle `None` bounds better
* Set the default number of process to max - 1
* Change display of the configuration


cdiener (approved):
LGTM.

PR #982, opened 2020-08-10
refactor: upgrade cobra.util.array to Python 3.6+

synchon:
* [x] refactor `cobra.util.array`
* [x] tests passed

This PR upgrades `cobra.util.array` to meet Python 3.6+ standard. `collections.namedtuple` is replaced with `typing.NamedTuple` (not a breaking change). "DataFrame" argument option for `array_type` is changed to "frame" to simplify option and maintain consistency with other options (breaking change).

### DIFF
--- a/src/cobra/test/test_util/test_array.py
+++ b/src/cobra/test/test_util/test_array.py
@@ -12,7 +12,7 @@ def test_dense_matrix(model):
     assert S.dtype == int
     assert np.allclose(S.max(), [59])
 
-    S_df = create_stoichiometric_matrix(model, array_type="frame", dtype=int)
+    S_df = create_stoichiometric_matrix(model, array_type="DataFrame", dtype=int)
     assert S_df.values.dtype == int
     assert np.all(S_df.columns == [r.id for r in model.reactions])
     assert np.all(S_df.index == [m.id for m in model.metabolites])
--- a/src/cobra/test/test_util/test_array.py
+++ b/src/cobra/test/test_util/test_array.py
@@ -26,7 +26,7 @@ def test_dense_matrix(model):
 
 def test_sparse_matrix(model):
     """Test sparse stoichiometric matrix creation."""
-    _ = pytest.importorskip("scipy")
+    pytest.importorskip("scipy")
     sparse_types = ["dok", "lil"]
 
     solution = model.optimize()
--- a/src/cobra/test/test_util/test_array.py
+++ b/src/cobra/test/test_util/test_array.py
@@ -1,4 +1,4 @@
-"""Test functions of array.py"""
+"""Test functions of array.py."""
 
 import numpy as np
 import pytest
@@ -7,11 +7,12 @@ from cobra.util import create_stoichiometric_matrix
 
 
 def test_dense_matrix(model):
+    """Test dense stoichiometric matrix creation."""
     S = create_stoichiometric_matrix(model, array_type="dense", dtype=int)
     assert S.dtype == int
     assert np.allclose(S.max(), [59])
 
-    S_df = create_stoichiometric_matrix(model, array_type="DataFrame", dtype=int)
+    S_df = create_stoichiometric_matrix(model, array_type="frame", dtype=int)
     assert S_df.values.dtype == int
     assert np.all(S_df.columns == [r.id for r in model.reactions])
     assert np.all(S_df.index == [m.id for m in model.metabolites])
@@ -24,7 +25,8 @@ def test_dense_matrix(model):
 
 
 def test_sparse_matrix(model):
-    scipy = pytest.importorskip("scipy")
+    """Test sparse stoichiometric matrix creation."""
+    _ = pytest.importorskip("scipy")
     sparse_types = ["dok", "lil"]
 
     solution = model.optimize()
--- a/src/cobra/util/array.py
+++ b/src/cobra/util/array.py
@@ -1,4 +1,4 @@
-"""Helper functions for array operations."""
+"""Helper functions for array operations and sampling."""
 
 from typing import TYPE_CHECKING, Any, NamedTuple, Optional, Union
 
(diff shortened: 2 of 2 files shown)


src/cobra/util/array.py
```diff
@@ -1,18 +1,25 @@
...
-from collections import namedtuple
+from typing import TYPE_CHECKING, Any, NamedTuple, Optional, Union
 
 import numpy as np
 import pandas as pd
 
 
+# Used to avoid cyclic reference and enable third-party static type checkers to work
+if TYPE_CHECKING:
+    import cobra
```
Midnighter:
```suggestion
    from cobra import Model
```

src/cobra/util/array.py
```diff
@@ -1,18 +1,25 @@
...
+
 try:
     from scipy.sparse import dok_matrix, lil_matrix
 except ImportError:
     dok_matrix, lil_matrix = None, None
 
 
-def create_stoichiometric_matrix(model, array_type="dense", dtype=None):
+def create_stoichiometric_matrix(
+    model: "cobra.core.Model", array_type: str = "dense", dtype: Optional[Any] = None
```
Midnighter: `dtype` doesn't have a common base type?

```suggestion
    model: "Model", array_type: str = "dense", dtype: Optional[Any] = None
```

src/cobra/util/array.py
```diff
@@ -69,40 +78,45 @@ def create_stoichiometric_matrix(model, array_type="dense", dtype=None):
...
+    If both `atol` and `rtol` are positive, the combined tolerance is the
+    maximum of the two; that is:
+
+    .. math:: \mathtt{tol} = \max(\mathtt{atol}, \mathtt{rtol} * \mathtt{smax})
+
+    Singular values smaller than `tol` are considered to be zero.
+
     """
     A = np.atleast_2d(A)
     u, s, vh = np.linalg.svd(A)
```
Midnighter:
```suggestion
    _, s, vh = np.linalg.svd(A)
```

src/cobra/util/array.py
```diff
@@ -112,67 +126,81 @@ def nullspace(A, atol=1e-13, rtol=0):
     return ns
 
 
-def constraint_matrices(model, array_type="dense", include_vars=False, zero_tol=1e-6):
+def constraint_matrices(
```
Midnighter, line 138: I wonder if this function is still in use anywhere or by anyone? Would be nice to remove it.

src/cobra/test/test_util/test_array.py
```diff
@@ -24,7 +25,8 @@ def test_dense_matrix(model):
 
 
 def test_sparse_matrix(model):
-    scipy = pytest.importorskip("scipy")
+    """Test sparse stoichiometric matrix creation."""
+    _ = pytest.importorskip("scipy")
```
Midnighter: You can just ignore the return value completely.

```suggestion
    pytest.importorskip("scipy")
```

Midnighter (changes requested):
Some minor suggestions in the comments. Two bigger questions:

1. I'm not sure about the API change from `"DataFrame"` to `"frame"`. I see the aesthetic appeal but I don't think it's worth breaking the API for that.
2. Can we remove `constraint_matrices`? Can we investigate this? Any opinions maybe @cdiener @pstjohn ?

src/cobra/util/array.py
synchon: This is only there for non-optlang based use cases. I think it's better to issue a deprecation for now and remove it completely in the major version upgrade.

Midnighter: Yes, that's a good solution if we indeed want that. At the moment I simply don't know if the format is still in use at all.

src/cobra/util/array.py
synchon: Scipy has a `nullspace` [implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.null_space.html) and I was wondering if we should maintain cobrapy's implementation. I don't see any cobrapy-specific thing being done.

src/cobra/util/array.py
synchon: Since we are moving towards a leaner code base, I think it's better if we drop it. Maybe we could keep it in the cookbook?!

src/cobra/util/array.py
synchon: > `dtype` doesn't have a common base type?

To maintain consistency, I think `numpy.dtype` could be used.

synchon:
> 1. I'm not sure about the API change from "DataFrame" to "frame". I see the aesthetic appeal but I don't think it's worth breaking the API for that.

I made the change to remain in sync with the heavy use case of `pandas.DataFrame` in `Summary` object. Do you support issuing a deprecation then?

src/cobra/util/array.py
Midnighter: Let's go with the deprecation warning for now.

src/cobra/util/array.py
Midnighter: If you look at the source code, it's very similar. https://github.com/scipy/scipy/blob/v1.5.2/scipy/linalg/decomp_svd.py#L330-L388 Let's not introduce a scipy dependence here but if you like you can adjust the code of the function to match the scipy implementation.

src/cobra/util/array.py
synchon: Okay.

Midnighter:
> I made the change to remain in sync with the heavy use case of `pandas.DataFrame` in `Summary` object. Do you support issuing a deprecation then?

I'm not sure how the two are related since both can use a data frame and the question is only about the keyword. But a deprecation warning would be fine, yes.

synchon:
I didn't refer to any functionality similarity but I rather wanted to point out that `pandas` prefers to use `frame` instead of `DataFrame` (for example, `pandas.Series.to_frame()`). Okay, I'll issue a deprecation there.

src/cobra/util/array.py
synchon: The functionality doesn't change to a large extent so I guess it's okay to leave it as it is.

src/cobra/util/array.py
cdiener: Those are used in the sampler, please don't remove them. Actually, most of the methods here are required for sampling since that does not use optlang and needs to set up all problem matrices.

synchon: I see. I won't issue a deprecation for the function then.

Midnighter (approved):
Looks all good to me now. We know some tests will fail but I'll merge it anyway if nothing else comes up.

Midnighter:
There was a complaint by black.

synchon:
Should be fixed now.

PR #984, opened 2020-08-12
Improve infrastructure

Midnighter:
* Combine continuous testing and deployment workflows
* Provide multiple issue templates


cdiener (approved):
Looks good. We should probably think about omitting the deprecation warnings from the pytest warnings. It's kind of hard to read.

Midnighter:
> We should probably think about omitting the deprecation warnings from the pytest warnings.

Most of them are actually our own :see_no_evil: We should fix those cases.

PR #987, opened 2020-08-14
refactor: cobra.util.context

synchon:
* [x] tests added/passed

This PR upgrades `cobra.util.context` to meet Python 3.6+ standards.

### DIFF
--- a/src/cobra/test/test_util/test_context.py
+++ b/src/cobra/test/test_util/test_context.py
@@ -1,11 +1,17 @@
 """Test functions of context.py."""
 
+from typing import TYPE_CHECKING
+
 import pytest
 
 from cobra.util import HistoryManager, get_context, resettable
 
 
-def test_history_manager():
+if TYPE_CHECKING:
+    from cobra import Model
+
+
+def test_history_manager() -> None:
     """Test initialization and resetting of HistoryManager."""
     # initialize HistoryManager
     history_manager = HistoryManager()
@@ -15,7 +21,7 @@ def test_history_manager():
     history_manager.reset()
 
 
-def test_get_context(model):
+def test_get_context(model: "Model") -> None:
     """Test if context retrieval is working."""
     with model as model:
         # reverse optimization direcion
@@ -29,7 +35,7 @@ def test_get_context(model):
         context._history[-1]
 
 
-def test_resettable():
+def test_resettable() -> None:
     """Test if resettable decorator is functional."""
     # decorate a dummy function
     @resettable
--- a/src/cobra/test/test_util/test_context.py
+++ b/src/cobra/test/test_util/test_context.py
@@ -17,6 +17,8 @@ def test_history_manager() -> None:
     history_manager = HistoryManager()
     # add non-functioning operation
     history_manager(lambda: 1)
+    # check size of the stack
+    assert history_manager.size() == 1
     # reset operations
     history_manager.reset()
 
@@ -26,13 +28,15 @@ def test_get_context(model: "Model") -> None:
     with model as model:
         # reverse optimization direcion
         model.objective_direction = "min"
+        # retrieve context
         context = get_context(model)
-        # should have history
-        assert context._history[-1].args[-1] == "max"
+        # check size of the context
+        if context:
+            assert context.size() == 1
 
     # there shouldn't be any history
-    with pytest.raises(IndexError):
-        context._history[-1]
+    if context:
+        assert context.size() == 0
 
 
 def test_resettable() -> None:
--- a/src/cobra/test/test_util/test_context.py
+++ b/src/cobra/test/test_util/test_context.py
@@ -1,0 +1,42 @@
+"""Test functions of context.py."""
+
+import pytest
+
+from cobra.util import HistoryManager, get_context, resettable
+
+
+def test_history_manager():
+    """Test initialization and resetting of HistoryManager."""
+    # initialize HistoryManager
+    history_manager = HistoryManager()
+    # add non-functioning operation
+    history_manager(lambda: 1)
+    # reset operations
+    history_manager.reset()
+
+
+def test_get_context(model):
+    """Test if context retrieval is working."""
+    with model as model:
+        # reverse optimization direcion
+        model.objective_direction = "min"
+        context = get_context(model)
+        # should have history
+        assert context._history[-1].args[-1] == "max"
+
+    # there shouldn't be any history
+    with pytest.raises(IndexError):
+        context._history[-1]
+
+
+def test_resettable():
+    """Test if resettable decorator is functional."""
+    # decorate a dummy function
+    @resettable
+    def change_my_name(old_name, new_name):
+        """Change old name to new name."""
+        if old_name != new_name:
+            old_name = new_name
+
+    # call the dummy function
+    change_my_name("hmm", "hmmm")
--- a/src/cobra/util/context.py
+++ b/src/cobra/util/context.py
@@ -90,6 +90,7 @@ def resettable(f: Callable[[Any], Any]):
         The decorated function.
 
     """
+
     def wrapper(self, new_value):
         context = get_context(self)
         if context:
--- a/src/cobra/util/context.py
+++ b/src/cobra/util/context.py
@@ -1,46 +1,68 @@
 """Context manager for the package."""
 
 from functools import partial
+from typing import Any, Callable
 
 
 class HistoryManager:
-    """Record a list of actions to be taken at a later time. Used to
-    implement context managers that allow temporary changes to a
-    :class:`~cobra.core.Model`.
+    """
+    Define a base context manager.
+
+    It records a list of actions to be taken at a later time.
+    This is used to implement context managers that allow temporary
+    changes to a `cobra.core.Model`.
 
     """
 
     def __init__(self):
-
-        # self._history just acts as a stack
+        """Initialize the class."""
+        # this acts like a stack
         self._history = []
 
-    def __call__(self, operation):
-        """Add the corresponding method to the history stack.
+    def __call__(self, operation: Callable[[Any], Any]):
+        """Add the corresponding operation to the history stack.
 
         Parameters
         ----------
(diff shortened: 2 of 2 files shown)


src/cobra/test/test_util/test_context.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_get_context(model):
+    """Test if context retrieval is working."""
+    with model as model:
+        # reverse optimization direcion
+        model.objective_direction = "min"
+        context = get_context(model)
+        # should have history
+        assert context._history[-1].args[-1] == "max"
```
Midnighter: This seems brittle to me. We should not access private class members. You could rather assert that before and after the context, the `objective_direction` is `"max"`.

src/cobra/test/test_util/test_context.py
```diff
@@ -0,0 +1,42 @@
...
+    """Test initialization and resetting of HistoryManager."""
+    # initialize HistoryManager
+    history_manager = HistoryManager()
+    # add non-functioning operation
+    history_manager(lambda: 1)
+    # reset operations
+    history_manager.reset()
+
+
+def test_get_context(model):
```
Midnighter:
```suggestion
def test_get_context(model: cobra.Model) -> None:
```

src/cobra/test/test_util/test_context.py
```diff
@@ -0,0 +1,42 @@
+"""Test functions of context.py."""
+
+import pytest
+
+from cobra.util import HistoryManager, get_context, resettable
+
+
+def test_history_manager():
```
Midnighter:
```suggestion
def test_history_manager() -> None:
```

src/cobra/test/test_util/test_context.py
```diff
@@ -0,0 +1,42 @@
...
+        context = get_context(model)
+        # should have history
+        assert context._history[-1].args[-1] == "max"
+
+    # there shouldn't be any history
+    with pytest.raises(IndexError):
+        context._history[-1]
+
+
+def test_resettable():
```
Midnighter:
```suggestion
def test_resettable() -> None:
```

src/cobra/test/test_util/test_context.py
```diff
@@ -0,0 +1,42 @@
...
+    """Test if resettable decorator is functional."""
+    # decorate a dummy function
+    @resettable
+    def change_my_name(old_name, new_name):
+        """Change old name to new name."""
+        if old_name != new_name:
+            old_name = new_name
+
+    # call the dummy function
+    change_my_name("hmm", "hmmm")
```
Midnighter, line 52: It seems like some assertion should be made here?

src/cobra/util/context.py
```diff
@@ -1,46 +1,68 @@
...
+    """
+    Define a base context manager.
+
+    It records a list of actions to be taken at a later time.
+    This is used to implement context managers that allow temporary
+    changes to a `cobra.core.Model`.
 
     """
 
     def __init__(self):
```
Midnighter:
```suggestion
    def __init__(self) -> None:
```

src/cobra/util/context.py
```diff
@@ -1,46 +1,68 @@
...
-
-        # self._history just acts as a stack
+        """Initialize the class."""
+        # this acts like a stack
         self._history = []
 
-    def __call__(self, operation):
-        """Add the corresponding method to the history stack.
+    def __call__(self, operation: Callable[[Any], Any]):
```
Midnighter:
```suggestion
    def __call__(self, operation: Callable[[Any], Any]) -> None:
```

src/cobra/util/context.py
```diff
@@ -1,46 +1,68 @@
...
-        operation : `function`
-            A function to be called at a later time
+        operation : callable
+            A function to be called at a later time.
 
         """
-
         self._history.append(operation)
 
     def reset(self):
```
Midnighter:
```suggestion
    def reset(self) -> None:
```

src/cobra/util/context.py
```diff
@@ -1,46 +1,68 @@
...
+    It records a list of actions to be taken at a later time.
+    This is used to implement context managers that allow temporary
+    changes to a `cobra.core.Model`.
 
     """
 
     def __init__(self):
-
-        # self._history just acts as a stack
+        """Initialize the class."""
```
Midnighter, line 22: In principle, `__init__` should always accept `**kwargs` and call `super().__init__(**kwargs)` at least. That is to enable multiple inheritance.

src/cobra/util/context.py
```diff
@@ -49,10 +71,24 @@ def get_context(obj):
     return None
```
Midnighter: Superfluous statement.

```suggestion
```

src/cobra/util/context.py
```diff
@@ -1,46 +1,68 @@
...
 
     def reset(self):
-        """Trigger executions for all items in the stack in reverse order"""
+        """Trigger executions for all items in the stack in reverse order."""
         while self._history:
             entry = self._history.pop()
             entry()
 
 
 def get_context(obj):
```
Midnighter: I'm not sure about the context type right now.

```suggestion
def get_context(obj) -> Optional[Any]:
```

src/cobra/util/context.py
```diff
@@ -49,10 +71,24 @@ def get_context(obj):
     return None
 
 
-def resettable(f):
-    """A decorator to simplify the context management of simple object
-    attributes. Gets the value of the attribute prior to setting it, and stores
-    a function to set the value to the old value in the HistoryManager.
+def resettable(f: Callable[[Any], Any]):
```
Midnighter: `f` is a terrible argument name.

```suggestion
def resettable(f: Callable[[Any], Any]) -> Callable[[Any], Any]:
```

Midnighter (changes requested):
Mostly, return type annotations were missing. See comments.

src/cobra/test/test_util/test_context.py
synchon: I think using `"Model"` with `TYPE_CHECKING` check is a bit cleaner.

src/cobra/test/test_util/test_context.py
synchon: I am not sure what assertion to use here. The idea is that the function should not throw any errors during runtime. The way it's implemented requires the method to be already a part of `cobra.core.Model` class (if the test should be related to the `Model`'s functionality). And, so I just made sure it works and thus the test function doesn't return anything useful. Any suggestion though?

src/cobra/test/test_util/test_context.py
synchon: I understand that accessing private class members, is not a good practice. But, if we just check the values before and after the context, we are not directly checking the functionality of the `get_context()` function but rather indirectly verifying whether the thing works as a whole. Will that indirect test be enough?

src/cobra/test/test_util/test_context.py
Midnighter: I don't really. Sorry.

synchon: Can we let it be for now and revisit later when we have some better idea?

src/cobra/test/test_util/test_context.py
Midnighter: The problem I see is that the test relies on a lot of class internal specifics which it shouldn't. You could implement a `__len__` method that returns the number of commands on the history stack and test that, for example?

src/cobra/util/context.py
synchon: Yes. Totally missed it.

src/cobra/util/context.py
synchon: Okay agreed.

src/cobra/util/context.py
synchon: Should be `HistoryManager`.

src/cobra/util/context.py
synchon: Will change it.

src/cobra/test/test_util/test_context.py
synchon: That is a better way. I'll implement that.

synchon: Done.

synchon:
@Midnighter I have addressed your comments. Does it look okay now?

src/cobra/test/test_util/test_context.py
```diff
@@ -0,0 +1,52 @@
...
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+from cobra.util import HistoryManager, get_context, resettable
+
+
+if TYPE_CHECKING:
+    from cobra import Model
```
Midnighter, line 11: Just as an informative comment. I think it would be okay to directly import in the tests. Once they're moved out of the source tree, there cannot be any circular references.

src/cobra/util/context.py
```diff
@@ -1,69 +1,112 @@
...
+    Parameters
+    ----------
+    obj: cobra.Object
+        The cobra.Object for which to search context manager.
 
-def get_context(obj):
-    """Search for a context manager"""
+    Returns
+    -------
+    None
```
Midnighter: Returns a `HistoryManager` instance ideally, right?

Midnighter (approved):
One small comment about the docstring otherwise it looks great to me. Thanks :slightly_smiling_face:

src/cobra/util/context.py
synchon: Yes. My bad. Forgot to update the docstring.

src/cobra/test/test_util/test_context.py
synchon: We will need to revamp the test suite as a whole when we move it out. I think we can handle it then.

src/cobra/util/context.py
synchon: Done.

synchon:
I have updated the docstring.

Issue #954, opened 2020-05-07
Discussion for Group feature in COBRApy


Hemant27031999:
Hi all, @draeger, @matthiaskoenig, @Midnighter, @cdiener, and other members of COBRApy. I am opening a new thread for discussing about the implementation of Group package in COBRApy as I could not find one best suited for this discussion.

Complete support for Group package has already been added in COBRApy to parse the group info to cobra models. Even if the information is present in the form of subsystems inside the notes of a reaction, that is also extracted and converted to the group structure of cobra models. However, the group information is currently not put in JSON format if cobra models are converted to JSON form, and hence this information is lost. One of the tasks of the [project #137](https://github.com/nrnb/GoogleSummerOfCode/issues/137) is to add complete support of Group feature in SBML-JSON form. I went through the Group specification document to understand the group structure and here is the format I propose for it:

```
{
    '+++',
    "listOfGroups" : {
        "metaid": "value",
        "sboterm": "value",
        "notes": "value",
        "annotation": {"content"},
        "groups": [
            {
                "id": "value",
                "name": "value",
                "notes": "value",
                "annotation": {"content"},
                "kind": "value",
                "listOfMembers": {
                    "metaid": "value",
                    "sboterm": "value",
                    "notes": "value",
                    "annotation": {"content"},
                    "members": [
                        {
                            "id": "value",
                            "name": "value",
                            "notes": "value",
                            "annotation": {"content"},
                            "idRef": "value",
                            "metaIdRef": "value"
                        },
                        '...'
                    ]
                }
            },
            '...'
        ],
    },
    '+++'
}
```
Some key-value pairs shown above are optional and may not be present like those which came due to extension from SBase class and are of type optional. Also, only one out of idRef and metaIdRef attribute is to be used inside the Member class. All these things will be handled inside the method that will do the parsing.

One thing I want to point here is that all the classes under group package i.e., ListOfGroups, Group, ListOfMembers, and Member extends the SBase class and hence can possibly have the attributes of SBase class also like notes and annotations. However, in the current implementation of COBRApy, only the notes and annotations corresponding to Group class are parsed to the cobra model. No parsing of notes and annotations corresponding to ListOfGroup, ListOfMembers, and Member class is done. So theoretically speaking, I think the parsing of notes and annotations of these classes should also be implemented. The above JSON format support this, but this information first has to be parsed to cobra models. Members, can you please provide your review regarding this?

cdiener:
[...]

Issue #937, opened 2020-02-23
Loss of relation elements in annotation when writing a SBML model again.


Hemant27031999:
## Problem Description
If we read a SBML model, from the corresponding XML file and write it back, then, if the annotation field of some component had more than one relation elements like "is" and "hasTaxon", they are lost and only one relation element exists in the newly written model, with all resources listed under it.

## Code Sample

I took the e_coli_core.xml model present inside the data directory, read it using 

> model = cobra.io.read_sbml_model("e_coli_core.xml")

and when I wrote it back using,

> cobra.io.write_sbml_model(model, "e_coli_model.xml")

the annotation format changed as follows :

**Annotation in the actual model :** 

```
...
<annotation>
      <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:bqmodel="http://biomodels.net/model-qualifiers/" xmlns:bqbiol="http://biomodels.net/biology-qualifiers/">
        <rdf:Description rdf:about="#e_coli_core">
          <bqbiol:hasTaxon>
            <rdf:Bag>
              <rdf:li rdf:resource="http://identifiers.org/taxonomy/511145" />
            </rdf:Bag>
          </bqbiol:hasTaxon>
          <bqmodel:is>
            <rdf:Bag>
              <rdf:li rdf:resource="http://identifiers.org/bigg.model/e_coli_core" />
            </rdf:Bag>
          </bqmodel:is>
          <bqmodel:isDescribedBy>
            <rdf:Bag>
              <rdf:li rdf:resource="http://identifiers.org/doi/10.1128/ecosalplus.10.2.1" />
            </rdf:Bag>
          </bqmodel:isDescribedBy>
        </rdf:Description>
      </rdf:RDF>
    </annotation>
...
```

**Annotation after writing the model :** 

```
...
<annotation>
      <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:vCard="http://www.w3.org/2001/vcard-rdf/3.0#" xmlns:vCard4="http://www.w3.org/2006/vcard/ns#" xmlns:bqbiol="http://biomodels.net/biology-qualifiers/" xmlns:bqmodel="http://biomodels.net/model-qualifiers/">
        <rdf:Description rdf:about="#meta_e_coli_core">
          <bqbiol:is>
            <rdf:Bag>
[...]

Issue #695, opened 2018-04-17
Import/export of SBML <notes>


bdelepine:
#### Problem description

Hi all,

Do you intend to support import/export of `<notes>` from/to SBML files?

When a SBML model is parsed with some `<notes>` under, for instance, a `<geneProduct>`, those notes are not imported. I would have expected to retrieve the notes with `mymodel.genes.mygene.notes`, but the attribute is always empty.

I didn't saw in `cobra.io.sbml3.parse_xml_into_model` and `cobra.io.sbml3.annotate_cobra_from_sbml` where the `<notes>` are imported. I believe they are not; which would explain why they are not available in the python object.

#### Code Sample

Gist to [mini_fbc2_notes.xml](https://gist.github.com/bdelepine/837395098aacf91287035367ab44ad39)

```python
import cobra

m = cobra.io.sbml3.read_sbml_model("mini_fbc2_notes.xml")

print([x.notes for x in m.genes])
# [{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]

print([x.notes for x in m.reactions])
# [{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]

print([x.notes for x in m.metabolites])
# [{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

#### Expected Output

```python
print([x.notes for x in m.genes])
# [{"This is an important note for humans."}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]

print([x.notes for x in m.reactions])
# [{"This is an important note for humans."}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]

print([x.notes for x in m.metabolites])
# [{"This is an important note for humans."}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

#### Output of `cobra.show_versions()`

<details>

System Information
==================
[...]

PR #988, opened 2020-08-15
Metadata fbc3 group

Hemant27031999:
## Description of Features

1. **Metadata Classes:** A separate directory for handling metadata information is made inside `cobra/core` directory. Every object derived from SBase can have meta information like annotation (CVTerms), notes, history attached to it. 

- The `Notes` class is holding a simple string containing notes data (XHTML string) and a dictionary, synchronized with the notes string, storing key-value pair of the form `<p> key : value </p>` present in the notes string. One can only modify this key-value pair data, he can't add new key-value pairs inside notes because notes are not a right place to store these key-value pairs.
- The `CVTerms` class for storing externally linked resources to each component derived from SBase. This class is maintaining the new format annotation as well as old format annotation simultaneously, and both are kept synchronized with each other. Changing one will modify the other accordingly. This new class for annotation can handle any type of annotation data (be it the case of nested annotation or alternative annotation). It can read the old format as well as the new format annotation data from JSON and other formats. At the time of writing back the model, the new data format is used because it contains the complete data organized in the same way as SBML.
- The `History` class used for storing the history, validating dates, etc, is now attached to each component derived from SBase.
- The `KeyValuePair` class for storing key-value pairs, defined by fbc-v3.

The last three metadata objects (i.e CVTerms, History, KeyValuePair) are present inside a single attribute of SBase (Object) class and can be accessed via `object.annotation.cvterms`, `object.annotation.history` and `object.annotation.key_value_data` attributes. Calling simply the annotation attribute (`object.annotation`) will return the annotation data in old format (making it backward compatible).

2. **Group to JSON:** The support of the group package is extended to JSON.

3. **JSON schema v2:** The version2 of JSON schema has been added which defines the new format annotation, history, key-value pair, notes, group package data, user-defined constraints data and basic SBML info.

## Issues Fixed

* [x] fixes issue #954: The support of group package has been extended to JSON and other formats.
* [x] fixes issue #856: The infinity values are also enabled for storing bounds.
* [ ] fixes issue #810: The `sbml_info` storing basic information of SBML is written to JSON to store the basic SBML document information like packages, level, version, notes, annotation attached to the SBML component etc. 
* [x] fixes issue #736: If annotation is present in the form of list of list, it is first modified and then data is read.
* [x] fixes issue #706: Single annotation resource are now put into a string. While reading from JSON also, if a single resource attribute is in the form of string, it is first fixed to list and then read into the model.
* [x] fixes issue #695: The notes data is now read into a string and this string is synchronized with a dictionary storing key-value pairs present inside notes string.
* [x] fixes issue #684: The complete metadata structure (CVTerms, Notes, History) has been redesigned with backward compatibility. Old format data is read, fixed and then written in new format metadata structures. The fbc-v3 "KeyValuePair" class is also a part of this new format annotation. Its corresponding class and its support to JSON and other format has been added. However, the SBML parser for it has to be updated when libsbml adds support for fbc-v3.
* [x] fixes issue #937: The annotation format has been updated, which is backward compatible too.
* [x] JSON schema v1: The JSON schema version 1 is modified to resolve some pre-existing issues.

## Tests 


Hemant27031999:
@matthiaskoenig please review.

Midnighter:
@Hemant27031999 I am really glad for all the work that you're putting into cobrapy. Some of the changes that you propose are quite far reaching. I assume that this work is based on discussions with @matthiaskoenig. In future, it would be good to start with a discussion in an issue. Both for feedback and for avoiding to work on features that are not desired by the community (I'm not saying this is the case here).

Will take some time to review. In the meantime, could you upgrade your work by merging in or rebasing on the latest devel branch, please? Possibly, after https://github.com/opencobra/cobrapy/pull/980 was merged so that the full test suite passes.

Hemant27031999:
> @Hemant27031999 I am really glad for all the work that you're putting into cobrapy.

Hey @Midnighter , thank you very much.



> I assume that this work is based on discussions with @matthiaskoenig.

Yes, all the features implemented here have been discussed with @matthiaskoenig and @draeger. Matthias will also go through the code once, maybe tomorrow or so.


> In the meantime, could you upgrade your work by merging in or rebasing on the latest devel branch, please?

Sure, I will fix it soon. For any type of required changes/documentation, just let me know.

cdiener:
Thanks for that. There is a lot of great stuff in here. I was confused about the user-defined constraints though. The code basically reinvents optlang Constraints. Also it seems like adding constraints via `model.add_cons_vars` would not be picked up by that. I think we should have only one way to add constraints and that functionality already exists. SBML does not allow all constraints we do (like constraints on only forward fluxes and custom variables). So I think the only thing you need is 

1. A function converting an SBML constraint to an optlang constraint.
2. A filter function that can tell you if any constraint in a model is (a) already managed via the stoichiometric matrix, and (b) is supported by SBML, meaning it involves always forward and reverse flux and those two coefficients sum to something close to zero and that it contains no other variables
3. A function that can convert a valid optlang constraint to an SBML node.

2 and 3 may be combined for efficiency.

Those 2-3 functions could live in the IO module since they are only relevant for importing and exporting. Everything else can then be managed by the existing API.

src/cobra/io/dict.py
```diff
@@ -139,36 +143,38 @@ def _annotation_to_dict(annotation: MetaData) -> Dict:
...
     return final_dict
 
 
 def _extract_annotation(data: Dict) -> MetaData:
-        cvterms = data["cvterms"] if "cvterms" in data else None
-        history = data["history"] if "history" in data else None
-        keyValueDict = data["history"] if "keyValueDict" in data else None
+    cvterms = data["cvterms"] if "cvterms" in data else None
+    history = data["history"] if "history" in data else None
+    keyValueDict = data["history"] if "key_value_data" in data else None
```
draeger: Would it be possible to define `String` literals as constants? Using such literal expressions repeatedly belongs to a possible source of error (because of typos) that can be relatively easily eliminated.

src/cobra/io/dict.py
```diff
@@ -139,36 +143,38 @@ def _annotation_to_dict(annotation: MetaData) -> Dict:
...
-        if "sbo" in data:
-            annotation["sbo"] = [data["sbo"]]
-        return annotation
+    if cvterms is not None or history is not None or keyValueDict is not None:
+        annotation = MetaData(cvterms, history, keyValueDict)
+    else:
+        annotation = MetaData()
+        annotation.cvterms.add_simple_annotations(data)
+    if "sbo" in data:
+        annotation["sbo"] = [data["sbo"]]
```
draeger: Here also: Please avoid using `String` literals repeatedly.

src/cobra/io/dict.py
```diff
@@ -180,7 +186,7 @@ def _update_optional(cobra_object, new_dict, optional_attribute_dict,
 def metabolite_to_dict(metabolite):
     new_met = OrderedDict()
     for key in _REQUIRED_METABOLITE_ATTRIBUTES:
-        if key == 'id':
+        if key == "id":
```
draeger, line 121: Again... Define as a `String` constant if possible.

src/cobra/io/dict.py
```diff
@@ -384,15 +401,18 @@ def group_from_dict(group: Dict, model: Model) -> Group:
...
                 elif member["type"] == "Metabolite":
                     cobra_obj = model.metabolites.get_by_id(
-                        F_REPLACE["F_SPECIE"](member["idRef"]))
+                        F_REPLACE["F_SPECIE"](member["idRef"])
+                    )
                     cobra_members.append(cobra_obj)
                 elif member["type"] == "Gene":
                     cobra_obj = model.genes.get_by_id(
-                        F_REPLACE["F_GENE"](member["idRef"]))
+                        F_REPLACE["F_GENE"](member["idRef"])
```
draeger, line 297: I just notice that there are many `String` literals all over the source code. I will not add further remarks about it. If possible, please go through the source code, identify `String` literals and replace them with constant expressions (constant variables).

src/cobra/core/user_defined_constraints.py
```diff
@@ -295,7 +295,7 @@ def constraint_from_expression(id=None, expression: 'str' = '',
         tree = ast.parse(source=expression, mode='eval')
         compute_nodes = UserDefinedConstraint.ComputeNumericNodes()
         tree = compute_nodes.visit(tree)
-        print(ast.dump(tree))
+        print((ast.dump(tree)))
```
draeger: Why are additional parentheses needed?

src/cobra/core/user_defined_constraints.py
```diff
@@ -44,6 +61,186 @@ def __init__(self, id=None, name=None, lower_bound: [int, float]=None,
...
+                    result = Num(n=node.left.n - node.right.n)
+                    return copy_location(result, node)
+                elif isinstance(node.op, Mult):
+                    result = Num(n=node.left.n * node.right.n)
+                    return copy_location(result, node)
+                elif isinstance(node.op, Div):
+                    result = Num(n=node.left.n / node.right.n)
+                    return copy_location(result, node)
+                elif isinstance(node.op, Mod):
+                    result = Num(n=node.left.n % node.right.n)
```
draeger: Please add blanks between variables and operators, e.g., change `n=xyz` to `n = xyz`.

src/cobra/core/formula.py
```diff
@@ -82,132 +81,135 @@ def parse_composition(self):
...
+    "Db": 262.000000,
+    "Sg": 266.000000,
+    "Bh": 264.000000,
+    "Hs": 277.000000,
+    "Mt": 268.000000,
+    "Ds": 281.000000,
+    "Rg": 272.000000,
+    "Cn": 285.000000,
+    "Uuq": 289.000000,
+    "Uuh": 292.000000,
```
draeger: Please note that elements naturally occur in different isotopes. Please indicate if these values are the average or the most likely molecular weight in form of a user-readable description. Better could be to use a dictionary whose values are lists of values for the different isotopes.

cobra/core/annotation.py
```diff
@@ -0,0 +1,159 @@
...
+import collections
+from cobra.core.object import Object
+
+QUALIFIER_TYPES = (
+    "is", "hasPart", "isPartOf", "isVersionOf", "hasVersion",
+    "isHomologTo", "isDescribedBy", "isEncodedBy", "encodes",
+    "occursIn", "hasProperty", "isPropertyOf", "hasTaxon",
+    "unknown", "bqm_is", "bqm_isDescribedBy", "bqm_isDerivedFrom",
+    "bqm_isInstanceOf", "bqm_hasInstance", "bqm_unknown",
+)
```
draeger: Please add a documentation comment here where to find the most recent list of qualifiers to ease later updats.

Hemant27031999:
Hey @cdiener, sorry for late reply.

The UserDefinedConstraint (and other fbc-v3) feature is not currently fully implemented because libsbml has not added support for it yet, and SBML parser for it can't be made without libsbml support. We requested them for this feature before starting of the project, and they have opened a tracker item for this, but it is still under progress.

https://www.pivotaltracker.com/n/projects/248655/stories/172219731

So we implemented the classes for making UserDefinedConstraint objects and a helper function to add constraints via a simple string expression.

> Also it seems like adding constraints via `model.add_cons_vars` would not be picked up by that.

Yeah, you are right. We forgot about it. The `model.add_cons_vars` is not currently synchronized with the UserDefinedConstraint classes, and we need this to make sure every constraint added to the model via this method must also be converted to a UserDefinedConstraint object, and this information should be stored in the model. 

> A filter function that can tell you if any constraint in a model is (a) already managed via the stoichiometric matrix, and (b) is supported by SBML, meaning it involves always forward and reverse flux and those two coefficients sum to something close to zero and that it contains no other variables

I currently don't have much knowledge of the solvers of COBRApy. Can we somehow get to know if a same constraint is already present in the model? And is there some way of knowing if a constraint is valid SBML constraint or not?

cdiener:
> I currently don't have much knowledge of the solvers of COBRApy. Can we somehow get to know if a same constraint is already present in the model?

Sure that would not be too hard. So the only constraints that don't need to be added are the equality constraints from the stoichiometric matrix. Those can be obtained from 

```
S_constraints = [met.constraint for met in model.metabolites]
```
> And is there some way of knowing if a constraint is valid SBML constraint or not?

All other constraints have been added manually by the user. To check if it is valid is harder. You would start by making a dictionary that maps each variable to the reaction (see the sampling classes for examples). Then you get the the components of the constraint with `constraint.get_linear_coefficients(constraint.variables)`. Finally you check that (a) all variables match to reactions, (b) each reaction is hit exactly twice, and (c) that the coefficients add up to something smaller than `model.tolerance`. Only if a,b, and c are true is the constraint supported and can be passed to SBML. Since the side effect of that check is parsing the involved reactions and coefficient it might be smart to return those as well.

matthiaskoenig:
@Hemant27031999 I went over most of the code and cleaned it up. Was much more work then expected but most of the metadata is now pretty clean and neat. I also updated to the latest develop code.

Could you
- [ ] add the `cvterms.to_dict` method which will handle the serialization of the cvterms. I added the similar methods on the other metadata objects. You can currently see in the two failing tests where this is still missing
- [ ] can you update the `metadata.ipynb I started for the the documentation of the metadata features. Here you should add the common use cases, i.e., adding CVTerms, adding ModelHistory, modifying notes, storying information as Key-Value pairs, ...

Hemant27031999:
@matthiaskoenig I have made the requested changes. All tests are passing now.

Hemant27031999:
Hey people, metadata and UserDefinedConstraints are two separate functionalities. So I have removed UserDefinedConstraint from this branch, it now majorly includes metadata and some JSON serialization. I shall link another PR for Constraint class.

draeger (changes requested):

src/cobra/core/user_defined_constraints.py
Hemant27031999: The python code format tool "black" doesn't allow spaces between the variable and operator when passed as arguments.

src/cobra/core/formula.py
Hemant27031999: Members, please provide review. I haven't updated it. It's been reformatted by "black" itself by adding double quotes in place of single quotes.

src/cobra/core/user_defined_constraints.py
Hemant27031999: I am not sure if I added them or the code formatter. By the way, I will update it. This file is actually a part of other the PR now. It's no longer on this branch.

Issue #900, opened 2019-08-21
model.summary() is inconsistent with and without 'with' context


bdelepine:
#### Problem description

Hello,

I want to use a context to safely change the C source, compute an FBA and then display the summary. I was surprised to see that the changes introduced in the context were ignored by the summary. I think this unexpected behavior because 1/ this is different than the behavior without using context, 2/ this leads to meaningless values if we additionally ask for FVA in the summary.

++

#### Code Sample

```python
import cobra
import cobra.test

# With context
model = cobra.test.create_test_model("ecoli")
with model as m:
    m.reactions.EX_glc__D_e.bounds = (0, 1000)
    m.reactions.EX_ac_e.bounds = (-10, 1000)
    solution = m.optimize()
    ans1 = m.summary()
    ans2 = m.summary(solution=solution)
    ans3 = m.summary(solution=solution, fva=0.95)
    
# Without context
m = cobra.test.create_test_model("ecoli")
m.reactions.EX_glc__D_e.bounds = (0, 1000)
m.reactions.EX_ac_e.bounds = (-10, 1000)
solution = m.optimize()
ans11 = m.summary()
ans22 = m.summary(solution=solution)
ans33 = m.summary(solution=solution, fva=0.95)

#
print(ans1, ans11)
print(ans2, ans22)
print(ans3, ans33)

```

#### Actual Output

ans1 and ans11 are different:
```
IN_FLUXES IN_FLUXES OUT_FLUXES OUT_FLUXES            OBJECTIVES           OBJECTIVES
    ID       FLUX       ID        FLUX                   ID                  FLUX   
     o2_e    19.8      h2o_e      50.1     BIOMASS_Ec_iJO1366_core_53p95M   0.982   
    nh4_e    10.6      co2_e      19.7                                        nan   
 glc__D_e      10      fe3_e      9.03                                        nan   
    fe2_e    9.04                  nan                                        nan   
     pi_e   0.948                  nan                                        nan    
 IN_FLUXES IN_FLUXES OUT_FLUXES OUT_FLUXES            OBJECTIVES           OBJECTIVES
    ID       FLUX       ID        FLUX                   ID                  FLUX   
    ac_e       10      h2o_e      16.4     BIOMASS_Ec_iJO1366_core_53p95M   0.247   
    o2_e     9.33      co2_e      9.85                                        nan   
     h_e     7.73                  nan                                        nan   
   nh4_e     2.67                  nan                                        nan   
[...]

PR #990, opened 2020-08-19
feature: element summary

Midnighter:
* Make the context test for the model summary closer to what was reported in #900 
* Add the ability to display atomic element uptake/secretion to the model summary. See https://gist.github.com/Midnighter/84590cbc5d46209897abcf6ec2ba812d for a preview

### DIFF
--- a/src/cobra/summary/model_summary.py
+++ b/src/cobra/summary/model_summary.py
@@ -108,6 +108,8 @@ class ModelSummary(Summary):
             optimum objective to be searched.
 
         """
+        super()._generate(model=model, solution=solution, fva=fva)
+
         coefficients = linear_reaction_coefficients(model)
         if solution is None:
             logger.info("Generating new parsimonious flux distribution.")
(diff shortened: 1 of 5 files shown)


src/cobra/summary/model_summary.py
```diff
@@ -396,7 +419,11 @@ def to_string(
         )
 
     def to_html(
-        self, names: bool = False, threshold: float = 1e-6, float_format: str = ".4G"
+        self,
+        names: bool = False,
+        element: str = "C",
+        threshold: float = 1e-6,
```
cdiener: I wonder if we should accept a `None` argument for threshold that initializes it to the model tolerance. This threshold would make little sense if the model tolerance is 1e-3 for instance. Can be left for a future PR though.

cdiener (commented):
Looks good. I love having usage previews in PRs. Kudos!

cdiener (approved):

src/cobra/summary/model_summary.py
Midnighter: I would have loved to do that but I wanted to structure the summary in a way that at the time of display, it does not access the model any longer. However, I could copy the model tolerance on initialization and store that value for later.

Midnighter: Okay, I have implemented something like what you asked for.

PR #993, opened 2020-08-25
refactor: cobra.util.solver

synchon:
* [x] tests added/passed

This PR upgrades `cobra.util.solver` to meet Python 3.6+ standards.

### DIFF
--- a/src/cobra/test/test_util/test_solver.py
+++ b/src/cobra/test/test_util/test_solver.py
@@ -54,23 +54,18 @@ def test_linear_reaction_coefficients(model: "Model") -> None:
     assert coefficients == {model.reactions.Biomass_Ecoli_core: 1}
 
 
-@pytest.mark.parametrize("solver", optlang_solvers)
-def test_fail_non_linear_reaction_coefficients(
-    model: "Model", solver: List[str]
-) -> Optional[ValueError]:
+def test_fail_non_linear_reaction_coefficients(model: "Model") -> None:
     """Test failure of non-linear coefficient identification in reaction."""
-    model.solver = solver
-    try:
+    model.solver = "optlang-glpk"
+
+    with pytest.raises(ValueError) as error:
         model.objective = model.problem.Objective(
             model.reactions.ATPM.flux_expression ** 2
         )
-    except ValueError:
-        pass
-    else:
         coefficients = su.linear_reaction_coefficients(model)
         assert coefficients == {}
-        with pytest.raises(ValueError):
-            model.reactions.ACALD.objective_coefficient = 1
+
+    assert "GLPK only supports linear objectives." in str(error.value)
 
 
 def test_add_remove(model: "Model") -> None:
--- a/src/cobra/test/test_util/test_solver.py
+++ b/src/cobra/test/test_util/test_solver.py
@@ -18,25 +18,25 @@ optlang_solvers = [f"optlang-{s}" for s in stable_optlang if s in su.solvers]
 
 
 def test_solver_list() -> None:
-    """Test if solvers are found."""
+    """Expect that at least the GLPK solver is found."""
     assert len(su.solvers) >= 1
     assert "glpk" in su.solvers
 
 
 def test_interface_str() -> None:
-    """Test string representation of solver interfaces."""
+    """Test the string representation of solver interfaces."""
     assert su.interface_to_str("nonsense") == "nonsense"
     assert su.interface_to_str("optlang.glpk_interface") == "glpk"
     assert su.interface_to_str("optlang-cplex") == "cplex"
 
 
 def test_solver_name() -> None:
-    """Test default solver name."""
+    """Test that the default LP solver name is GLPK."""
     assert su.get_solver_name() == "glpk"
 
 
 def test_choose_solver(model: "Model") -> Optional[su.SolverNotFound]:
-    """Test if solver switching is working."""
+    """Test that solver switching is working."""
     so = su.choose_solver(model, "glpk")
     assert su.interface_to_str(so) == "glpk"
 
@@ -49,7 +49,7 @@ def test_choose_solver(model: "Model") -> Optional[su.SolverNotFound]:
 
 
 def test_linear_reaction_coefficients(model: "Model") -> None:
-    """Test if linear coefficients are identifiable in objective."""
+    """Test that linear coefficients are identifiable in objective."""
     coefficients = su.linear_reaction_coefficients(model)
     assert coefficients == {model.reactions.Biomass_Ecoli_core: 1}
 
--- a/src/cobra/test/test_util/test_solver.py
+++ b/src/cobra/test/test_util/test_solver.py
@@ -120,7 +120,7 @@ def test_absolute_expression(model: "Model") -> None:
 
 
 @pytest.mark.parametrize("solver", optlang_solvers)
-def test_fix_objective_as_constraint(solver: List[str], model: "Model") -> None:
+def test_fix_objective_as_constraint(solver: str, model: "Model") -> None:
     """Test fixing present objective as a constraint."""
     model.solver = solver
     with model as m:
@@ -135,9 +135,7 @@ def test_fix_objective_as_constraint(solver: List[str], model: "Model") -> None:
 
 
 @pytest.mark.parametrize("solver", optlang_solvers)
-def test_fix_objective_as_constraint_minimize(
-    model: "Model", solver: List[str]
-) -> None:
+def test_fix_objective_as_constraint_minimize(model: "Model", solver: str) -> None:
     """Test fixing present objective as a constraint but as a minimization."""
     model.solver = solver
     model.reactions.Biomass_Ecoli_core.bounds = (0.1, 0.1)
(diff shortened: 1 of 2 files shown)


src/cobra/util/solver.py
```diff
@@ -49,23 +55,26 @@
 has_primals = [NUMERIC, FEASIBLE, INFEASIBLE, SUBOPTIMAL, ITERATION_LIMIT, TIME_LIMIT]
 
 
-def linear_reaction_coefficients(model, reactions=None):
-    """Coefficient for the reactions in a linear objective.
+def linear_reaction_coefficients(
+    model: "Model", reactions: Optional[List["Reaction"]] = None
+) -> Union[Dict["Reaction", float], dict]:
```
Midnighter: I don't understand right now why the union with `dict`?

src/cobra/util/solver.py
```diff
@@ -95,34 +104,46 @@ def _valid_atoms(model, expression):
...
+    value: Union[
+        optlang.interface.Objective, optlang.symbolics.Basic, Dict["Reaction", float],
+    ],
+    additive: bool = False,
+) -> Optional[Exception]:
     """Set the model objective.
 
     Parameters
     ----------
     model : cobra model
```
Midnighter:
```suggestion
    model : cobra.Model
```

src/cobra/util/solver.py
```diff
@@ -133,9 +154,8 @@ def set_objective(model, value, additive=False):
     if isinstance(value, dict):
         if not model.objective.is_Linear:
             raise ValueError(
-                "can only update non-linear objectives "
-                "additively using object of class "
-                "model.problem.Objective, not %s" % type(value)
+                "You can only update non-linear objectives additively using object of "
+                f"class optlang.interface.Objective, not {type(value)}"
```
Midnighter:
```suggestion
                f"class optlang.interface.Objective, not of {type(value)}."
```

src/cobra/util/solver.py
```diff
@@ -230,38 +251,42 @@ def get_solver_name(mip=False, qp=False):
...
             if solver_name in solvers:
                 return solver_name
-    raise SolverNotFound("no mip-capable solver found")
+    raise SolverNotFound("No MIP-capable solver found.")
 
 
-def choose_solver(model, solver=None, qp=False):
+def choose_solver(
+    model: "Model", solver: Optional[str] = None, qp: bool = False
+) -> Union[ModuleType, Exception]:
```
Midnighter: I would not account for exceptions as return values.

```suggestion
) -> ModuleType:
```

src/cobra/util/solver.py
```diff
@@ -275,86 +300,111 @@ def choose_solver(model, solver=None, qp=False):
...
-    namedtuple
+    NamedTuple
         A named tuple with variable and two constraints (upper_constraint,
         lower_constraint) describing the new variable and the constraints
         that assign the absolute value of the expression to it.
+
     """
-    Components = namedtuple(
-        "Components", ["variable", "upper_constraint", "lower_constraint"]
+    Components = NamedTuple(
```
Midnighter: Can you define `Components` at module level, please, rather than within the function and then set the return type of this function to `-> Components`.

src/cobra/util/solver.py
```diff
@@ -422,21 +479,49 @@ def fix_objective_as_constraint(
     return bound
 
 
-def check_solver_status(status, raise_error=False):
-    """Perform standard checks on a solver's status."""
+def check_solver_status(
+    status: str = None, raise_error: bool = False
+) -> Union[None, UserWarning, OptimizationError]:
```
Midnighter:
```suggestion
) -> None:
```

src/cobra/util/solver.py
```diff
@@ -422,21 +479,49 @@ def fix_objective_as_constraint(
...
         )
     else:
-        raise OptimizationError("solver status is '{}'".format(status))
+        raise OptimizationError(f"Solver status is '{status}'.")
 
 
-def assert_optimal(model, message="optimization failed"):
+def assert_optimal(
+    model: "Model", message: str = "Optimization failed"
+) -> Optional[OptimizationError]:
```
Midnighter:
```suggestion
) -> None:
```

src/cobra/test/test_util/test_solver.py
```diff
@@ -7,26 +9,34 @@
...
+def test_interface_str() -> None:
+    """Test string representation of solver interfaces."""
     assert su.interface_to_str("nonsense") == "nonsense"
     assert su.interface_to_str("optlang.glpk_interface") == "glpk"
     assert su.interface_to_str("optlang-cplex") == "cplex"
 
 
-def test_solver_name():
+def test_solver_name() -> None:
+    """Test default solver name."""
```
Midnighter:
```suggestion
    """Test that the default LP solver name is GLPK."""
```

src/cobra/test/test_util/test_solver.py
```diff
@@ -7,26 +9,34 @@
...
+
+
 stable_optlang = ["glpk", "cplex", "gurobi"]
-optlang_solvers = ["optlang-" + s for s in stable_optlang if s in su.solvers]
+optlang_solvers = [f"optlang-{s}" for s in stable_optlang if s in su.solvers]
 
 
-def test_solver_list():
+def test_solver_list() -> None:
+    """Test if solvers are found."""
```
Midnighter:
```suggestion
    """Expect that at least the GLPK solver is found."""
```

src/cobra/test/test_util/test_solver.py
```diff
@@ -7,26 +9,34 @@
...
     assert len(su.solvers) >= 1
     assert "glpk" in su.solvers
 
 
-def test_interface_str():
+def test_interface_str() -> None:
+    """Test string representation of solver interfaces."""
```
Midnighter:
```suggestion
    """Test the string representation of solver interfaces."""
```

src/cobra/test/test_util/test_solver.py
```diff
@@ -7,26 +9,34 @@
...
 
-def test_solver_name():
+def test_solver_name() -> None:
+    """Test default solver name."""
     assert su.get_solver_name() == "glpk"
 
 
-def test_choose_solver(model):
+def test_choose_solver(model: "Model") -> Optional[su.SolverNotFound]:
+    """Test if solver switching is working."""
```
Midnighter:
```suggestion
    """Test that solver switching is working."""
```

src/cobra/test/test_util/test_solver.py
```diff
@@ -38,13 +48,17 @@ def test_choose_solver(model):
             su.choose_solver(model, qp=True)
 
 
-def test_linear_reaction_coefficients(model):
+def test_linear_reaction_coefficients(model: "Model") -> None:
+    """Test if linear coefficients are identifiable in objective."""
```
Midnighter:
```suggestion
    """Test that linear coefficients are identifiable in objective."""
```

src/cobra/test/test_util/test_solver.py
```diff
@@ -38,13 +48,17 @@ def test_choose_solver(model):
...
     coefficients = su.linear_reaction_coefficients(model)
     assert coefficients == {model.reactions.Biomass_Ecoli_core: 1}
 
 
 @pytest.mark.parametrize("solver", optlang_solvers)
-def test_fail_non_linear_reaction_coefficients(model, solver):
+def test_fail_non_linear_reaction_coefficients(
+    model: "Model", solver: List[str]
+) -> Optional[ValueError]:
```
Midnighter: In this test it'd be nicer to only set GLPK and expect the `ValueError` always.

```suggestion
) -> None:
```

Midnighter (changes requested):
Thanks for upgrading those modules @synchon :slightly_smiling_face: Just some minor comments. Also, please be aware that the type of the `solver` argument (provided by the fixture of the same name) in the test cases is `str` and not `List[str]`.

src/cobra/util/solver.py
synchon: Because the function returns empty dictionary if there are no linear terms in the objective.

src/cobra/util/solver.py
synchon: Okay you have a point there. Will change it. 👍

src/cobra/util/solver.py
Midnighter: An empty dictionary can still be of type `Dict["Reaction", float]` so you don't need the union.

synchon: I see. Okay I'll make the change.

src/cobra/util/solver.py
synchon: Sure no problem.

src/cobra/test/test_util/test_solver.py
synchon: I have updated the test.

synchon:
@Midnighter I have addressed your comments and also updated the test case for `test_fail_non_linear_reaction_coefficients`. Does it look okay?

Midnighter (approved):

synchon:
Shall we merge it?

Midnighter:
I pushed two more commits onto devel for slightly optimizing the type annotations. Only saw that in my IDE, sorry.

PR #995, opened 2020-08-26
feat: load remote models

### LINKED ISSUE #973: Extract the load_model functionality from cameo
This one is long overdue and should finally conclude the proper separation of cobrapy and cameo (as far as I remember). Making a good implementation of load_model in cobrapy will increase its user friendliness a lot and means that we can get rid of the nasty `cobra.test.create_test_model` from our documentation.

Midnighter:
* [x] fix #973 
* [x] tests added/passed

This PR adds a new function `load_model` that is similar to the cameo version. The function can currently load models from BiGG and BioModels. More repositories can be added in future. I chose to not make `load_model` also be able to load local models in different formats because we have specific functions for that.

A further feature is an automatic disk cache that stores the SBML documents. Thus loading the same model identifier will be much faster the second time.


cdiener:
Can you separate the feature PR from all the formatting adjustments? Bit hard to review like that :sweat_smile:

Midnighter:
> Can you separate the feature PR from all the formatting adjustments? Bit hard to review like that sweat_smile

Sorry, formatting was necessary for CI to pass. You could review all commits up until the last one that should exclude those style changes.

Midnighter:
Alright, I've pushed a commit with the style changes to devel and dropped it from this branch. Hopefully things are clearer now.

Midnighter:
@cdiener I've updated the branch. _Should_ pass CI now.

setup.cfg
```diff
@@ -51,35 +51,40 @@ keywords =
...
+    appdirs ~= 1.4
 	depinfo
+    diskcache ~= 5.0
 	future
+    httpx ~= 0.14
 	importlib_resources
 	numpy ~= 1.13
 	optlang ~= 1.4
 	pandas ~= 1.0
+    pydantic ~= 1.6
```
cdiener, line 64: Do you need to trigger validation with pydantic or does that happen automatically during instantiation? I don't think it adds that much to the codebase here but if you have plans to use it in more places it may be okay.

setup.cfg
```diff
@@ -51,35 +51,40 @@ keywords =
 
 [options]
 zip_safe = True
-install_requires = 
+install_requires =
+    appdirs ~= 1.4
 	depinfo
+    diskcache ~= 5.0
 	future
+    httpx ~= 0.14
```
cdiener, line 59: Would it be much harder to download using the native Python http module? This an unstable version so it would probably break quite often.

setup.cfg
```diff
@@ -51,35 +51,40 @@ keywords =
...
 	importlib_resources
 	numpy ~= 1.13
 	optlang ~= 1.4
 	pandas ~= 1.0
+    pydantic ~= 1.6
 	python-libsbml-experimental == 5.18.1
+    rich ~= 6.0
```
cdiener, line 66: If we depend on rich we should probably update our logging to use that. Can be a different PR though.

src/cobra/io/web/bigg_models_repository.py
```diff
@@ -0,0 +1,84 @@
...
+        *,
+        url: Union[httpx.URL, str] = "http://bigg.ucsd.edu/static/models/",
+        **kwargs,
+    ):
+        """
+        Initialize a BiGG Models repository interface.
+
+        Parameters
+        ----------
+        url : httpx.URL or str, optional
```
cdiener: Why should you be able to set this? This should be constant and immutable for BIGG. Otherwise this would provide a security risk where you can make it look like you're downloading from BIGG even if you aren't.

src/cobra/io/web/biomodels_repository.py
```diff
@@ -0,0 +1,113 @@
...
+        *,
+        url: Union[str, httpx.URL] = "https://www.ebi.ac.uk/biomodels/model/",
+        **kwargs,
+    ):
+        """
+        Initialize a BioModels repository interface.
+
+        Parameters
+        ----------
+        url : httpx.URL or str, optional
```
cdiener: Same as before. URL should be immutable.

src/cobra/io/web/bigg_models_repository.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Other Parameters
+        ----------------
+        kwargs
+            Passed to the parent constructor in order to enable multiple inheritance.
+
+        """
+        super().__init__(url=url, **kwargs)
+
+    def get_sbml(self, model_id: str) -> bytes:
```
cdiener, line 40: There seems to be redundancy between the Biomodels and BIGG implementations. Maybe first build a general SBML DB loader and then juts set the base URL in the BIGG and Biomodels classes.

src/cobra/io/web/bigg_models_repository.py
```diff
@@ -0,0 +1,84 @@
...
+
+from io import BytesIO
+from typing import Union
+
+import httpx
+
+from .abstract_model_repository import AbstractModelRepository
+
+
+class BiGGModels(AbstractModelRepository):
```
cdiener, line 12: I like this design. Would be great if we could a DB package that adds more like AGORA and CarveME.

src/cobra/io/web/load.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+    else:
+        data = _fetch_model(model_id=model_id, repositories=repositories)
+    return get_model_from_gzip_sbml(data)
+
+
+def _cached_load(
+    model_id: str,
+    repositories: Iterable[AbstractModelRepository],
+    cache_directory: pathlib.Path,
```
cdiener: This together with the maximum cache dir size and a time when to clean it up should be set in the configuration and not here.

src/cobra/io/web/load.py
```diff
@@ -0,0 +1,201 @@
...
+    Parameters
+    ----------
+    model_id : str
+        The identifier of the desired metabolic model. This is typically repository
+        specific.
+    repositories : iterable, optional
+        An iterable of repository accessor instances. The model_id is searched in order.
+    cache : bool, optional
+        Whether or not to use the local caching mechanism (default yes).
+    cache_directory : pathlib.Path or str, optional
```
cdiener: Set a maximum size for this directory. For instance a lot of my workflows download hundreds or thousands of models. I would not want cobrapy using GBs of storage that I can't control.

setup.cfg
```diff
@@ -51,35 +51,40 @@ keywords =
 
 [options]
 zip_safe = True
-install_requires = 
+install_requires =
+    appdirs ~= 1.4
```
cdiener, line 55: Some of the deps do not support Python 3.5 so we will need to drop support for Python <3.6.

cdiener (commented):
So I do like the general design. It will allow for a lot of additional DBs. I am a bit worried about all the added dependencies. That will lead to more stuff breaking so the extra functionality provided by it should be justified. In the end I'll leave that up to you since you are probably the person that will need to fix it :sweat_smile:

setup.cfg
Midnighter: Validation happens at instantiation time indeed. It can also be configured to additionally run at assignment. It does not add incredibly much for the moment but I would love to use it for our `dict`, `json`, `yaml` IO as one can automatically generate a JSON schema from pydantic models.

cdiener: How fast is pydantic? Model loading needs to be pretty performant so that may generate some overhead. (Sorry, this is a bit off-topic).

setup.cfg
Midnighter: AFAIK, the httpx API is fairly stable by now and will soon see a 1.0 release. You're right that it's probably not too much harder, although it is quite convenient.

setup.cfg
Midnighter: Nice alternative is tqdm but rich indeed opens the door to many console improvements.

src/cobra/io/web/bigg_models_repository.py
Midnighter: Fair point. Just a habit in case someone hosts their own BiGG database or the URL has changed and we haven't caught up yet (or new URL with old code version).

src/cobra/io/web/bigg_models_repository.py
Midnighter: They are very similar but there are some slight variations, for example, URL parameters, content-length header. Other databases may be also slightly different. That's why I opted for 'duplicating' the code.

src/cobra/io/web/bigg_models_repository.py
Midnighter: :+1: thanks. I definitely had in mind that more interfaces can be added.

Midnighter: Do you mean a separate Python package entirely or a sub-package within cobrapy?

setup.cfg
cdiener: I think rich is fine since it bundles a lot of console stuff. We don't have that many progress bars that we need tqdm.

src/cobra/io/web/bigg_models_repository.py
cdiener: Fair enough. :+1:

src/cobra/io/web/load.py
Midnighter: Yes, good idea.

src/cobra/io/web/load.py
Midnighter: You can disable the caching entirely, of course, but I will add it to the configuration.

setup.cfg
Midnighter: We already have since we introduced f-strings.

src/cobra/io/web/bigg_models_repository.py
cdiener: Separate package.

Midnighter:
> So I do like the general design. It will allow for a lot of additional DBs. I am a bit worried about all the added dependencies. That will lead to more stuff breaking so the extra functionality provided by it should be justified. In the end I'll leave that up to you since you are probably the person that will need to fix it sweat_smile

Haha, yes, I was fearing that as well. I'm hoping that slightly stricter pinning will contain the problem a bit. I think we could live without `rich` easily and can also replace `httpx`.

setup.cfg
Midnighter: It's the most performant of the structured data packages that I know. Can be further improved by using something like `ujson` or `orjson`. Should run some benchmarks first, of course, but I think it's a huge advantage to be able to define the data structures as classes, have good validation, and be able to generate a JSON schema automatically.

Midnighter:
I didn't fully understand you so far, do you think it's better to close this PR and create a separate package now? (And have cobrapy depend on it?) I think it's a reasonable idea but would like something aking to the `load_model` function in cobrapy.

cdiener:
Oh sorry. Not at all. I just meant that additional database loaders (inherited classes) should be in a separate "upgrade" package. So cobrapy comes with load_model as well as biomodels/BIGG access but additional model DBs would be in a separate package.

matthiaskoenig:
Quick comment: Many download urls and model urls do not only have base_url prefixes but also suffixes. So it is much better praxis to just provide a regular expression URL which replaces the identifier, e.g.
```
https://modeldb.co/models/{id}?download=true&format=zip
``` 
This is especially important if url parameters are required for download.

src/cobra/io/web/bigg_models_repository.py
Midnighter: I was thinking, though, you can host your own on-premise installation of BiGG so maybe that's a reason to keep it flexible?

Midnighter:
> Quick comment: Many download urls and model urls do not only have base_url prefixes but also suffixes. So it is much better praxis to just provide a regular expression URL which replaces the identifier, e.g.
> 
> ```
> https://modeldb.co/models/{id}?download=true&format=zip
> ```
> 
> This is especially important if url parameters are required for download.

This is actually an area where httpx really shines. Similarly to the native `urllib.request.Request` object, you can construct and manipulate the URL components separately. Especially URL parameters are simply passed as a dictionary and automatically url-encoded. https://www.python-httpx.org/api/#url

Midnighter:
* I've moved the cache configuration to the global configuration object
* I've made the URLs faux-immutable but inspectable by property (can still be changed by assigning to `_url`)
* I've kept all the new dependencies because
    * appdirs is tiny and helpful
    * pydantic is a single dependency and holds great future potential for us, too
    * rich also provides the opportunity for future improvements
    * httpx because I trust that their API will remain stable and that it will make any future web connections a breeze

@cdiener please give it another look

src/cobra/core/configuration.py
```diff
@@ -45,6 +48,13 @@ class Configuration(metaclass=Singleton):
         A default number of processes to use where multiprocessing is
         possible. The default number corresponds to the number of available
         cores (hyperthreads) minus one.
+    cache_directory : pathlib.Path or str, optional
+        A path where the model cache should reside if caching is desired. The
+        default directory depends on the operating system.
+    max_cache_size : int, optional
+        The allowed maximum size of the model cache in bytes (default 1 GB).
+    cache_experiation : int, optional
```
matthiaskoenig:
```suggestion
    cache_expiration : int, optional
```

src/cobra/core/configuration.py
```diff
@@ -80,6 +94,12 @@ def _set_default_processes(self) -> None:
         if self.processes > 1:
             self.processes -= 1
```
matthiaskoenig, line 96: What is this line here?! It looks like the processes are reduced every time this function is called. This does not make sense and looks like a bug. Please document what this is for.

src/cobra/io/web/biomodels_repository.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        response.raise_for_status()
+        files = BioModelsFilesResponse.parse_obj(response.json())
+        is_found = False
+        for model in files.main:
+            if model.name.endswith("xml"):
+                is_found = True
+                break
+        if not is_found:
+            RuntimeError(f"'{model_id}' does not seem to contain an SBML document.")
```
matthiaskoenig: Python has a `for` `else` for this use case which is much cleaner to read, also `model_id` -> `model` in RunTimeError
``` 
        is_found = False
        for model in files.main:
            if model.name.endswith("xml"):
                is_found = True
                break
        if not is_found:
            RuntimeError(f"'{model_id}' does not seem to contain an SBML document.")
``` 
=>
```
        for model in files.main:
            if model.name.endswith("xml"):
                break
        else:
            RuntimeError(f"'{model}' does not seem to contain an SBML document.")
```

matthiaskoenig: In addition you have to check that the file is an SBML file. There are other XML files in the Combine archives on biomodels such as SED-ML files which also have xml endings.

src/cobra/io/web/load.py
```diff
@@ -0,0 +1,188 @@
...
+
+logger = logging.getLogger(__name__)
+configuration = Configuration()
+
+
+def load_model(
+    model_id: str,
+    repositories: Iterable[AbstractModelRepository] = (BiGGModels(), BioModels()),
+    cache: bool = True,
+) -> "Model":
```
matthiaskoenig, line 31: Should probably be `cobra.core.Model`

matthiaskoenig (approved):

src/cobra/core/configuration.py
Midnighter: The function is only called once when the configuration is instantiated. The configuration is a singleton so it is only initialized once ever. The purpose is to set the default number of processes to `nproc - 1` so that cobrapy by default doesn't hog all CPUs. I can extend the docstring but it's not a change of this PR.

src/cobra/io/web/biomodels_repository.py
Midnighter: I didn't know about `for`... `else` that's awesome. Thanks for teaching me something new about Python :slightly_smiling_face: 

> In addition you have to check that the file is an SBML file. There are other XML files in the Combine archives on biomodels such as SED-ML files which also have xml endings.

My observation about the BioModels response was that under the `"main"` key, you only have the model file, all the other files are in the `"additional"` key (which I ignore). https://www.ebi.ac.uk/biomodels/docs/#/operations/Model-related%20operations/get_model_files__modelId_ Do you know this to be false?

src/cobra/io/web/load.py
Midnighter: Since it's imported as follows, I think it's fine.

```py
if TYPE_CHECKING:
    from cobra.core import Model
```

src/cobra/core/configuration.py
```diff
@@ -56,10 +66,14 @@ def __init__(self, **kwargs) -> None:
         self.lower_bound = None
         self.upper_bound = None
         self.processes = None
+        self._cache_directory = None
+        self.max_cache_size = 1024 ** 3
```
cdiener: This seems large as default especially since the expiration is infinite.

cdiener (approved):
Looks good. I would set a default expiration of 1 year and make cache space smaller.

Midnighter:
> Looks good. I would set a default expiration of 1 year and make cache space smaller.

1 GB of disk space doesn't seem like a lot these days then again that's easily space for a few thousand models which most users will never fill up. I'll change it.

I decided against an expiration time by default because I didn't want users to be surprised by a changing model from one run to another.

cdiener:
I don't know. At some point I had a chromebook that only had 16GB of disk space for instance.

Midnighter:
I've included it now with a limit of 100 MB. Let's see how it performs in the wild. We can certainly tweak in future.

synchon:
The PR looks great! Just a quick reminder about the documentation of the added feature.

Midnighter:
> The PR looks great! Just a quick reminder about the documentation of the added feature.

You are :100: spot on but I just don't have the energy at the moment.

src/cobra/io/web/biomodels_repository.py
matthiaskoenig: Did not know about the `main` flag. The flag should clearly identify the SBML model, so no need to check if the `*.xml` is an SBML in this case.

PR #996, opened 2020-08-28
User defined constraints

Hemant27031999:
## Description of feature
**User Defined Constraint:** This new feature from fbc-v3 has been added to COBRApy as well as to JSON and other formats also. A user can now add his own constraints to the model. He can do this even more easily by just passing the constraint expression (for eg. "2v1 - 3v2*v2 +0.4 * v3" where v1, v2 and v3 are either reaction ids to define fluxes or some variable) along with the lower and upper bounds.

## Tests
Tests for the functionality has been added along with a new example model.

#### Note: 

- This PR is build upon the [metadata pull request](https://github.com/opencobra/cobrapy/pull/988/), and should be merged only after that.
- The libsbml library currently doesn't support fbc-v3, so SBML parser has not been added for it.


Midnighter:
It's almost impossible to tell what this PR adds because it includes all of the changes of the other branch. @matthiaskoenig can you open a branch on this repository, please, for the other work and then make a pull request from this branch targetting that new branch?

PR #999, opened 2020-09-01
refactor: cobra.util.util

synchon:
* [x] tests added/passed

This PR refactors `cobra.util.util` module and adds additional tests for the module.

### DIFF
--- a/src/cobra/test/test_util/test_util.py
+++ b/src/cobra/test/test_util/test_util.py
@@ -1,9 +1,37 @@
-"""Test functions of util.py"""
+"""Test functions of util.py ."""
 
-from cobra.util import show_versions
+from cobra.util import AutoVivification, format_long_string, show_versions
 
 
-def test_show_versions(capsys):
+def test_format_long_string_with_long_string() -> None:
+    """Test functionality by formatting a long string."""
+    long_string = (
+        "This is a really long string, but is it long enough."
+        "I hope it is long enough so that format_long_string() function works."
+    )
+    expected_output = "This is a really long string, but is it long en..."
+    assert expected_output == format_long_string(long_string)
+
+
+def test_format_long_string_with_short_string() -> None:
+    """Test functionality by formatting a short string."""
+    short_string = "This is short string."
+    assert short_string == format_long_string(short_string)
+
+
+def test_autovivification() -> None:
+    """Test proper functionality of autovivification."""
+    test_data = AutoVivification()
+    test_data["a"]["b"] = 1
+    test_data["c"]["d"] = 2
+    assert test_data["a"] == {"b": 1}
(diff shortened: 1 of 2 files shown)


src/cobra/util/util.py
```diff
@@ -1,27 +1,62 @@
...
+    -------
+    str
+        The shortened string.
+
+    """
     if len(string) > max_length:
-        string = string[: max_length - 3]
-        string += "..."
+        string = f"{string[: max_length - 3]}..."
     return string
```
Midnighter: Can you replace the body of the function with [textwrap.shorten](https://docs.python.org/3/library/textwrap.html#textwrap.shorten), please?

```py
return textwrap.shorten(string, width=max_length, placeholder="...")
```

Midnighter: A future refactoring could then be to replace all calls to `format_long_string` with `textwrap.shorten` directly.

src/cobra/util/util.py
```diff
@@ -1,27 +1,62 @@
...
+
+        Returns
+        -------
+        Any
+            The retrieved object.
 
-    def __getitem__(self, item):
+        """
         try:
             return dict.__getitem__(self, item)
```
Midnighter:
```suggestion
            value = super().__getitem__(self, item)
```

src/cobra/util/util.py
```diff
@@ -1,27 +1,62 @@
...
         except KeyError:
             value = self[item] = type(self)()
             return value
```
Midnighter:
```suggestion
        return value
```

src/cobra/test/test_util/test_util.py
```diff
@@ -1,9 +1,37 @@
...
+    )
+    expected_output = "This is a really long string, but is it long en..."
+    assert expected_output == format_long_string(long_string)
+
+
+def test_format_long_string_with_short_string() -> None:
+    """Test functionality by formatting a short string."""
+    short_string = "This is short string."
+    assert short_string == format_long_string(short_string)
+
```
Midnighter: You could work with `@pytest.mark.parametrize` to condense these two test definitions into one and simply test different instances.

Midnighter (changes requested):
Nice improvements. Some comments but looks good. The single module + test module PRs are much easier to review :smiley:

src/cobra/util/util.py
synchon: Perfect.

src/cobra/test/test_util/test_util.py
synchon: Yes I'll update it.

Midnighter (approved):
Will merge when tests have passed.

PR #1011, opened 2020-10-09
Chore upgrade optlang

Midnighter:
Necessary changes for updating optlang to 1.4.7. At the moment this is failing.


cdiener:
Can you rerun the CI to see what's going on? looks like the test model files were changed though.

Midnighter:
Seems like it was too old to just restart. I've rebased on latest and pushed again.

cdiener:
I think the problem is at https://github.com/opencobra/cobrapy/blob/devel/src/cobra/flux_analysis/gapfilling.py#L99. Just need to wrap it in an if block that checks if the attribute exists before setting it.

Midnighter:
Those are all just warnings for GLPK but the actual problem is a failed validation of the solution in all cases. Do you have any ideas?

```py
      def fill(self, iterations=1):
          """Perform the gapfilling by iteratively solving the model, updating
          the costs and recording the used reactions.
      
      
          Parameters
          ----------
          iterations : int
              The number of rounds of gapfilling to perform. For every
              iteration, the penalty for every used reaction increases
              linearly. This way, the algorithm is encouraged to search for
              alternative solutions which may include previously used
              reactions. I.e., with enough iterations pathways including 10
              steps will eventually be reported even if the shortest pathway
              is a single reaction.
      
          Returns
          -------
          iterable
              A list of lists where each element is a list reactions that were
              used to gapfill the model.
      
          Raises
          ------
          RuntimeError
              If the model fails to be validated (i.e. the original model with
              the proposed reactions added, still cannot get the required flux
              through the objective).
          """
          used_reactions = list()
          for i in range(iterations):
              self.model.slim_optimize(
                  error_value=None, message="gapfilling optimization failed"
              )
              solution = [
                  self.model.reactions.get_by_id(ind.rxn_id)
                  for ind in self.indicators
                  if ind._get_primal() > self.integer_threshold
              ]
              if not self.validate(solution):
  >               raise RuntimeError(
                      "failed to validate gapfilled model, "
                      "try lowering the integer_threshold"
                  )
  E               RuntimeError: failed to validate gapfilled model, try lowering the integer_threshold
```

cdiener:
Looks like the tolerance issue has crept it's way into the test problems. We should probably add more tolerances in optlang. You could try to use the fix proposed in https://github.com/opencobra/cobrapy/issues/941#issuecomment-658842054 . But @gregmedlock probably has a better grasp what's a good strategy.

cdiener:
I sent some adjustments to the GLPK interface to optlang. With those the gapfilling tests pass.

Midnighter:
> I sent some adjustments to the GLPK interface to optlang. With those the gapfilling tests pass.

~I lost track, are those merged and released or which PR is that in?~

https://github.com/opencobra/optlang/pull/231

Midnighter:
Using your optlang branch locally, I then get an error in ROOM:

```
___________________________________________________________________________________ test_room_sanity[glpk] ___________________________________________________________________________________

model = <Model e_coli_core at 0x7f1f6d588b70>, all_solvers = 'glpk'

    def test_room_sanity(model, all_solvers):
        """Test optimization criterion and optimality for ROOM."""
        model.solver = all_solvers
        sol = model.optimize()
        with model:
            model.reactions.PYK.knock_out()
            knock_sol = model.optimize()

        with model:
            # Internally uses pFBA as reference solution.
            add_room(model)
            model.reactions.PYK.knock_out()
            room_sol = model.optimize()

        with model:
            # Use FBA as reference solution.
            add_room(model, solution=sol)
            model.reactions.PYK.knock_out()
            room_sol_ref = model.optimize()

        flux_change = (sol.fluxes - knock_sol.fluxes).abs().sum()
        flux_change_room = (sol.fluxes - room_sol.fluxes).abs().sum()
        flux_change_room_ref = (sol.fluxes - room_sol_ref.fluxes).abs().sum()
        # Expect the ROOM solution to have smaller flux changes in
        # reactions compared to a normal FBA.
        assert flux_change_room < flux_change or np.isclose(
            flux_change_room, flux_change, atol=1e-06
        )
        # Expect the FBA-based reference to have less change in
        # flux distribution.
>       assert flux_change_room_ref > flux_change_room or np.isclose(
            flux_change_room_ref, flux_change_room, atol=1e-06
        )
E       assert (15.798173818751824 > 17.609129149207746 or False)
E        +  where False = <function isclose at 0x7f1f7f13f510>(15.798173818751824, 17.609129149207746, atol=1e-06)
E        +    where <function isclose at 0x7f1f7f13f510> = np.isclose
```

cdiener:
Let me check maybe I was running mine with Cplex by accident.

EDIT: can reproduce the failure.

cdiener:
I think the test does not really check if ROOM works, it's more a linear MOMA test. 

Room does not minimize the distance between fluxes but the number of reactions that have to be adjusted by more than a small change to get the old objective.

And that works. for that example the room_sol is 5 (#reactions changed significantly). And you can check that this is better than the trivial solution by quite a bit (5 vs 29 adjustments):

```
In [30]: flux_change = (sol.fluxes - room_sol.fluxes).abs()

In [31]: sum(flux_change > 0.03 * sol.fluxes.abs() + 1e-3 + 1e-6)
Out[31]: 5

In [32]: flux_change = (sol.fluxes - knock_sol.fluxes).abs()

In [33]: sum(flux_change > 0.03 * sol.fluxes.abs() + 1e-3 + 1e-6)
Out[33]: 29
```

This test is more likely to be correct for linear ROOM. 

I can fix the test directly in this PR if you want.

Midnighter:
Please go ahead.

cdiener:
Probably won't pass until https://github.com/opencobra/optlang/pull/231 is released. But it seems to work on my branch. 

EDIT: ROOM even passes with the current version.

cdiener:
Let's see. Should be good now with optlang 1.5.1. This should be followed by a PR activating Python 3.9 everywhere.

cdiener (approved):
Okay, looks like we are finally good.

Midnighter:
Do we need to update the pickled models? I expected so but the tests pass...

cdiener:
Let me check.

Midnighter:
🤷🏼‍♂️ cool

Midnighter:
I'll go through cameo tomorrow just to make sure a release won't break things there. Same for memote. Then I'll release a new version 🎉

cdiener:
Sure, no hurry from my side.

Issue #739, opened 2018-07-29
Create symbolic GPRs or otherwise improve the GPR handling


Midnighter:

synchon:
The `cobra.core.gene` module hasn't been updated in a while and this might be a good issue to start with. Any thoughts on this one? I can have a go if nobody is working on this.

matthiaskoenig:
Yes, this needs urgent updating. My thoughts
- this currently sucks, more specifically there is a lot of manual conversion between GPR strings in the model formats and the internal GPR representation
- this is speed sensitive, i.e. a lot of the model parsing (at least for SBML) goes in processing the GPRs, because they have to be converted between trees and strings. I.e. the processing and handling cannot get (much) slower as it currently is.
- this should be future proof. A major limitation of current GPRs is that the miss the stoichiometry, i.e. how many subunits of one gene and another form the final protein. The main issue is that proteins are not actually represented, but at least the subunits should be encoded correctly (these are factors 2-3 incorrect otherwise). A new implementation should be able to account for stoichiometry. E.g. a heterodimer protein A2B2 should have the GPR (2 G_A and 2 G_B)
- this should have a unique way for serialization so that GPRs can be compared for equality.

Midnighter:
Another aspect of this was the idea to integrate genes with `optlang.Symbol`s such that one could use sympy to directly evaluate the Boolean expressions. I think taking all of this on, though, is quite a major piece of refactoring and new work. We should plan this out in a bit more detail before writing any code.

matthiaskoenig:
@Midnighter I agree, this is a major issue and needs some planning. 
Just listed things which I noticed with the GPRs. By no means should this all be solved at once.

cdiener:
Sorry just saw that. @matthiaskoenig, could you comment on what you mean by 

> this should be future proof. A major limitation of current GPRs is that the miss the stoichiometry, i.e. how many subunits of one gene and another form the final protein. The main issue is that proteins are not actually represented, but at least the subunits should be encoded correctly (these are factors 2-3 incorrect otherwise). A new implementation should be able to account for stoichiometry. E.g. a heterodimer protein A2B2 should have the GPR (2 G_A and 2 G_B)

[...]

PR #1013, opened 2020-10-15
Shifting GPR from AST to sympy

akaviaLab:
* [X] description of feature
This shifts GPR from AST to sympy, which will enable to compare GPRs by logic, and not by string
Added function gpr_eq() which compares GPRs by logic
Added parse_gpr_sympy, eval_gpr_sympy and sympy2str which are equivalent to the previous (ast) version
* [X] tests added/passed
Added test for gpr_eq()

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -302,3 +302,10 @@ gurobi.log
 /documentation_builder/test*\.*
 /.benchmarks
 /.testmondata
+/.idea/.gitignore
+/.idea/cobrapy.iml
+/.idea/misc.xml
+/.idea/modules.xml
+/.idea/inspectionProfiles/profiles_settings.xml
+/.idea/inspectionProfiles/Project_Default.xml
+/.idea/vcs.xml
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ parse = (?P<major>\d+)
 	\.(?P<minor>\d+)
 	\.(?P<patch>\d+)
 	(?P<release>[a]*)(?P<num>\d*)
-serialize = 
+serialize =
 	{major}.{minor}.{patch}{release}{num}
 	{major}.{minor}.{patch}
 tag_name = {new_version}
@@ -15,7 +15,7 @@ tag_name = {new_version}
 name = cobra
 url = https://opencobra.github.io/cobrapy
 download_url = https://pypi.org/project/cobra
-project_urls = 
+project_urls =
 	Source Code = https://github.com/opencobra/cobrapy
 	Documentation = https://cobrapy.readthedocs.io
 	Bug Tracker = https://github.com/opencobra/cobrapy/issues
@@ -23,7 +23,7 @@ author = The cobrapy core development team.
 author_email = cobra-pie@googlegroups.com
 maintainer = Moritz E. Beber
 maintainer_email = moritz.beber@gmail.com
-classifiers = 
+classifiers =
 	Development Status :: 5 - Production/Stable
 	Intended Audience :: Science/Research
 	License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)
@@ -39,7 +39,7 @@ license = LGPL-2.0-or-later OR GPL-2.0-or-later
 description = COBRApy is a package for constraint-based modeling of metabolic networks.
 long_description = file: README.rst, INSTALL.rst
 long_description_content_type = text/x-rst
-keywords = 
+keywords =
 	metabolism
 	biology
 	constraint-based
@@ -51,7 +51,7 @@ keywords =
 
 [options]
 zip_safe = True
-install_requires = 
+install_requires =
 	appdirs ~=1.4
 	depinfo
 	diskcache ~=5.0
@@ -67,24 +67,25 @@ install_requires =
 	ruamel.yaml ~=0.16
 	six
 	swiglpk
-tests_require = 
+	sympy
+tests_require =
 	tox
 packages = find:
-package_dir = 
+package_dir =
 	= src
 
 [options.packages.find]
 where = src
 
 [options.package_data]
-cobra = 
+cobra =
 	io/*.json
 	test/data/*
 
 [options.extras_require]
-array = 
+array =
 	scipy
-development = 
+development =
 	black
 	bumpversion
 	isort
@@ -96,7 +97,7 @@ universal = 1
 [bumpversion:part:release]
 optional_value = placeholder
 first_value = placeholder
-values = 
+values =
 	placeholder
 	a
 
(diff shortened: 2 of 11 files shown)


Midnighter:
@akaviaLab thank you for the pull request! It's a pretty big change so (a) it would have been nice to discuss the design in an issue before working on an implementation and (b) it will take a little bit of time to review it. It's definitely something I've wanted in cobrapy for a long time so I am happy that you are making it happen.

matthiaskoenig:
Also please provide some benchmarks on this. How does it affect loading of genome-scale models which require a lot of parsing of GPRs. If this is 20-30% increased load time everything is fine. If it takes much longer this could be problematic. Please run some benchmark tests with models like Recon3D which are very big.

akaviaLab:
Using parse_expr from sympy (the original pull request) fails on RECON3D because some of the GPRs are too long.
I rewrote it to use AST as an intermediate step
So I profiled RECON 2.2 and RECON3D with the three various options (focusing on the median of loading 20 times)
parse_expr - slow, and fails RECON3D
ast_visitor - 7% increase in RECON2.2, 21% increase in RECON3D 
ast_func -  5% increase in RECON2.2, 21% increase in RECON3D

[load_sbml_timings.xlsx](https://github.com/opencobra/cobrapy/files/5476477/load_sbml_timings.xlsx)

I would happy to change my code so the CI runs the SBML loading benchmark on the three different versions of the sympy code, but I don't know how. If you would like me to do that, please let me know.
So far, I ran it manually on my computer, changing the code each time.

I think the best option is ast_visitor, since it seems cleaner. Once it has been decided, I can patch it so only the relevant code remains.

cdiener:
Does it help the benchmarks to use symengine?

akaviaLab:
I don't know about symengine. I have a Mac, so installing the C++ symengine from source would be a pain and  I'd prefer not doing it.
I think that 21% slowdown was acceptable to the other developers. If need be, I can try it with symengine

Midnighter:
> I don't know about symengine. I have a Mac, so installing the C++ symengine from source would be a pain and I'd prefer not doing it.

There should be pre-compiled wheels for Mac. So it's just a pip install away.

akaviaLab:
[load_sbml_timings.xlsx](https://github.com/opencobra/cobrapy/files/5483896/load_sbml_timings.xlsx)
Okay.
Installing it makes it speedier.
No slower for RECON2.2, about 14% slower for RECON3D.

Any comments on the pull request?

akaviaLab:
see #739 for some discussion

cdiener:
Okay took me a while to get through that. So, as far as I understand this is different from what was discussed in #739. I think the idea was that GPRs themselves will be sympy expressions which will also make them faster to evaluate and nicer to manage. However, as I understand this PR they are still strings and are re-parsed everytime they are evaluated. This explains why you don't see faster processing. I think it would be more efficient to have GPRs be sympy expressions and genes be symbols/entities. Parsing from any format should immediately yield sympy expressions and writing to any format should just generate the output from the expression. This would remove the overhead of parsing on evaluation (such as *.functional or knocking the genes). 

As far as implementation goes my general comments would be:

- No need to leave the old parsers in there, just substitute them with the new (maintaining the name). So just replace parse_gpr.
- We should be aware that this could be run with symengine and sympy. Right now optlang always brings in sympy (even with symengine) but this might be different in the future and the code should be able to run with only symengine as well. I think the best way is to use the `optlang.symbolics` interface. I think this should bring everything you need for the GPRs and would make it agnostic to the backend (sympy or symengine).

matthiaskoenig:
> However, as I understand this PR they are still strings and are re-parsed everytime they are evaluated. 

Yes, this. The parsing should be done exactly 1 time on model loading. Everything on the internal cobra model should work with symbolic expressions.

akaviaLab:
Okay, just to make sure I got this
- GPRs should be symbolic expressions, which are parsed everytime things are read (so check that import and export from all formats work with symbolic)
- I'm not sure what this phrase means "genes by symbols/entities". Can you please clarify?
- Use optlang.symbolics instead of direct sympy, which should be agnostic to sympy, symengine (or any replacements in the future)
- benchmark speed changes
- remove all old stuff when new stuff works
- Have I got this correct? If so, let me know if you prefer closing this pull request and opening a new one, or discussing when it is partly done, or what works for you guys.

Thank you!

cdiener:
Thanks for getting back to us so quickly @akaviaLab!

> GPRs should be symbolic expressions, which are parsed everytime things are read (so check that import and export from all formats work with symbolic)

Yes, exactly! All parsers should return GPRs as symbolic expressions.

> I'm not sure what this phrase means "genes by symbols/entities". Can you please clarify?

Sorry, that was a typo. Was supposed to read "genes *be* symbols/entities". This way you can get them from a sympy expression using `expression.atoms`. Basically the Gene class would now derive from optlang.symbolics.Symbol.

> Use optlang.symbolics instead of direct sympy, which should be agnostic to sympy, symengine (or any replacements in the future)

That would be great! But if there is something missing from optlang.symbolics it would probably not be a blocking issue if you use sympy for those parts for now. This could be fixed in an upcoming PR.

It's up to you whether you want to continue here or open a new PR.

akaviaLab:
Hi @cdiener 
It has been a long time, and I went through some things. They're mostly done now, and I'm planning to tackle this again, so I've been looking at optlang.symbolics.Symbol

It looks like I should make Gene be a dual-parent class, something like
import optlang.symbolics.Symbol as Symbol

class Gene(Species, Symbol): or class Gene(Symbol, Species)

I don't see how having gene only inherit from Symbol (class Gene(Symbol)) would work well. 

If I've got this right, I'll start working on it, with everything else discussed (including all imports) and open a different PR when it is somewhat ready.

If Gene should be class Gene(Symbol): how should it deal with reactions and so forth?

Thank you

cdiener:
Yeah, that sounds good. An issue that we're having though is that cobra should be able to use sympy *or* symengine and [symengine does not support parsing of logical expressions](https://github.com/symengine/symengine.py/issues/334). So we can not use symengine to parse logical expressions which defeats the purpose a bit.

So it seems this needs to be delayed a bit.

akaviaLab:
Okay.
I could have it be something like this

# When https://github.com/symengine/symengine.py/issues/334 is resolved,
change it to optlang.symbolics.Symbol
 from sympy import Symbol as Symbol

Would that work?

cdiener:
That would work. Though I would probably benchmark the parsing and eval with what we have now. We probably would not want everything to be 10x slower than before and sympy is sometimes not super fast. I can run this and report back.

akaviaLab:
See https://github.com/opencobra/cobrapy/discussions/1128 for further discussion

PR #1017, opened 2020-10-28
refactor: cobra.sampling.hr_sampler

synchon:
* [x] tests added/passed

This PR refactors `cobra.sampling.hr_sampler` module to improve Python 3.6+ compatibility.

### DIFF
--- a/src/cobra/sampling/hr_sampler.py
+++ b/src/cobra/sampling/hr_sampler.py
@@ -2,6 +2,7 @@
 
 import ctypes
 import logging
+from abc import ABC, abstractmethod
 from multiprocessing import Array
 from time import time
 from typing import TYPE_CHECKING, NamedTuple, Optional, Tuple
@@ -120,7 +121,7 @@ def shared_np_array(
     return np_array
 
 
-class HRSampler:
+class HRSampler(ABC):
     """
     The abstract base class for hit-and-run samplers.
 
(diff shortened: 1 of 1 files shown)


src/cobra/sampling/hr_sampler.py
```diff
@@ -201,9 +189,48 @@ def __init__(self, model, thinning, nproj=None, seed=None):
...
+            - "variable_bounds" is a compound matrix `[lb ub]` containing the
+              lower and upper bounds for all variables.
+            - "nullspace" is a matrix containing the nullspace of the equality
+              constraints. Each column is one basis vector.
+            - "homogeneous" is a boolean which indicates whether the sampling
+              problem is homogeneous, e.g., whether there exists no non-zero
+              fixed variables or constraint.
+
+        """
+        Problem = NamedTuple(
```
cdiener: This won't work. This hast to be defined outside of a class so you can pickle it. See comment on step() below.

cdiener (changes requested):
Looks good but introduced bugs. Just make sure you maintain the membership/non-membership structure as before. in the long run, we should probably change to use sparse arrays instead of memory sharing which is a bit unstable anyways. This would also be faster.

src/cobra/sampling/hr_sampler.py
synchon: Okay I see. A strange is that the tests are passing for Python 3.6 and 3.7 and is similar on my local system too. Any idea why it's not failing?

cdiener: On Linux, they should pass because of fork vs. spawn when running multiprocessing. The multiprocess backend changed quite a bit between minor versions on Mac and Windows so I would not be too surprised about it working on one version and not the other.

synchon: That's a fair point. Okay I'll revert it back.

synchon: Done.

synchon:
@cdiener Does it make sense to move `step()` into a separate module like `cobra.sampling.util` ? I propose this as it's used for both ACHR and OptGP samplers. If it makes sense, then I'll open another PR after closing this to address it.

cdiener:
Yep, make sense. Would maybe call it "core" and not "util" since this is the low-level sampling stepper.

cdiener (approved):
Thanks, looks great now.

synchon:
@Midnighter Shall we merge this?

Midnighter:
I'm not the only one who can merge :sweat_smile: Thank you for your continued effort to modernize the codebase! Really appreciate it.

synchon:
I thought it's best to run it by you before merging. 😁

PR #1022, opened 2020-11-26
Add --cov-append to Toxfile testenv

wanderrful:
Hi there,

Was running tests on your project to see if I could contribute somewhere, when I noticed that Tox is currently configured to create a million dotfiles in the project root folder:

image

We can fix this by just adding the `--cov-append` to the command in your Toxfile.

image

I know that the million dotfiles will go away once all of the testenvs are finished, but it can still be pretty annoying. Especially when running tasks in parallel!

It's just a very minor, quality-of-life tweak.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,7 @@ deps=
     pytest-raises
     jsonschema
 commands =
-    pytest --cov=cobra --cov-report=term {posargs: --benchmark-skip}
+    pytest --cov=cobra --cov-report=term --cov-append {posargs: --benchmark-skip}
 
 [testenv:isort]
 skip_install = True


wanderrful:
Also, separate topic -- FYI your current numpy version (1.19.4) doesn't work on Windows. The advice I found was to revert it to version 1.19.3, which worked for me. I was able to get Tox to work by manually adjusting the version with its Toxenv folder's `pip` copy.

Midnighter:
Hey @wanderrful, thank you for making a contribution and planning to contribute more!

> Especially when running tasks in parallel!

Are you sure that there are no race conditions when coverage is recorded in parallel? I thought that was the point of all of those files in the first place.

> Also, separate topic -- FYI your current numpy version (1.19.4) doesn't work on Windows. The advice I found was to revert it to version 1.19.3, which worked for me. I was able to get Tox to work by manually adjusting the version with its Toxenv folder's pip copy.

We don't require the latest version of numpy but we also don't forbid it. So by default the latest version would be installed, yes. I also notice that the test suite passes fine for Windows where 1.19.4 was installed. Does this only affect a certain Python version maybe?

Somehow your change to the test command seems to not apply the `--benchmark-skip` now...

A few weird things going on here.

Midnighter:
Hey @wanderrful, thank you for making a contribution and planning to contribute more!

> Especially when running tasks in parallel!

Are you sure that there are no race conditions when coverage is recorded in parallel? I thought that was the point of all of those files in the first place.

> Also, separate topic -- FYI your current numpy version (1.19.4) doesn't work on Windows. The advice I found was to revert it to version 1.19.3, which worked for me. I was able to get Tox to work by manually adjusting the version with its Toxenv folder's pip copy.

We don't require the latest version of numpy but we also don't forbid it. So by default the latest version would be installed, yes. I also notice that the test suite passes fine for Windows where 1.19.4 was installed. Does this only affect a certain Python version maybe?

Somehow your change to the test command seems to not apply the `--benchmark-skip` now...

A few weird things going on here.

dotPiano:
@Midnighter regarding numpy 1.19.4 on WIndows: the problem appeared in the May 2020 Update, I guess you are testing on a different/older build. https://developercommunity.visualstudio.com/content/problem/1207405/fmod-after-an-update-to-windows-2004-is-causing-a.html

wanderrful:
Is there some way I can re-build this project using the Github Actions CI pipeline you guys have setup? I don't see the button in their UI anywhere, unfortunately. Worst case scenario, I can just push a blank commit.

I imagine there was/is something a little off going on with the CI environment.

cdiener:
I triggered a rebuild.

Midnighter:
Sorry for the long wait. I'm not sure that the `--cov-append` flag does what you say it does https://pytest-cov.readthedocs.io/en/latest/readme.html#coverage-data-file

I have not played around with it but to me this means it would require manual deletion of the file during development?

wanderrful:
> 
> 
> Sorry for the long wait. I'm not sure that the `--cov-append` flag does what you say it does https://pytest-cov.readthedocs.io/en/latest/readme.html#coverage-data-file
> 
> I have not played around with it but to me this means it would require manual deletion of the file during development?

That's actually the default behavior even without the `--cov-append`, which is why it's already in the gitignore: https://github.com/opencobra/cobrapy/blob/devel/.gitignore#L202. So I think the second paragraph after the part you're talking about isn't clarifying what the `--cov-append` flag does but rather what the default behavior in general is.

Just ran `tox -e py36` locally on my machine and it leaves behind the same `.coverage` file, except while tests are running it adds the thousand or so temp coverage files to the directory as well as pictured in the original post here.

cdiener:
I think what @Midnighter means is not that it creates the `.coverage` file but that it will combine several test runs. 
So let's say you have a function covered by a test, run the tests and then delete the test and run tests again. Then it would still show the function as covered since it had a test in the first run (but not in the second). This is probably not a problem for CI but would give incorrect coverage on local development if you don't delete the `.coverage` by hand before running tests (the default without the flag).

wanderrful:
Sure, thanks for elaborating. Will close then.

PR #1023, opened 2020-11-26
#972 Cleanup for exceptions.py

wanderrful:
Progress with #972.  Cleans up Python 2.x references for `exceptions.py`.

Also, while I was in there, I cleaned up the Flake8 complaints for the file, which were just missing Docstrings.

This is pretty much just a cosmetic PR.

### DIFF
--- a/src/cobra/exceptions.py
+++ b/src/cobra/exceptions.py
@@ -3,23 +3,34 @@ import optlang.interface
 
 
 class OptimizationError(Exception):
+    """Exception for Optimization issues."""
+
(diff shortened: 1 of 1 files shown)


Midnighter (approved):
Thanks :smiley:

Midnighter:
Hmm, benchmarks are also run here when they are supposed to be skipped...

Midnighter:
Since these are only changes to docstrings, I'm pulling this in despite the problems with CI.

Issue #1030, opened 2021-01-25
cobrapy does not seem to work with libsbml 5.19


tillea:
#### Problem description

I'm trying to update the cobrapy package in Debian to be up to date with the latest version in the next Debian stable release.  Debian has libsbml version 5.19.0 packaged.  When building against this version I get a lot of test suite errors that seem to be caused by incompatibilities with sbml.  For completeness and to let you know what versions are used here is a full build log:

[python-cobra_0.20.0-1_amd64.zip](https://github.com/opencobra/cobrapy/files/5867974/python-cobra_0.20.0-1_amd64.zip)

You can find the error messages at the end of the log.

Please let me know if I can provide any more detailed information.

Kind regards, Andreas.

cdiener:
So it is kind of expected since the only supported version of libSBML is currently libsbml-experimental 5.18.3 (https://github.com/opencobra/cobrapy/blob/devel/setup.cfg#L65). @matthiaskoenig would know more about how feasible it would be to update the dependency.

tillea:
On Tue, Jan 26, 2021 at 02:31:27PM -0800, Christian Diener wrote:
> So it is kind of expected since the only supported version of libSBML is currently libsbml-experimental 5.18.3 (https://github.com/opencobra/cobrapy/blob/devel/setup.cfg#L65).
Sure.  I have read this.  However, at the time of writing this restriction libsbml 5.19 was probably not released yet and thus it was not clear whether there might be any conflicts or not.  Thus I was simply trying.

> @matthiaskoenig would know more about how feasible it would be to update the dependency.
As I said this is simply needed if cobrapy should be delivered with the next Debian release.  I would like to do so but I have no idea how hard the adaption might be.

Kind regards, Andreas.

Midnighter:
What is the deadline for wrapping up the packaging for the Debian release?

tillea:
On Wed, Jan 27, 2021 at 02:20:42AM -0800, Moritz E. Beber wrote:
> What is the deadline for wrapping up the packaging for the Debian release?
According to the [Freeze Timeline](https://release.debian.org/bullseye/freeze_policy.html) its 2021-02-12 - but we probably need some days for the packaging and testing.  So well, to say it bluntly, its "right now". ;-)
Kind regards, Andreas.

matthiaskoenig:
Should be no problem to update. Complete backwards compatibility.
Just write 
```
python-libsbml-experimental>=5.19.0
```
[...]

PR #1034, opened 2021-02-06
chore: depend on base libsbml, pin strictly

Midnighter:
This is to finalize #1030 and make a release that can be included in the upcoming Debian release.

@cdiener and @matthiaskoenig, I think you're best positioned to assess. I've briefly described why I think that switching from python-libsbml-experimental to python-libsbml is okay here https://github.com/opencobra/cobrapy/issues/1030#issuecomment-774441844.

Additionally, I've pinned the version exactly. This is because in the past there has been a delay between a new libsbml release and the availability of wheels. This has caused unnecessary headache for environments where building from sources is not an option and our considerable portion of Windows users.

If possible, I'd like to get this in and make a release today. So that @tillea has a chance to prepare the Debian package.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,7 +62,7 @@ install_requires =
 	optlang <1.4.6
 	pandas ~=1.0
 	pydantic ~=1.6
-	python-libsbml-experimental >=5.19.0
+	python-libsbml ==5.19.0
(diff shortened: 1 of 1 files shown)


cdiener (approved):
Makes sense to me.

PR #1047, opened 2021-02-23
Fix the doc building on RTD

cdiener:
* [ ] fix #(issue number)
* [ ] tests added/passed

## Description

Docs were not built on RTD since version 0.17.0. This switches the documentation builder to the normal installation and removes the mock code which is not necessary anymore. A preview of the latest version can be found at https://cobrapy-cdiener.readthedocs.io/en/latest/ .

## After merge

We need to switch RTD to use the devel branch and not the deleted master branch. I can do that.

### DIFF
--- a/documentation_builder/conf.py
+++ b/documentation_builder/conf.py
@@ -14,7 +14,6 @@
 
 import sys
 from os.path import dirname, join
-from unittest.mock import Mock
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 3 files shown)


Midnighter (approved):
Nice, I've been wanting to get rid of the module mocking for a long time. As long as there are libSBML wheels, installing the project on RTD should be smooth.

Midnighter:
I wonder if we should merge these changes into branches with older versions as well so that we can put documentation online for those... 🤔

cdiener:
RTD allows to build specific versions. Don't know if you can pick tags there though...

EDIT: You can activate older versions at https://readthedocs.org/projects/cobrapy/versions/ @Midnighter.

Midnighter:
Sure, we can activate them but I suspect the docs won't build without these changes.

cdiener:
It might work. I activated package installation on RTD so it should not complain about missing packages. The other changes here was just to remove the mocked modules but having them in should not break the build per se.

PR #1050, opened 2021-03-02
Fix #1049, updated wording pfba

### LINKED ISSUE #1049: Change wording of pFBA description in the docs
### Checklist

- [x ] There are [no similar issues or pull requests](https://github.com/opencobra/cobrapy/issues) for this yet.

### Is your feature related to a problem? Please describe it.

I want it to be clear what objective value means, and what to expect from the pFBA output by reading this section [https://cobrapy.readthedocs.io/en/latest/simulating.html#Running-pFBA](url).

I think the wording of what to expect from the pFBA output is confusing. Specifically, the phrase "objective value" is used a bit sloppily, because it's used to mean two completely different things: 
1. The value of the internal function's objective (which differ between FBA and pFBA)
2. The value of the objective for standard FBA.

I would recommend to change this sentence "These functions should give approximately the same objective value." to something like:

> "These functions will likely give completely different objective values, because the objective value shown from pFBA is defined as `sum(abs(pfba_solution.fluxes.values))`, while the objective value for standard FBA is defined as the flux through the reactions being optimized for (e.g. `fba_solution.fluxes["Biomass_Ecoli_core"]`).
>
> Both pFBA and FBA should return identical results for the flux being optimized for (e.g. `abs(fba_solution.fluxes["Biomass_Ecoli_core"] - pfba_solution.fluxes["Biomass_Ecoli_core"]`)

Redefining the words used in the standard output from running pFBA or FBA, which seems more complicated than simply clarifying behavior in the docs. 

## Additional context

I'm clearly not alone in being confused by the current wording, and my expected behavior for comparing outputs from these two functions:
https://github.com/opencobra/cobrapy/issues/518
https://github.com/opencobra/cobrapy/issues/635

Both of these issues could have been mitigated with these simple doc wording changes.

matthiaskoenig:
* [x] fix #1049

I fixed the wording as suggested in #1049.


Midnighter (approved):
Sounds good to me. There is also a variable for the previous objective value but it's kind of internal so I guess it's better not to document it.

PR #1059, opened 2021-03-21
refactor: cobra.sampling.optgp

synchon:
* [x] tests added/passed

This PR refactors `cobra.sampling.optgp` module to improve Python 3.6+ compatibility.


cdiener (changes requested):
Would be great if that would also change all instances of multiprocessing.Pool to use the context manager like:

```python
with mp.Pool(...) as pool:
    ...
```
This was not supported in Python 2.

synchon:
Yes I kept that as I had another PR to replace the old usage of `multiprocessing.Pool` with the use of context managers. There are a few places across the code base which will need this change, hence I would like to take it all in one go. Is that okay?

synchon:
Latest commit is to rebase on current `devel`.

cdiener (approved):
Makes sense, go ahead.

PR #1063, opened 2021-03-23
refactor: remove six usage

synchon:
* [x] tests added/passed

This PR removes `six` usage from the code base and updates installation config files.

### DIFF
--- a/src/cobra/test/test_core/test_dictlist.py
+++ b/src/cobra/test/test_core/test_dictlist.py
@@ -9,7 +9,6 @@ from copy import copy, deepcopy
 from pickle import HIGHEST_PROTOCOL, dumps, loads
 
 import pytest
-from six.moves import range
 
 from cobra.core import DictList, Object
 
--- a/src/cobra/test/test_flux_analysis/test_variability.py
+++ b/src/cobra/test/test_flux_analysis/test_variability.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import
 
 import numpy as np
 import pytest
-from six import iteritems
 
 from cobra.exceptions import Infeasible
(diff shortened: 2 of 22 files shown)


synchon:
The test is failing due to network issue when performing `test_web` tests.

src/cobra/test/test_core/test_dictlist.py
```diff
@@ -9,7 +9,6 @@
 from pickle import HIGHEST_PROTOCOL, dumps, loads
 
 import pytest
-from six.moves import range
```
Midnighter, line 12: Was `range` an unused import? It's not removed from any statement in the module.

src/cobra/test/test_flux_analysis/test_variability.py
```diff
@@ -6,7 +6,6 @@
 
 import numpy as np
 import pytest
-from six import iteritems
```
Midnighter, line 9: Again an unused import?

Midnighter (approved):
Looks good! It's time we enabled flake8 so that unused imports and variables would be caught immediately.

src/cobra/test/test_core/test_dictlist.py
synchon: I think it shadowed Python 3's `range`.

src/cobra/test/test_flux_analysis/test_variability.py
synchon: Yes this one was unused.

synchon:
I would say we wait till we have completely fixed all the existing `flake8` issues. This would help to catch only the future issues and not the old ones.

PR #1064, opened 2021-03-23
feat: add a dispatch to continuously update COBRApy

Midnighter:
In @jonrkarr 's group, they have created https://biosimulators.org/, a platform that facilitates performing computational experiments with different modeling tools. They have already prepared a wrapper around cobrapy in order to make it available on the platform https://github.com/biosimulators/Biosimulators_COBRApy.

In this PR I suggest to dispatch a workflow on the https://github.com/biosimulators/Biosimulators_COBRApy repo, whenever a new version of cobrapy gets released. I hope that this is agreeable. I will open another discussion on whether and in how far we want to support this in cobrapy.

### DIFF
--- a/.github/workflows/biosimulators_dispatch.yml
+++ b/.github/workflows/biosimulators_dispatch.yml
@@ -6,12 +6,7 @@ on:
 
 jobs:
   dispatch:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: true
-      matrix:
-        os: [ubuntu-latest]
-        python-version: [3.8]
+    runs-on: ubuntu-latest
 
     steps:
       - name: Get tag
--- a/.github/workflows/biosimulators_dispatch.yml
+++ b/.github/workflows/biosimulators_dispatch.yml
@@ -1,0 +1,26 @@
+name: BioSimulators Dispatch
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  dispatch:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [3.8]
+
+    steps:
+      - name: Get tag
+        id: tag
+        run: echo "::set-output name=version::${GITHUB_REF#refs/tags/}"
+      - name: Dispatch new release version to BioSimulators
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: Continuous integration
+          repo: biosimulators/Biosimulators_COBRApy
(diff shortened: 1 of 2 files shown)


cdiener:
For better encapsulation, I would prefer this in its own YAML. There is a way to trigger an action if another action finishes successfully I think. Or you could make it dependent on a release being created. Basically the release workflow should still succeed even if that step fails.

synchon:
> Basically the release workflow should still succeed even if that step fails.

This makes sense. I guess if that can be assured then a single YAML is also okay.

cdiener:
Personally, I would still put it in its own file. This is not directly related to cobrapy so it makes sense to have it somewhat separate. Also easier to remove or re-add it this way (just moving a file vs editing the workflow).

synchon:
Yeah that does make sense.

Midnighter:
> Personally, I would still put it in its own file. This is not directly related to cobrapy so it makes sense to have it somewhat separate. Also easier to remove or re-add it this way (just moving a file vs editing the workflow).

Right now it would be easier for me to implement ignoring the outcome status of this step rather than configuring the right triggers to start a new workflow. This step happens at the end of the release job so it's quite easy to remove.

Midnighter:
Well, I guess this would do it.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

Midnighter:
No idea how to test this but I guess it should do the trick. Let me know what you think @cdiener

cdiener (approved):
Much better, thanks :+1:

jonrkarr:
@Midnighter thanks for exploring this with COBRApy so quickly.

Once the changes are integrated in, you'll be able to test the action from here https://github.com/opencobra/cobrapy/actions.
- Select the action
- There will be a button "Run workflow"

You'll be able to check the downstream action here: https://github.com/biosimulators/Biosimulators_COBRApy/actions/workflows/ci.yml
https://github.com/biosimulators/Biosimulators/issues

For example, here's the downstream that I triggered when @Midnighter and I chatted:
https://github.com/biosimulators/Biosimulators_COBRApy/actions/runs/680257574
https://github.com/biosimulators/Biosimulators/issues/145

Midnighter:
We could actually use the same mechanism for the website deployment. Seems a bit cleaner to trigger a release there than to commit and push changes from this repo's workflow.

cdiener:
Agreed!

Midnighter:
I had to set the correct remote reference where the dispatched workflow should run. See https://github.com/opencobra/cobrapy/commit/4f1214a123924f78abbb0e4b3284accc33e8debb but now it works!

PR #1067, opened 2021-03-23
update: add a few more stuff to pyproject.toml

synchon:
* [x] tests added/passed

This PR adds the `build-backend` explicitly and also adds a few more options for `black`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,30 @@ requires = [
   'setuptools>=40.6.0',
   'wheel'
 ]
+build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 88
 python-version = ['py36']
+include = '\.pyi?$'
+exclude = '''
+
+(
+  /(
+      \.eggs         # exclude a few common directories in the
+    | \.git          # root of the project
+    | \.hg
+    | \.mypy_cache
+    | \.tox
+    | \.venv
+    | _build
+    | buck-out
+    | build
+    | dist
+    | documentation_builder
+    | release-notes
+    | scripts
+    | benchmarks
+  )/
+)
+'''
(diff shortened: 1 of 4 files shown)


synchon:
Rebased on latest `devel`.

Midnighter:
I agree with setting the build-backend but are the includes and excludes worth the trouble of maintaining the list?

synchon:
I have been using it like this for a few projects and seems to cause no mental overhead. Also, if you see only the last entries are added by me, except those, rest are taken from `black`'s docs.

Midnighter:
I usually run black as `black src` directly so I've never had the need to exclude directories. And anyway only Python files are considered as far as I know. So I don't see the need personally. Does this configuration help when running black integrated in VSCode or something?

synchon:
I don't think it does some trick for editor integration but I might be wrong. I usually follow this as the official docs put this as an example. Also, makes the lookup hierarchy a bit more explicit.

synchon:
Rebased on latest `devel`.

pyproject.toml
```diff
@@ -3,7 +3,30 @@ requires = [
...
+    | \.mypy_cache
+    | \.tox
+    | \.venv
+    | _build
+    | buck-out
+    | build
+    | dist
+    | documentation_builder
+    | release-notes
+    | scripts
```
Midnighter: Actually it would be nice if the Python scripts here are maintained with black.

Midnighter (approved):
I don't see the real benefit of adding this configuration but I also don't see a disadvantage. So fine with me.

pyproject.toml
synchon: Okay I'll remove `documentation_builder` and `scripts` from the list. `dist` is usually left out and `release-notes` doesn't have any Python files.

Midnighter: Sorry, the UI was unspecific there. The comment was only on the line with `scripts`.

synchon: No worries.

PR #1070, opened 2021-03-26
Speed up Windows multiprocessing via file dump

### LINKED ISSUE #997: Speed up FVA on Windows.
Hi,

FVA on Windows is very slow, even for small models. This is due to known limitations of the `multiprocessing` library, but it would still be nice to minimize the problem and reduce the runtime. On my system with 12 virtual cores FVA takes 44s for the `textbook`. With some profiling it appears that most of the time is spent initializing the process pool, in particular in `ForkingPickler.dump()`.

```
import cProfile
from cobra.flux_analysis.variability import flux_variability_analysis
import cobra.test

cProfile.run('flux_variability_analysis(model)')
```

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    0.000    0.000   42.796   21.398 pool.py:183(__init__)
       48    0.003    0.000   42.736    0.890 reduction.py:58(dump)
...
```

Looking the code in `variability.py` I see two possible improvements:
1. Two pools are created for each FVA call. One for maximization, one for minimization. Can this be avoided?
2. The long time spent in `dump` is probably due to the size/complexity of the model class, which is copied to each process. Would it be possible to pass a simpler representation (for example only the constraint matrices)? This seems ideal, but maybe too complicated given the structure of cobrapy.

Cheers,
Mattia

Midnighter:
* [x] fix #997
* [ ] tests added/passed

### DIFF
--- a/src/cobra/flux_analysis/deletion.py
+++ b/src/cobra/flux_analysis/deletion.py
@@ -1,6 +1,5 @@
 from functools import partial
 from itertools import product
-from multiprocessing import Pool
 from typing import List, Set, Union
 
 import pandas as pd
@@ -10,6 +9,7 @@ from cobra.core import Configuration, Gene, Reaction
 from cobra.flux_analysis.moma import add_moma
 from cobra.flux_analysis.room import add_room
 from cobra.manipulation.delete import find_gene_knockout_reactions
+from cobra.util import ProcessPool
 from cobra.util import solver as sutil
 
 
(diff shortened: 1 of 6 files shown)


Midnighter:
I'm not very happy with this yet. The `multiprocessing.Pool` is also used in other modules. It might be worth introducing our own pool class as @dotPiano has shown in the linked issue. However, I would like to keep the pool management via a context and I wasn't quite up to the task this evening. Let's see how this goes and I can always refactor later.

cdiener (commented):
Looks good. Speed-up seems to be pretty minor going by the tests, but only a small part of those is FVA. I would add a note in the comments that this should be removed once that performance problem is gone in CPython. In the long run, we probably don't want to fix general Python performance problems in cobrapy.

cdiener (approved):
:+1:

Midnighter:
Please take another look @cdiener, I have refactored the pool code into a class and use that where previously the `multiprocessing.Pool` was used.

@dotPiano since this is a more common problem on Windows, we might even move this class into its own package so that you and others can use it in your other projects. Let me know if that is of interest to you.

cdiener:
Before putting work into a separate package I would check if the already existing alternative implementations (loky etc) actually have this limitation as well (the don't use cPickle).

cdiener:
Looks good, so go ahead.

PR #1073, opened 2021-04-05
feat: add community model.

sirno:
Add community module and class to core. WIP.

I have written a community class (as in [stolyar2007](https://pubmed.ncbi.nlm.nih.gov/17353934/)). It merges a list of cobrapy models by creating a deepcopy of each model, renaming the reactions (appending the model id) and connecting them using a additional layer of exchange reactions.

If there is interest in adding this, I can make tests etc... For my purpose it works well in my own code base as is.

* [ ] tests added/passed

### DIFF
--- a/src/cobra/core/community.py
+++ b/src/cobra/core/community.py
@@ -1,0 +1,74 @@
+"""Define Community Model class."""
+
+from copy import deepcopy
+
+from cobra.core.dictlist import DictList
+from cobra.core.model import Model
+
+
+class Community(Model):
+    """Class representation of cobra community model.
+
+    Parameters
+    ----------
+    models : list
+        List of model objects.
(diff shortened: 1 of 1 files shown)


cdiener:
This may be a bit out of scope for cobrapy. Great that it works for you though! There is also [MICOM](https://github.com/micom-dev/micom).

Midnighter:
I would also say that community modelling is out of scope for cobrapy but this could be another nice addition for a cookbook recipe. However, MICOM is a well maintained alternative with many dev quality of life additions.

sirno:
By now I have stepped back from using `cobrapy` for this purpose. It is much slower than other libraries, such as `reframed`. I will close this for that matter.

PR #1082, opened 2021-04-27
Update versions

Midnighter:
* Update and relax version pins slightly
* Test on Python 3.9

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,7 +63,7 @@ install_requires =
 	pandas ~=1.0
 	pydantic ~=1.6
 	python-libsbml ==5.19.0
-	rich ~=8.0
+	rich >=8.0
 	ruamel.yaml ~=0.16
 	swiglpk
 tests_require =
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -63,7 +63,7 @@ install_requires =
 	pandas ~=1.0
 	pydantic ~=1.6
 	python-libsbml ==5.19.0
-	rich ~=8.0
+	rich >=8.0
 	ruamel.yaml ~=0.16
```
cdiener, line 67: I think ruamel.yaml could be relaxed as well. I don't see any issue with 0.17.

cdiener (approved):

setup.cfg
Midnighter: `~=0.16` means `>=0.16;<1.0` since rich is releasing major versions pretty quickly, the same doesn't work there. So `~=8.0` means `>=8.0;<9.0` but the latest is already above 10. The API seems pretty stable, though.

PR #1084, opened 2021-04-30
refactor: cobra.manipulation.modify

synchon:
* [x] tests added/passed

This PR adds typing annotations to `cobra.manipulation.modify` and upgrades it to follow Python 3.6+ standards.

### DIFF
--- a/src/cobra/manipulation/modify.py
+++ b/src/cobra/manipulation/modify.py
@@ -1,17 +1,17 @@
-# -*- coding: utf-8 -*-
-
-from __future__ import absolute_import
+"""Provide functions to modify model components."""
 
 from ast import NodeTransformer
 from itertools import chain
-from warnings import warn
+from typing import TYPE_CHECKING, Dict
 
-from cobra.core import Reaction
-from cobra.core.gene import ast2str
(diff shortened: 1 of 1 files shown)


Midnighter (changes requested):
Thanks for the extension of docstrings and general upgrades.

I have two points that I think should be changed:

1. I think it'd be better to move the `Renamer` class to the top level and pass the `renamer_dict` as an argument to the initialization.
2. Since we're not adding any new attributes or behaviour to the `_GeneEscaper` initialization, I don't see a need to define `__init__` at all.

synchon:
With the latest commits, I hope I have addressed your changes.

Midnighter:
That's strange, we already upgrade pip in the workflow. Shouldn't happen unless tox installs a different version...

Midnighter (approved):

cdiener:
Same problem as in optlang. Need to require pip>=21.1 in safety and disable it for Python 2 :sob:

Midnighter:
I think it needs to go into the tox dependencies, right?

cdiener:
Yep in the tox dependencies for the safety step.

Midnighter:
Another indirect solution https://github.com/Midnighter/structurizr-python/pull/76/files

PR #636, opened 2017-12-05
refactor: remove solvers

Midnighter:
* remove shipped solvers
* remove `solver` argument from functions
* adjust test cases
* remove compiled extension
* remove manybuild factory
* remove ArrayBasedModel

These are very extensive changes and I apologize for putting all of them into too few separate commits. I hope you can bear with me. Let me know if any of your code is broken by these changes.

PR #1097, opened 2021-06-04
refactor: cobra.flux_analysis.reaction

synchon:
* [x] tests added/passed

This PR adds typing annotations for `cobra.flux_analysis.reaction` and its unit tests. It also upgrades the code style for them to maintain Python 3.6+ compatibility.

### DIFF
--- a/src/cobra/flux_analysis/reaction.py
+++ b/src/cobra/flux_analysis/reaction.py
@@ -1,50 +1,52 @@
-# -*- coding: utf-8 -*-
-
-""" functions for analyzing / creating objective functions """
-
-from __future__ import absolute_import, division
+"""Provide functions for analyzing / creating objective functions."""
 
 from operator import attrgetter
+from typing import TYPE_CHECKING, Dict, Union
 from warnings import warn
 
-from cobra.core import Reaction
+from ..core import Reaction
 
 
-def assess(model, reaction, flux_coefficient_cutoff=0.001, solver=None):
-    """Assesses production capacity.
+if TYPE_CHECKING:
+    from cobra import Model
 
-    Assesses the capacity of the model to produce the precursors for the
-    reaction and absorb the production of the reaction while the reaction is
-    operating at, or above, the specified cutoff.
+
+def assess(
+    model: "Model",
+    reaction: Union[Reaction, str],
+    flux_coefficient_cutoff: float = 0.001,
+) -> Union[bool, Dict]:
+    """Assess production capacity.
+
+    Assess the capacity of the `model` to produce the precursors for the
+    `reaction` and absorb the production of the `reaction` while the
+    `reaction` is operating at, or above, the specified
+    `flux_coefficient_cutoff`.
 
     Parameters
     ----------
     model : cobra.Model
-        The cobra model to assess production capacity for
-
-    reaction : reaction identifier or cobra.Reaction
-        The reaction to assess
-
-    flux_coefficient_cutoff :  float
-        The minimum flux that reaction must carry to be considered active.
-
-    solver : basestring
-        Solver name. If None, the default solver will be used.
+        The cobra model to assess production capacity for.
+    reaction : str or cobra.Reaction
+        The reaction to assess.
+    flux_coefficient_cutoff : float, optional
+        The minimum flux that reaction must carry to be considered active
+        (default 0.001).
 
     Returns
     -------
     bool or dict
         True if the model can produce the precursors and absorb the products
-        for the reaction operating at, or above, flux_coefficient_cutoff.
-        Otherwise, a dictionary of {'precursor': Status, 'product': Status}.
-        Where Status is the results from assess_precursors and
-        assess_products, respectively.
+        for the reaction operating at, or above, `flux_coefficient_cutoff`.
+        Otherwise, a dictionary of {'precursor': Status, 'product': Status},
+        where 'Status' is the results from `assess_precursors` and
+        `assess_products`, respectively.
 
     """
     reaction = model.reactions.get_by_any(reaction)[0]
     with model as m:
         m.objective = reaction
-        if _optimize_or_value(m, solver=solver) >= flux_coefficient_cutoff:
+        if _optimize_or_value(m) >= flux_coefficient_cutoff:
             return True
         else:
             results = dict()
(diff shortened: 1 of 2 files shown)


src/cobra/flux_analysis/reaction.py
```diff
@@ -1,50 +1,52 @@
...
+    from cobra import Model
 
 
-def assess(model, reaction, flux_coefficient_cutoff=0.001, solver=None):
-    """Assesses production capacity.
+def assess(
+    model: "Model",
+    reaction: Union[Reaction, str],
+    flux_coefficient_cutoff: float = 0.001,
+) -> Union[bool, Dict]:
```
Midnighter: Yes, passing the solver as an argument is old style but removing the argument is breaking the existing API. I'm a bit hesitant about that.

src/cobra/flux_analysis/reaction.py
```diff
@@ -1,50 +1,52 @@
...
+        Otherwise, a dictionary of {'precursor': Status, 'product': Status},
+        where 'Status' is the results from `assess_precursors` and
+        `assess_products`, respectively.
 
     """
     reaction = model.reactions.get_by_any(reaction)[0]
     with model as m:
         m.objective = reaction
-        if _optimize_or_value(m, solver=solver) >= flux_coefficient_cutoff:
+        if _optimize_or_value(m) >= flux_coefficient_cutoff:
```
Midnighter: I would replace this with a direct call to the method.
```suggestion
        if m.slim_optimize(error_value=0.0) >= flux_coefficient_cutoff:
```

src/cobra/flux_analysis/reaction.py
```diff
@@ -57,45 +59,47 @@ def assess(model, reaction, flux_coefficient_cutoff=0.001, solver=None):
...
+        reactant that is not produced in sufficient quantities.
 
     """
     reaction = model.reactions.get_by_any(reaction)[0]
     result_key = dict(reactants="produced", products="capacity")[side]
     get_components = attrgetter(side)
     with model as m:
         m.objective = reaction
-        if _optimize_or_value(m, solver=solver) >= flux_coefficient_cutoff:
+        if _optimize_or_value(m) >= flux_coefficient_cutoff:
```
Midnighter:
```suggestion
        if m.slim_optimize(error_value=0.0) >= flux_coefficient_cutoff:
```

src/cobra/flux_analysis/reaction.py
```diff
@@ -111,7 +115,7 @@ def assess_component(model, reaction, side, flux_coefficient_cutoff=0.001, solve
             joint_demand += demand_reaction
         m.add_reactions([joint_demand])
         m.objective = joint_demand
-        if _optimize_or_value(m, solver=solver) >= flux_coefficient_cutoff:
+        if _optimize_or_value(m) >= flux_coefficient_cutoff:
```
Midnighter:
```suggestion
        if m.slim_optimize(error_value=0.0) >= flux_coefficient_cutoff:
```

src/cobra/flux_analysis/reaction.py
```diff
@@ -121,7 +125,7 @@ def assess_component(model, reaction, side, flux_coefficient_cutoff=0.001, solve
             # Calculate the maximum amount of the
             with m:
                 m.objective = demand_reaction
-                flux = _optimize_or_value(m, solver=solver)
+                flux = _optimize_or_value(m)
```
Midnighter:
```suggestion
                flux = m.slim_optimize(error_value=0.0)
```

src/cobra/flux_analysis/reaction.py
```diff
@@ -138,82 +142,113 @@ def assess_component(model, reaction, side, flux_coefficient_cutoff=0.001, solve
...
-        dictionary of the required and the produced fluxes for each reactant
-        that is not produced in sufficient quantities.
+        each individual precursor at the specified threshold but not all
+        precursors at the required level simultaneously. Otherwise, a
+        dictionary of the required and the produced fluxes for each
+        reactant that is not produced in sufficient quantities.
 
     """
-    warn("use assess_component instead", DeprecationWarning)
+    warn("Use cobra.sampling.reaction.assess_component() instead.", DeprecationWarning)
```
Midnighter, line 228: I still learn things about the codebase 🙂 It's actually a bit odd to me that `assess_component` is part of the sampling sub-package.

Midnighter (changes requested):
I think, most critical to discuss is the change in API (solver argument) which was not announced as deprecated but also was removed (almost?) everywhere else in the code base. The decision on that will also affect how the test should be structured.

src/cobra/flux_analysis/reaction.py
synchon: You are right. I missed that that #636 was meant to replace explicit passing of solver as an argument. I will revert this change and issue a deprecation warning for the argument.

src/cobra/flux_analysis/reaction.py
synchon: Yeah that is surprising. I think once the deprecations are executed, there will be a better understanding of how to restructure the project better.

src/cobra/flux_analysis/reaction.py
synchon: I see that you have opted to remove the complete usage of `_optimize_or_value()` from the module. In that case, I will issue a deprecation warning for this as well.

Midnighter: Since it was never intended for public use (name starts with `_`), you can remove it right away.

src/cobra/flux_analysis/reaction.py
Midnighter: Actually, you're making a good argument **for** removing it right away 😄

synchon: Having noticed no usage of it, I was tempted to remove it. But, I think based on your point, it's fair to issue a deprecation and then remove it so that it gets no user by surprise and helps us keep a track in the future.

src/cobra/flux_analysis/reaction.py
synchon: That is true. I'm debating whether a private function should enjoy a deprecation notice or not. What do you think about that? 🤔

Midnighter (approved):
Sorry, this got lost in my notifications.

synchon:
No problem. Thanks!

PR #1121, opened 2021-10-07
don't compare objective values for infeasible solutions

cdiener:
* [X] fix unreported issue
* [X] fixes tests themselves
* [ ] add an entry to the [next release](../release-notes/next-release.md) - only relevant for unit tests

I noticed that the io tests fail for the mini model with cplex. The mini model is infeasible and trying to get the same objective value for infeasible solutions is somewhat arbitrary and worked for GLPK just by chance. `compare_models` now checks for the solution status as well.

### DIFF
--- a/src/cobra/test/test_io/conftest.py
+++ b/src/cobra/test/test_io/conftest.py
@@ -4,8 +4,9 @@
 
 from __future__ import absolute_import
 
+from optlang.interface import OPTIMAL
 from os.path import join
-from pickle import dump, load
+from pickle import load
 
 import pytest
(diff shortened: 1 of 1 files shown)


Midnighter (changes requested):
Nice catch! Can I ask you to please use the status from optlang, i.e., `optlang.interface.OPTIMAL` instead of comparing to the string value `"optimal"`? I know it's done like that in many places but one improvement at a time 🙂

Midnighter (approved):
Thanks!

PR #1143, opened 2022-01-08
Bump sbml

### LINKED ISSUE #1142: pip install cobra fails in Python 3.10
Executing 
```bash 
pip install cobra
```
on Arch Linux with the latest Python version (3.10.1) fails. Apparently, `cobra` 0.22.1 depends on `python-libsbml==5.19.0`, which has a bug (see [https://github.com/sbmlteam/libsbml/issues/173](https://github.com/sbmlteam/libsbml/issues/173)) that makes it impossible to install in python 3.10 and that was fixed in `python-libsbml` 5.19.1 or 5.19.2. Anecdotally, forcing the install with a constraints.txt that specifies the 5.19.2 version of that package works with no apparent issues.

Also, installing `cobra` through `pip` kinda works up until `python-libsbml` turn, but importing it gives a 

```python
[...]__init__.py:63: UserWarning: No LP solvers found
  warn("No LP solvers found")
```
error.

Midnighter:
* [x] fix #1142
* [x] Bump libsbml version and start testing Python 3.10
* [ ] tests added/passed

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


cdiener:
As you mentioned already Swiglpk does not support Python 3.10 yet, see the issues there. Need to update the actions there first since they're also broken due to Python 2 not working anymore there. I wanted to get on that but haven't found the time. 

OSQP also has dependencies that don't support 3.10 yet.

cdiener (approved):
👍

PR #1150, opened 2022-02-16
Update package pins and repair the CI

cdiener:

This fixes some issues currently wrecking the CI. Black now has a major version and will complain about some additional formattings now. For now this PR pins black to the previous versions and the codebase will need to be reformatted in the future.

The numpy pin was very old. Is there a mechanism to actually update the version pins periodically? This had led to breakages before and often does not play nice with tox safety.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ parse = (?P<major>\d+)
 	\.(?P<minor>\d+)
 	\.(?P<patch>\d+)
 	(?P<release>[a]*)(?P<num>\d*)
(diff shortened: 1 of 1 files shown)


cdiener:
Okay does not work since it affects all numpy versions. There is an additional issue with older numpy versions having CVEs but not having newer versions available on those older python versions.

Midnighter:
The pin `~=1.13` anyway means `>=1.13, <2.0` so it wouldn't prevent installation of the newest version as would happen on the CI.

PR #1152, opened 2022-02-17
refactor: cobra.io.yaml

synchon:
* [ ] fix #(issue number)
* [ ] tests added/passed

This PR adds typing annotations for cobra.io.yaml and its unit tests. It also upgrades the code style for them to maintain Python 3.6+ compatibility.

### DIFF
--- a/src/cobra/io/yaml.py
+++ b/src/cobra/io/yaml.py
@@ -10,6 +10,8 @@ from .dict import model_from_dict, model_to_dict
 
 
 if TYPE_CHECKING:
+    from io import TextIOBase
+
     from cobra import Model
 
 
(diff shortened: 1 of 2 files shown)


src/cobra/io/yaml.py
```diff
@@ -1,89 +1,111 @@
...
 YAML_SPEC = "1.2"
 
 
-class MyYAML(YAML):
-    def dump(self, data, stream=None, **kwargs):
-        inefficient = False
+class CobraYAML(YAML):
+    """Define custom subclass for YAML I/O."""
+
+    def dump(self, data: Dict, stream: Optional[Any] = None, **kwargs: Any) -> str:
```
Midnighter: I think the type of `stream` can be https://docs.python.org/3/library/io.html#io.TextIOBase rather than `Any`.

synchon: What about we use something more specific: https://docs.python.org/3/library/io.html#io.StringIO ?

Midnighter: My thinking was that if someone passes a stream, it will not be `StringIO` but some other text stream. So it's a bit more broad to go with `TextIOBase`.

synchon: Okay sounds fair. I'll change it.

Midnighter (approved):
Thank you 🙂

PR #1156, opened 2022-02-24
Modify CI configuration

### LINKED ISSUE #1155: How to handle numpy on older Python version?
We have a problem with the vulnerability in numpy. Numpy has stopped supporting 3.6 and 3.7 so even when the safety database is updated, the numpy versions installed on those older Python versions will be an outdated one and never pass.

How should we handle this?

* An easy option is to remove the safety check on those versions
* Ignore this specific vulnerability in the safety command
* Stop supporting 3.6 & 7 ourselves

Midnighter:
* [x] fix #1155 
* [ ] tests added/passed

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,6 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ['3.10']
+    timeout-minutes: 60
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ on:
     - devel
 
 jobs:
-  test:
+  lint:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ['3.10']
-    timeout-minutes: 60
+    timeout-minutes: 30
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,45 @@
...
+        os: [ubuntu-latest]
+        python-version: ['3.10']
+    timeout-minutes: 60
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
```
cdiener, line 31: Maybe add `cache: 'pip'` as well to make it a bit faster? See [here](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/).

cdiener (commented):
Like you said. The safety check is not really noticeable at all now...

.github/workflows/lint.yml
Midnighter: I'm actually not worried about the time it takes to install dependencies but the additional nodes that are required for a separate safety workflow.

cdiener: If the nodes would run for only half the time that would be fine though. But that wouldn't happen here since the time is spent in the other tox workflows and tox does not support caching.

cdiener: But yeah, it's fine for now I think.

cdiener (approved):

Midnighter:
@cdiener I've split safety into a separate workflow that only runs on push to stable & devel as well as tags. Let's try that out for a bit?

cdiener:
Oh yeah, already approved it anyways. I think either option is okay but this one looks great. Go ahead!

akaviaLab:
One thing about this. It seems like the timeout is set too low on some of the checks (not the lint one). I have had multiple recent PR that pass lint, but some of the tests are cancelled due to timeout. For example, see #1163. It looks like the tests failed, but actually they were just cancelled.
Can we increase the timeout?

PR #1158, opened 2022-02-25
Add sympy type conversions to GPR

akaviaLab:
* [X] description of feature
Added as_symbolic() to the GPR class. Also added eq to the GPR class which uses symbolic logic to compare them. Added tests for symbolic and tests for equality.

There are three methods to get to symbols, as_symbolic(), parse_gpr_sympy_ast_visitor and parse_sympy_via_names. If any of them seems better code, I can shift to it. 
The speed seems to be about the same, but I haven't managed to run a benchmark on a bigger model. If you want to modify test_gpr_equality_benchmark to use large_model and then tell me if one of the methods is faster, I would appreciate it.

Once approved, I plan to add one commit that will remove the extra methods, so there will only be one.

* [X] tests added
test_gpr_as_symbolic
test_gpr_as_symbolic_boolean
test_gpr_equality
test_gpr_equality_with_bolean_logic
test_gpr_inequality_boolean
test_gpr_symbolism_benchmark
test_gpr_equality_benchmark
test_gpr_symbolic_methods (not necessary when only one method exists)
Added as_symbolic to GPR which uses sympy. Added eq to GPR which compares the logic using sympy.


src/cobra/core/gene.py
```diff
@@ -559,8 +655,99 @@ def _repr_html__(self):
...
+        Calls _symbolic_gpr()
+        """
+        # noinspection PyTypeChecker
+        if names:
+            GPRGene_dict = {gid: Symbol(names[gid]) for gid in self.genes}
+        else:
+            GPRGene_dict = None
+        return self._symbolic_gpr(self, GPRGene_dict=GPRGene_dict)
+
+    def _symbolic_gpr(
```
cdiener, line 661: That's the one I would prefer. Should be the fastest as well since it directly converts from the pre-parsed tree.

src/cobra/core/gene.py
```diff
@@ -22,6 +22,12 @@
 from typing import FrozenSet, Iterable, Set, Tuple, Union
 from warnings import warn
 
+# When https://github.com/symengine/symengine.py/issues/334 is resolved, change it to
+# optlang.symbolics.Symbol
+from sympy import Symbol
+from sympy.logic.boolalg import And as sp_And
```
cdiener: Better do somehting like `import sympy.logic.boolalg as spl` and then use `spl.Or` and `spl.And`. Just makes it more explicit where those are coming from.

cdiener: Also check with isort. Looks like that import order would not pass the linter.

src/cobra/core/gene.py
```diff
@@ -122,6 +128,96 @@ def parse_gpr(str_expr: str) -> Tuple:
...
+        return Symbol("")
+    sympifier = GPRSympifier()
+    sympy_exp = sympifier.visit(tree)
+    return sympy_exp
+
+
+def parse_sympy_via_names(gpr_to_sympy):
+    sympy_str = ast2sympy_str(gpr_to_sympy)
+    if sympy_str:
+        return eval(sympy_str)
```
cdiener: Didn't you say that this will fail for larger GPRs anyways?

src/cobra/core/gene.py
```diff
@@ -559,8 +655,99 @@ def _repr_html__(self):
...
+                    )
+                elif isinstance(op, And):
+                    # noinspection PyTypeChecker
+                    sym_exp = sp_And(
+                        *[self._symbolic_gpr(i, GPRGene_dict) for i in expr.values]
+                    )
+                else:
+                    raise TypeError("unsupported operation " + op.__class__.__name)
+                return sym_exp
+            elif expr is None or (isinstance(expr, list) and len(expr) == 0):
```
cdiener: might even do `elif not expr:`. This will behave the same for the two cases and catch some others.

src/cobra/core/gene.py
```diff
@@ -559,8 +655,99 @@ def _repr_html__(self):
...
+                return Symbol("")
+            else:
+                raise TypeError("unsupported operation  " + repr(expr))
```
cdiener: Should probably be `TypeError("Unsupported Expression " + repr(expr)`.

cdiener (commented):
Looks pretty good already I would say. Some small comments. Also, testing boolean expressions for equivalence is apparently NP-complete so there will be no efficient way to do so anyway. So we should expect this to be slow. I think we would also want a `GPR.from_sympy` classmethod at some point. That would enable us to do things like getting the CNF with sympy etc.

src/cobra/core/gene.py
akaviaLab: Ran isort on the modified version

src/cobra/core/gene.py
akaviaLab: No. I'm pretty sure that this version will not crash, since the crashing version was using sympy.parse_expr

cdiener: Oh yeah sorry, I overlooked the previous part. But that seems a bit insecure due to the unprotected eval.

akaviaLab:
Now only safety is failing.

akaviaLab:
Added from_symbolic just to be complete. I tried setting up equality via AST, not via sympy, and it seemed slower, so I've not added it.

src/cobra/io/sbml.py
```diff
@@ -929,7 +929,7 @@ def write_sbml_model(cobra_model, filename, f_replace=F_REPLACE, **kwargs):
     ----------
     cobra_model : cobra.core.Model
         Model instance which is written to SBML
-    filename : string
+    filename : string or filehandle
```
cdiener, line 932: I assume that had nothing to do with the PR? Not really a big change so we can merge it in here as well.

src/cobra/core/gene.py
```diff
@@ -559,8 +565,156 @@ def _repr_html__(self):
...
+        Returns
+        -------
+        GPR:
+            returns a new GPR while setting  self.body as
+            Parsed AST tree that has the gene rules
+            This function also sets self._genes with the gene ids in the AST
+
+        """
+
+        def _sympy_to_ast(
```
cdiener, line 727: Very cool that you actually went through the hassle of parsing the tree. That should be pretty fast.

src/cobra/core/gene.py
```diff
@@ -559,8 +565,156 @@ def _repr_html__(self):
...
+            if sympy_expr.func is spl.Or:
+                return BoolOp(
+                    op=Or(), values=[_sympy_to_ast(i) for i in sympy_expr.args]
+                )
+            elif sympy_expr.func is spl.And:
+                return BoolOp(
+                    op=And(), values=[_sympy_to_ast(i) for i in sympy_expr.args]
+                )
+            elif not sympy_expr.args:
+                return Name(id=sympy_expr.name)
```
cdiener, line 739: Doesn't this also have to populate the genes? Or maybe that is done somewhere else implicitly.

cdiener (commented):
Looks good, let's just wait for the CI fix to be merged so this can be rebased on top. That will fix the safety issues. using the sympy equality is fine here since that is the only real equivalence test we have available. I think the AST equality just checks for the exact same expression which is probably no what we want anyway.

src/cobra/core/gene.py
akaviaLab: Thank you. The only other alternative would have been to print SYMPY as str and then parse the str as ast which would have been slow.

src/cobra/core/gene.py
akaviaLab: gpr.update_genes() at the end (line 701) repopulates the genes. I guess I could do it explicitly in _sympy_to_ast, but that would complicate the function and I don't see the point. Might be a bit faster, but not much.

src/cobra/io/sbml.py
akaviaLab: Yeah, nothing to do with PR.

akaviaLab (commented):
When I say I made an AST equality, I tried to do something like sympy.equals - compare the truth tables, and return false if they disagree. It worked correctly, but it was a bit slower than sympy. They probably optimized it.

cdiener:
> When I say I made an AST equality, I tried to do something like sympy.equals - compare the truth tables, and return false if they disagree. It worked correctly, but it was a bit slower than sympy. They probably optimized it.

That makes sense. From what I read it looks like `sympy.equals` tries to find a set of transformations so that both sides become equivalent which should be more efficient than truth tables on average.

akaviaLab:
Rebased on updated devel

src/cobra/test/test_core/test_gpr.py
```diff
@@ -48,8 +48,8 @@ def test_one_gene_gpr() -> None:
 
 # Gets an iterable of all combinations of genes except the empty list. Used to
 # evaluate AND gprs
-def powerset_ne(iterable):
-    """powerset_ne([1,2,3]) --> (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)"""
+def powerset_ne(iterable: Iterable) -> Iterator:
```
Midnighter: Is it of gene?

```suggestion
def powerset_ne(iterable: Iterable[Gene]) -> Iterator[Gene]:
```

src/cobra/test/test_core/test_gpr.py
```diff
@@ -84,8 +84,8 @@ def test_and_gpr(gpr_input, num_genes, gpr_genes, gpr_output_string) -> None:
 
 # Gets an iterable of all combinations of genes except the complete list and the empty
 # list. Used to evaluate OR gprs
-def all_except_complete(iterable):
-    """all_except_complete([1,2,3]) --> (1,) (2,) (3,) (1,2) (1,3) (2,3)"""
+def all_except_one(iterable: Iterable) -> Iterator:
```
Midnighter: As above.

Midnighter (approved):
Looks good.

src/cobra/test/test_core/test_gpr.py
akaviaLab: Same answer.

src/cobra/test/test_core/test_gpr.py
akaviaLab: No. Currently this is an Iterable of strings, used on a set.
So I guess it could be Iterable[str], but I'm not sure that is very helpful. Do you want me to change it to that?

Midnighter: It's true that it's just in the tests but I still think it's helpful, yes.

akaviaLab: Done

akaviaLab:
Can this be merged before something changes in the tree and I have to fight with git about rebasing please?

Midnighter:
If you're happy @cdiener this can be merged from my side.

PR #1161, opened 2022-03-03
Refactor core

### LINKED ISSUE #972: Upgrade code base to pure Python 3.6+
We decided some time ago to drop Python 2.7 and only support 3.6 and later versions. Our legacy codebase offers many opportunities for cleaning up old syntax.

---

## Check List for Upgrading Modules

* [ ] remove file encoding shebang `# -*- coding: utf-8 -*-`
* [ ] carefully evaluate imports and remove compatibility imports (typically remove all imports from `__future__`)
* [ ] remove `object` from class inheritance
* [ ] remove `six` and `future` usage
* [ ] use f-strings for string interpolation
* [ ] add type annotations
* [ ] format code with `isort` and `black`
* [ ] ensure that the code you work on passes flake8 as defined in the `tox.ini` (that means with plugins `flake8-bugbear` and `flake8-docstrings`
* [ ] update existing and add missing docstrings as reported following https://numpydoc.readthedocs.io/en/latest/format.html
* [ ] make use of context blocks (`with ...`) where possible (in particular for `multiprocessing.Pool`)

akaviaLab:
* [X] fix #972 for gene.py, formula.py
Not sure I did this right, so please let me know.

### DIFF
--- a/src/cobra/core/dictlist.py
+++ b/src/cobra/core/dictlist.py
@@ -2,7 +2,16 @@
 
 import re
 from itertools import islice
-from typing import Any, Callable, Pattern, Union, Iterable, Tuple, Iterator, Type
+from typing import (
+    Any,
+    Callable,
+    Iterable,
+    Iterator,
+    Pattern,
+    Tuple,
+    Type,
+    Union,
+)
 
 from numpy import bool_
 
@@ -64,7 +73,7 @@ class DictList(list):
         return [getattr(i, attribute) for i in self]
 
     def get_by_any(self, iterable: list) -> list:
-        """Get a list of members using several different ways of indexing
+        """Get a list of members using several different ways of indexing.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


akaviaLab:
I fixed the issues in gene, and also tried to refactor DictList. I'm not sure I got the type right - should id be only string?

akaviaLab:
Gah. In dictlist.py isort wants the imports from Typing to be one line. Black ones them to be multiple lines. Not sure how to resolve this.

synchon:
Hi @akaviaLab ! Thanks for helping out with the codebase refactoring.

Before going into reviewing the changes, it would be great if you don't reference the main issue #972 here in this PR as doing so will close the issue when this PR is merged. We want to do this at the very last PR which completes the whole refactoring.

You can check out PR #1153 for example, to see the changes we are aiming for. I'm sure you have went through the issue to see the things we want to have. Personally, I think passing `flake8` check is something which helps a lot while refactoring.

It would also really help if you can have one commit for all the changes in one file. We decided to have this unofficial format to help us with cutting the clutter while reviewing.

Edit: I wanted to add that we also follow an unofficial format to have changes of only one file and its corresponding test file in one PR.

akaviaLab:
Okay. I will close this, and redo it with one commit per file, one file and one test per PR, don't reference 972.

PR #1162, opened 2022-03-04
Refactor/io gene

akaviaLab:
This PR adds typing annotations for cobra.core.gene and  test_gpr. test_gene has only one function, and it was typed. I guess test_gene and test_gpr can be folded into one file, but that seems like a different PR. It also upgrades the code style for them to maintain Python 3.6+ compatibility.

PR #1167, opened 2022-03-06
Refactor/core gene

akaviaLab:
This should have all the changes in #1162 including suggestions by @Midnighter and should pass all tests.


Midnighter (approved):
The `**kwargs` should still be there in the `super` call due to multiple inheritance. I'm happy to elaborate on that. However, as it's an unlikely scenario also leaving it as you propose is fine.

PR #1168, opened 2022-03-07
Refactor/metabolite

akaviaLab:
Refactor metabolite.py and test_metabolite.py to be Python 3.6+ compliant including
removing future, typing, removing bugbears, documentation

### DIFF
--- a/src/cobra/core/metabolite.py
+++ b/src/cobra/core/metabolite.py
@@ -50,9 +50,9 @@ class Metabolite(Species):
         id: Optional[str] = None,
         formula: Optional[str] = None,
         name: Optional[str] = "",
-        charge: Optional[str] = None,
+        charge: Optional[float] = None,
         compartment: Optional[str] = None,
-    ):
+    ) -> None:
         """Initialize Metaboblite cobra Species.
 
         Parameters
@@ -312,7 +312,7 @@ class Metabolite(Species):
                 <td><strong>Name</strong></td><td>{format_long_string(self.name)}</td>
             </tr><tr>
                 <td><strong>Memory address</strong></td>
-                <td>{"0x0%x" % id(self)}</td>
+                <td>{id(self):#x}</td>
             </tr><tr>
                 <td><strong>Formula</strong></td><td>{self.formula}</td>
             </tr><tr>
--- a/src/cobra/core/metabolite.py
+++ b/src/cobra/core/metabolite.py
@@ -104,7 +104,7 @@ class Metabolite(Species):
             return self.model.constraints[self.id]
 
     @property
-    def elements(self) -> Union[None, Dict[str, Union[int, float]]]:
+    def elements(self) -> Optional[Dict[str, Union[int, float]]]:
         """Get dicitonary of elements and counts.
 
         Dictionary of elements as keys and their count in the metabolite
--- a/src/cobra/core/metabolite.py
+++ b/src/cobra/core/metabolite.py
@@ -12,10 +12,11 @@ from cobra.util.util import format_long_string
 
 
 if TYPE_CHECKING:
-    from cobra.core import Solution
-    from cobra.summary.metabolite_summary import MetaboliteSummary
     from optlang.interface import Container
     from pandas import DataFrame
+
+    from cobra.core import Solution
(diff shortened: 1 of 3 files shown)


src/cobra/core/metabolite.py
```diff
@@ -40,7 +44,30 @@ class Metabolite(Species):
...
 
-    def __init__(self, id=None, formula=None, name="", charge=None, compartment=None):
+    # noinspection PyShadowingBuiltins
+    def __init__(
+        self,
+        id: str = None,
+        formula: str = None,
+        name: str = "",
+        charge: float = None,
+        compartment: str = None,
```
Midnighter: Always make type annotations `Optional` when the argument can be `None`, please.

src/cobra/core/metabolite.py
```diff
@@ -40,7 +44,30 @@ class Metabolite(Species):
...
+    ):
```
Midnighter:
```suggestion
    ) -> None:
```

src/cobra/core/metabolite.py
```diff
@@ -40,7 +44,30 @@ class Metabolite(Species):
...
+        formula : str
+            Chemical formula (e.g. H2O)
+        name : str
+            A human readable name.
+        charge : float
+           The charge number of the metabolite
+        compartment: str or None
+           Compartment of the metabolite.
+        """
         Species.__init__(self, id, name)
```
Midnighter:
```suggestion
        super().__init__(id=id, name=name)
```

src/cobra/core/metabolite.py
```diff
@@ -49,31 +76,48 @@ def __init__(self, id=None, formula=None, name="", charge=None, compartment=None
...
             the optlang constraint for this metabolite
         """
         if self.model is not None:
             return self.model.constraints[self.id]
 
     @property
-    def elements(self):
-        """Dictionary of elements as keys and their count in the metabolite
-        as integer. When set, the `formula` property is update accordingly"""
+    def elements(self) -> Union[None, Dict[Any, Union[int, float]]]:
```
Midnighter:
```suggestion
    def elements(self) -> Optional[Dict[str, Union[int, float]]]:
```

Midnighter: Or can it be something other than `str`?

src/cobra/core/metabolite.py
```diff
@@ -251,30 +303,23 @@ def summary(self, solution=None, fva=None):
...
             <tr>
-                <td><strong>Metabolite identifier</strong></td><td>{id}</td>
+                <td><strong>Metabolite identifier</strong></td><td>{self.id}</td>
             </tr><tr>
-                <td><strong>Name</strong></td><td>{name}</td>
+                <td><strong>Name</strong></td><td>{format_long_string(self.name)}</td>
             </tr><tr>
                 <td><strong>Memory address</strong></td>
-                <td>{address}</td>
+                <td>{"0x0%x" % id(self)}</td>
```
Midnighter:
```suggestion
                <td>{id(self):#x}</td>
```

Midnighter (changes requested):
Overall, it looks good. I'm in doubt about the `Any` again. Would have to look more thoroughly at the code myself to make a call. Which is also an option. I can add fix ups afterwards if you prefer.

src/cobra/core/metabolite.py
akaviaLab: I think that metabolite is only str, but DictList (as defined) is more complicated. I'll go into detail why I used Any in dictlist in that PR.

src/cobra/core/metabolite.py
```diff
@@ -40,40 +44,80 @@ class Metabolite(Species):
...
             the optlang constraint for this metabolite
         """
         if self.model is not None:
             return self.model.constraints[self.id]
 
     @property
-    def elements(self):
-        """Dictionary of elements as keys and their count in the metabolite
-        as integer. When set, the `formula` property is update accordingly"""
+    def elements(self) -> Union[None, Dict[str, Union[int, float]]]:
```
Midnighter:
```suggestion
    def elements(self) -> Optional[Dict[str, Union[int, float]]]:
```

Midnighter (approved):
LGTM

PR #1170, opened 2022-03-07
chore: relax the libSBML version pin

### LINKED ISSUE #1160: Make libSBML requirement as permissive as possible (e.g., `>= 5.19.0` rather than `== 5.19.2`)
### Checklist

### Is your feature related to a problem? Please describe it.
It would be useful to be able to use COBRApy together with other modeling tools in the same Python environment, including ones which have alternative version requirements, such as for the experimental version (python-libsmbl-experimental) which supports the draft packages such as spatial.

The libSBML requirement is currently `python-libsbml==5.19.2`. If this specific microversion isn't required, ideally this would be changed to something like `python-libsbml>=5.19.0` (or whatever minimum version is required).

Alternatives include forking COBRApy, editing the requirements, and installing it from the source; using a older version; or some other hacking. It would be best to change this at the source, rather than managing it downstream.

## Additional context

N/A

Midnighter:
* [x] fix #1160
* [ ] tests added/passed

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ parse = (?P<major>\d+)
 	\.(?P<minor>\d+)
 	\.(?P<patch>\d+)
 	(?P<release>[a]*)(?P<num>\d*)
-serialize = 
+serialize =
 	{major}.{minor}.{patch}{release}{num}
 	{major}.{minor}.{patch}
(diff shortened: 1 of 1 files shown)


cdiener (approved):
:+1:

PR #1171, opened 2022-03-08
Refactor/core reaction

akaviaLab:
Updated reaction.py and test_core_reaction.py to be Python 3.6+ compliant.
I didn't fully fix memo in deepcopy, because I don't know what it is supposed to be and/or do.

### DIFF
--- a/src/cobra/core/reaction.py
+++ b/src/cobra/core/reaction.py
@@ -26,8 +26,8 @@ from warnings import warn
 
 if TYPE_CHECKING:
     from optlang.interface import Variable
-    from . import Model, Solution
-    from ..summary import ReactionSummary
+    from cobra import Model, Solution
+    from cobra.summary import ReactionSummary
     import pandas as pd
 
 from ..exceptions import OptimizationError
@@ -558,7 +558,7 @@ class Reaction(Object):
             raise RuntimeError(f"reaction '{self.id}' is not part of a model")
         # Due to below all-catch, which sucks, need to reraise these.
         except (RuntimeError, OptimizationError) as err:
-            raise
+            raise err
         # Would love to catch CplexSolverError and GurobiError here.
         except Exception as err:
             raise OptimizationError(
@@ -1487,7 +1487,8 @@ class Reaction(Object):
         verbose: bool
             setting verbosity of function (default True)
         fwd_arrow : AnyStr, optional
-            Str or bytes that encode forward irreversible reaction arrows (default None).
+            Str or bytes that encode forward irreversible reaction arrows (default
+            None).
         rev_arrow : AnyStr, optional
             Str or bytes that encode backward irreversible reaction arrows (default
             None).
--- a/src/cobra/core/reaction.py
+++ b/src/cobra/core/reaction.py
@@ -1,8 +1,4 @@
-# -*- coding: utf-8 -*-
-
 """Define the Reaction class."""
-
-from __future__ import absolute_import, print_function
 
 import hashlib
 import re
@@ -11,10 +7,27 @@ from copy import copy, deepcopy
 from functools import partial
 from math import isinf
 from operator import attrgetter
-from typing import FrozenSet
+from typing import (
+    TYPE_CHECKING,
+    AnyStr,
+    Dict,
+    FrozenSet,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
 from warnings import warn
 
-from future.utils import raise_from, raise_with_traceback
+
+if TYPE_CHECKING:
+    from optlang.interface import Variable
+    from cobra import Model, Solution
+    from cobra.summary import ReactionSummary
+    from pandas import DataFrame
 
 from cobra.core.configuration import Configuration
 from cobra.core.gene import GPR, Gene
@@ -33,12 +46,6 @@ from cobra.util.util import format_long_string
 
 config = Configuration()
 
-# precompiled regular expressions
-# Matches and/or in a gene reaction rule
-and_or_search = re.compile(r"\(| and| or|\+|\)", re.IGNORECASE)
-uppercase_AND = re.compile(r"\bAND\b")
-uppercase_OR = re.compile(r"\bOR\b")
-gpr_clean = re.compile(" {2,}")
 # This regular expression finds any single letter compartment enclosed in
 # square brackets at the beginning of the string. For example [c] : foo --> bar
 compartment_finder = re.compile(r"^\s*(\[[A-Za-z]\])\s*:*")
@@ -49,7 +56,9 @@ _reverse_arrow_finder = re.compile("<(-+|=+)")
 
 
 class Reaction(Object):
-    """Reaction is a class for holding information regarding
+    """Define the cobra.Reaction class.
+
+    Reaction is a class for holding information regarding
     a biochemical reaction in a cobra.Model object.
 
     Reactions are by default irreversible with bounds
@@ -63,20 +72,30 @@ class Reaction(Object):
     ----------
     id : string
         The identifier to associate with this reaction
-    name : string
+    name : string, optional
         A human readable name for the reaction
-    subsystem : string
+    subsystem : string, optional
         Subsystem where the reaction is meant to occur
-    lower_bound : float
+    lower_bound : float, optional
         The lower flux bound
-    upper_bound : float
+    upper_bound : float, optional
         The upper flux bound
+    **kwargs:
+        Further keyword arguments are passed on to the parent class.
     """
 
+    # noinspection PyShadowingBuiltins
     def __init__(
-        self, id=None, name="", subsystem="", lower_bound=0.0, upper_bound=None
-    ):
-        Object.__init__(self, id, name)
+        self,
+        id: Optional[str] = None,
+        name: Optional[str] = "",
+        subsystem: Optional[str] = "",
+        lower_bound: Optional[float] = 0.0,
+        upper_bound: Optional[float] = None,
+        **kwargs,
+    ) -> None:
+        """Initialize the Reaction class."""
+        super().__init__(id, name, **kwargs)
         self._gpr = GPR()
         self.subsystem = subsystem
 
@@ -85,7 +104,7 @@ class Reaction(Object):
 
         # A dictionary of metabolites and their stoichiometric coefficients in
         # this reaction.
-        self._metabolites = {}
+        self._metabolites = dict()
 
         # self.model is None or refers to the cobra.Model that
         # contains self
@@ -99,10 +118,19 @@ class Reaction(Object):
             upper_bound if upper_bound is not None else config.upper_bound
         )
 
-    def _set_id_with_model(self, value):
+    def _set_id_with_model(self, value: str) -> None:
+        """Set Reaction id in model, check that it doesn't already exist.
+
+        The function will rebuild the model reaction index.
+
+        Parameters
+        ----------
+        value: str
+            string that represents the id.
+        """
         if value in self.model.reactions:
             raise ValueError(
-                "The model already contains a reaction with" " the id:", value
+                f"The model already contains a reaction with the id: {value}"
             )
         forward_variable = self.forward_variable
         reverse_variable = self.reverse_variable
@@ -112,31 +140,32 @@ class Reaction(Object):
         reverse_variable.name = self.reverse_id
 
     @property
-    def reverse_id(self):
+    def reverse_id(self) -> str:
         """Generate the id of reverse_variable from the reaction's id."""
         return "_".join(
             (self.id, "reverse", hashlib.md5(self.id.encode("utf-8")).hexdigest()[0:5])
         )
 
     @property
-    def flux_expression(self):
-        """Forward flux expression
+    def flux_expression(self) -> Union["Variable", None]:
+        """Get Forward flux expression.
 
         Returns
         -------
-        sympy expression
+        flux_expression
             The expression representing the the forward flux (if associated
             with model), otherwise None. Representing the net flux if
             model.reversible_encoding == 'unsplit' or None if reaction is
-            not associated with a model"""
+            not associated with a model
+        """
         if self.model is not None:
             return 1.0 * self.forward_variable - 1.0 * self.reverse_variable
         else:
             return None
 
     @property
-    def forward_variable(self):
-        """An optlang variable representing the forward flux
+    def forward_variable(self) -> Union["Variable", None]:
+        """Get an optlang variable representing the forward flux.
 
         Returns
         -------
@@ -150,8 +179,8 @@ class Reaction(Object):
             return None
 
     @property
-    def reverse_variable(self):
-        """An optlang variable representing the reverse flux
+    def reverse_variable(self) -> Union["Variable", None]:
+        """Get an optlang variable representing the reverse flux.
 
         Returns
         -------
@@ -159,16 +188,14 @@ class Reaction(Object):
             An optlang variable for the reverse flux or None if reaction is
             not associated with a model.
         """
-
         if self.model is not None:
             return self.model.variables[self.reverse_id]
         else:
             return None
 
     @property
-    def objective_coefficient(self):
-        """Get the coefficient for this reaction in a linear
-        objective (float)
+    def objective_coefficient(self) -> float:
+        """Get the coefficient for this reaction in a linear objective (float).
 
         Assuming that the objective of the associated model is summation of
         fluxes from a set of reactions, the coefficient for each reaction
@@ -178,29 +205,37 @@ class Reaction(Object):
         return linear_reaction_coefficients(self.model, [self]).get(self, 0)
 
     @objective_coefficient.setter
-    def objective_coefficient(self, value):
+    def objective_coefficient(self, value: float) -> None:
         if self.model is None:
             raise AttributeError("cannot assign objective to a missing model")
         if self.flux_expression is not None:
             set_objective(self.model, {self: value}, additive=True)
 
-    def __copy__(self):
+    def __copy__(self) -> "Reaction":
+        """Copy the Reaction."""
         cop = copy(super(Reaction, self))
         return cop
 
-    def __deepcopy__(self, memo):
+    # Unclear what memo should be
+    def __deepcopy__(self, memo) -> "Reaction":
+        """Copy the reaction with memo."""
         cop = deepcopy(super(Reaction, self), memo)
         return cop
 
     @staticmethod
-    def _check_bounds(lb, ub):
+    def _check_bounds(lb: Union[int, float], ub: Union[int, float]) -> None:
         if lb > ub:
             raise ValueError(
-                "The lower bound must be less than or equal to the upper "
-                "bound ({} <= {}).".format(lb, ub)
+                f"The lower bound must be less than or equal to the upper bound "
+                f"({lb} <= {ub})."
             )
 
-    def update_variable_bounds(self):
+    def update_variable_bounds(self) -> None:
+        """Update and correct variable bounds.
+
+        Sets the forward_variable and reverse_varaible bounds based on lower and
+        upper bounds. This function corrects for bounds defined as inf or -inf.
+        """
         if self.model is None:
             return
         # We know that `lb <= ub`.
@@ -225,8 +260,8 @@ class Reaction(Object):
             )
 
     @property
-    def lower_bound(self):
-        """Get or set the lower bound
+    def lower_bound(self) -> Union[int, float]:
+        """Get or set the lower bound.
 
         Setting the lower bound (float) will also adjust the associated optlang
         variables associated with the reaction. Infeasible combinations,
@@ -240,15 +275,15 @@ class Reaction(Object):
 
     @lower_bound.setter
     @resettable
-    def lower_bound(self, value):
+    def lower_bound(self, value: Union[int, float]) -> None:
         # Validate bounds before setting them.
         self._check_bounds(value, self._upper_bound)
         self._lower_bound = value
         self.update_variable_bounds()
 
     @property
-    def upper_bound(self):
-        """Get or set the upper bound
+    def upper_bound(self) -> Union[int, float]:
+        """Get or set the upper bound.
 
         Setting the upper bound (float) will also adjust the associated optlang
         variables associated with the reaction. Infeasible combinations,
@@ -262,15 +297,15 @@ class Reaction(Object):
 
     @upper_bound.setter
     @resettable
-    def upper_bound(self, value):
+    def upper_bound(self, value: Union[int, float]) -> None:
         # Validate bounds before setting them.
         self._check_bounds(self._lower_bound, value)
         self._upper_bound = value
         self.update_variable_bounds()
 
     @property
-    def bounds(self):
-        """Get or set the bounds directly from a tuple
+    def bounds(self) -> Tuple[Union[float, int], Union[float, int]]:
+        """Get or set the bounds directly from a tuple.
 
         Convenience method for setting upper and lower bounds in one line
         using a tuple of lower and upper bound. Invalid bounds will raise an
@@ -283,7 +318,7 @@ class Reaction(Object):
 
     @bounds.setter
     @resettable
-    def bounds(self, value):
+    def bounds(self, value: Union[Tuple, list]) -> None:
         lower, upper = value
         # Validate bounds before setting them.
         self._check_bounds(lower, upper)
@@ -292,9 +327,9 @@ class Reaction(Object):
         self.update_variable_bounds()
 
     @property
-    def flux(self):
+    def flux(self) -> Union[int, float]:
         """
-        The flux value in the most recent solution.
+        Get the flux value in the most recent solution.
 
         Flux is the primal value of the corresponding variable in the model.
 
@@ -334,24 +369,20 @@ class Reaction(Object):
             check_solver_status(self._model.solver.status)
             return self.forward_variable.primal - self.reverse_variable.primal
         except AttributeError:
-            raise RuntimeError("reaction '{}' is not part of a model".format(self.id))
+            raise RuntimeError(f"reaction '{self.id}' is not part of a model")
         # Due to below all-catch, which sucks, need to reraise these.
         except (RuntimeError, OptimizationError) as err:
-            raise_with_traceback(err)
+            raise err.with_traceback()
         # Would love to catch CplexSolverError and GurobiError here.
         except Exception as err:
-            raise_from(
-                OptimizationError(
-                    "Likely no solution exists. Original solver message: {}."
-                    "".format(str(err))
-                ),
-                err,
-            )
+            raise OptimizationError(
+                f"Likely no solution exists. Original solver message: {str(err)}."
+            ) from err
 
     @property
-    def reduced_cost(self):
+    def reduced_cost(self) -> Union[int, float]:
         """
-        The reduced cost in the most recent solution.
+        Get the reduced cost in the most recent solution.
 
         Reduced cost is the dual value of the corresponding variable in the
         model.
@@ -390,23 +421,20 @@ class Reaction(Object):
             check_solver_status(self._model.solver.status)
             return self.forward_variable.dual - self.reverse_variable.dual
         except AttributeError:
-            raise RuntimeError("reaction '{}' is not part of a model".format(self.id))
+            raise RuntimeError(f"reaction '{self.id}' is not part of a model")
         # Due to below all-catch, which sucks, need to reraise these.
         except (RuntimeError, OptimizationError) as err:
-            raise_with_traceback(err)
+            raise err.with_traceback()
         # Would love to catch CplexSolverError and GurobiError here.
         except Exception as err:
-            raise_from(
-                OptimizationError(
-                    "Likely no solution exists. Original solver message: {}."
-                    "".format(str(err))
-                ),
-                err,
-            )
+            raise OptimizationError(
+                f"Likely no solution exists. Original solver message: {str(err)}."
+            ) from err
 
     # read-only
     @property
-    def metabolites(self):
+    def metabolites(self) -> Dict:
+        """Get a dictionary of metabolites and coefficients."""
         return self._metabolites.copy()
 
     @property
@@ -419,7 +447,7 @@ class Reaction(Object):
         """
         return frozenset(self._genes)
 
-    def _update_genes_from_gpr(self, new_gene_names: set = None) -> None:
+    def _update_genes_from_gpr(self, new_gene_names: Optional[Set] = None) -> None:
         """Update genes of reation based on GPR.
 
         Parameters
@@ -458,10 +486,7 @@ class Reaction(Object):
                     if not len(g.reactions) and self.model and g in self.model.genes:
                         remove_genes(self.model, [g], False)
                 except KeyError:
-                    warn(
-                        "could not remove old gene %s from reaction %s"
-                        % (g.id, self.id)
-                    )
+                    warn(f"could not remove old gene {g.id} from reaction {self.id}")
 
     @property
     def gene_reaction_rule(self) -> str:
@@ -507,12 +532,12 @@ class Reaction(Object):
         return self._gpr.to_string(names=names)
 
     @property
-    def gpr(self):
+    def gpr(self) -> GPR:
         """Return the GPR associated with the reaction."""
         return self._gpr
 
     @gpr.setter
-    def gpr(self, value: GPR):
+    def gpr(self, value: GPR) -> None:
         """Set a new GPR for the reaction, using GPR() class.
 
         Parameters
@@ -524,7 +549,7 @@ class Reaction(Object):
         self._update_genes_from_gpr()
 
     @property
-    def functional(self):
+    def functional(self) -> bool:
         """All required enzymes for reaction are functional.
 
         Returns
@@ -541,8 +566,8 @@ class Reaction(Object):
         return True
 
     @property
-    def x(self):
-        """The flux through the reaction in the most recent solution.
+    def x(self) -> Union[int, float]:
+        """Get the flux through the reaction in the most recent solution.
 
         Flux values are computed from the primal values of the variables in
         the solution.
@@ -551,8 +576,8 @@ class Reaction(Object):
         return self.flux
 
     @property
-    def y(self):
-        """The reduced cost of the reaction in the most recent solution.
+    def y(self) -> Union[int, float]:
+        """Get the reduced cost of the reaction in the most recent solution.
 
         Reduced costs are computed from the dual values of the variables in
         the solution.
@@ -561,33 +586,41 @@ class Reaction(Object):
         return self.reduced_cost
 
     @property
-    def reversibility(self):
-        """Whether the reaction can proceed in both directions (reversible)
+    def reversibility(self) -> bool:
+        """Whether the reaction can proceed in both directions (reversible).
 
         This is computed from the current upper and lower bounds.
+
+        Returns
+        -------
+        bool
 
         """
         return self._lower_bound < 0 < self._upper_bound
 
     @reversibility.setter
-    def reversibility(self, value):
+    def reversibility(self, value: bool) -> None:
         warn("Setting reaction reversibility is ignored")
 
     @property
-    def boundary(self):
+    def boundary(self) -> bool:
         """Whether or not this reaction is an exchange reaction.
 
-        Returns `True` if the reaction has either no products or reactants.
+        Returns
+        -------
+        bool:   `True` if the reaction has either no products or reactants.
         """
         return len(self.metabolites) == 1 and not (self.reactants and self.products)
 
     @property
-    def model(self):
-        """returns the model the reaction is a part of"""
+    def model(self) -> "Model":
+        """Return the model the reaction is a part of."""
         return self._model
 
-    def _update_awareness(self):
-        """Make sure all metabolites and genes that are associated with
+    def _update_awareness(self) -> None:
+        """Update awareness for genes and metaoblites of the reaction.
+
+        Make sure all metabolites and genes that are associated with
         this reaction are aware of it.
 
         """
@@ -596,8 +629,8 @@ class Reaction(Object):
         for x in self._genes:
             x._reaction.add(self)
 
-    def remove_from_model(self, remove_orphans=False):
-        """Removes the reaction from a model.
+    def remove_from_model(self, remove_orphans: Optional[bool] = False) -> None:
+        """Remove the reaction from a model.
 
         This removes all associations between a reaction the associated
         model, metabolites and genes.
@@ -612,15 +645,16 @@ class Reaction(Object):
         """
         self._model.remove_reactions([self], remove_orphans=remove_orphans)
 
-    def delete(self, remove_orphans=False):
-        """Removes the reaction from a model.
+    def delete(self, remove_orphans: Optional[bool] = False) -> None:
+        """Remove the reaction from a model.
 
         This removes all associations between a reaction the associated
         model, metabolites and genes.
 
         The change is reverted upon exit when using the model as a context.
 
-        Deprecated, use `reaction.remove_from_model` instead.
+        .. deprecated ::
+        use `reaction.remove_from_model` instead.
 
         Parameters
         ----------
@@ -634,8 +668,10 @@ class Reaction(Object):
         )
         self.remove_from_model(remove_orphans=remove_orphans)
 
-    def __setstate__(self, state):
-        """Probably not necessary to set _model as the cobra.Model that
+    def __setstate__(self, state: Dict) -> None:
+        """Set state fo reaction.
+
+        Probably not necessary to set _model as the cobra.Model that
         contains self sets the _model attribute for all metabolites and genes
         in the reaction.
 
@@ -662,8 +698,8 @@ class Reaction(Object):
             setattr(x, "_model", self._model)
             x._reaction.add(self)
 
-    def copy(self):
-        """Copy a reaction
+    def copy(self) -> "Reaction":
+        """Copy a reaction.
 
         The referenced metabolites and genes are also copied.
 
@@ -685,13 +721,19 @@ class Reaction(Object):
             i._model = model
         return new_reaction
 
-    def __add__(self, other):
-        """Add two reactions
+    def __add__(self, other: "Reaction") -> "Reaction":
+        """Add two reactions and return a new one.
 
         The stoichiometry will be the combined stoichiometry of the two
         reactions, and the gene reaction rule will be both rules combined by an
         and. All other attributes (i.e. reaction bounds) will match those of
-        the first reaction
+        the first reaction.
+
+        Does not modify in place.
+
+        Returns
+        -------
+        Reaction - new reaction with the added properties.
 
         """
         new_reaction = self.copy()
@@ -704,15 +746,22 @@ class Reaction(Object):
 
     __radd__ = __add__
 
-    def __iadd__(self, other):
+    def __iadd__(self, other: "Reaction") -> "Reaction":
+        """Add two reactions in place and return the modified first one.
 
+        The stoichiometry will be the combined stoichiometry of the two
+        reactions, and the gene reaction rule will be both rules combined by an
+        and. All other attributes (i.e. reaction bounds) will match those of
+        the first reaction.
+
+        Modifies in place.
+        """
         self.add_metabolites(other._metabolites, combine=True)
         gpr1 = self.gene_reaction_rule.strip()
         gpr2 = other.gene_reaction_rule.strip()
         if gpr1 != "" and gpr2 != "":
-            self.gene_reaction_rule = "(%s) and (%s)" % (
-                self.gene_reaction_rule,
-                other.gene_reaction_rule,
+            self.gene_reaction_rule = (
+                f"({self.gene_reaction_rule}) and " f"({other.gene_reaction_rule})"
             )
         elif gpr1 != "" and gpr2 == "":
             self.gene_reaction_rule = gpr1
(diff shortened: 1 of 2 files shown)


src/cobra/test/test_core/test_core_reaction.py
```diff
@@ -44,7 +44,7 @@ def test_gpr() -> None:
 
 
 def test_gpr_uppercase():
```
synchon: Please add the return type annotation as `None`.

src/cobra/test/test_core/test_core_reaction.py
```diff
@@ -54,7 +54,10 @@ def test_gpr_uppercase():
 
 @pytest.mark.parametrize("input_gpr", ["(a1 or a2", "(forT or "])
 def test_gpr_malformed(input_gpr):
```
synchon: Please add the argument type annotation as `str` and return type annotation as `None`.

src/cobra/core/reaction.py
```diff
@@ -1,20 +1,33 @@
...
+)
 from warnings import warn
 
-from future.utils import raise_from, raise_with_traceback
+
+if TYPE_CHECKING:
+    from optlang.interface import Variable
+    from cobra import Model, Solution
+    from cobra.summary import ReactionSummary
+    from pandas import DataFrame
```
synchon: Please use `import pandas as pd` and then refer `DataFrame` as `pd.DataFrame` so as to maintain the standard used by `pandas` community and also make it easy for us to understand foreign imports.

src/cobra/core/reaction.py
```diff
@@ -1,20 +1,33 @@
...
 
 from cobra.core.configuration import Configuration
```
synchon: Please use relative imports (wherever applicable) to refer to internal imports, for example: `from .configuration import Configuration` .

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
     ----------
     id : string
         The identifier to associate with this reaction
```
synchon: Please mention the default value for the parameter in the docstring as:
`The identifier to associate with this reaction (default None).`

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
-    name : string
+    name : string, optional
```
synchon: Please adapt the previous comment for this docstring as well. Also, prefer `str` for docstring as well and keep it consistent throughout.

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
         A human readable name for the reaction
-    subsystem : string
+    subsystem : string, optional
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
         Subsystem where the reaction is meant to occur
-    lower_bound : float
+    lower_bound : float, optional
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
         The lower flux bound
-    upper_bound : float
+    upper_bound : float, optional
```
synchon, line 82: Same as above.

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
     """
 
+    # noinspection PyShadowingBuiltins
     def __init__(
-        self, id=None, name="", subsystem="", lower_bound=0.0, upper_bound=None
-    ):
-        Object.__init__(self, id, name)
+        self,
+        id: Optional[str] = None,
+        name: Optional[str] = "",
```
synchon: Please only use `str` for the type annotation.

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
+        subsystem: Optional[str] = "",
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -62,20 +71,30 @@ class Reaction(Object):
...
+        lower_bound: Optional[float] = 0.0,
```
synchon: Please adapt the previous comment for this type annotation as well.

src/cobra/core/reaction.py
```diff
@@ -84,7 +103,7 @@ def __init__(
 
         # A dictionary of metabolites and their stoichiometric coefficients in
         # this reaction.
-        self._metabolites = {}
+        self._metabolites = dict()
```
synchon: Please use literal syntax like so: `{}` in place of `dict()` .

src/cobra/core/reaction.py
```diff
@@ -98,10 +117,19 @@ def __init__(
...
+    def _set_id_with_model(self, value: str) -> None:
+        """Set Reaction id in model, check that it doesn't already exist.
+
+        The function will rebuild the model reaction index.
+
+        Parameters
+        ----------
+        value: str
+            string that represents the id.
+        """
```
synchon, line 136: As the function can raise exception, please add that in the docstring as well.

src/cobra/core/reaction.py
```diff
@@ -111,31 +139,32 @@ def _set_id_with_model(self, value):
         reverse_variable.name = self.reverse_id
 
     @property
-    def reverse_id(self):
+    def reverse_id(self) -> str:
```
synchon, line 149: Please add a detailed docstring for this method.

src/cobra/core/reaction.py
```diff
@@ -111,31 +139,32 @@ def _set_id_with_model(self, value):
...
         """Generate the id of reverse_variable from the reaction's id."""
         return "_".join(
             (self.id, "reverse", hashlib.md5(self.id.encode("utf-8")).hexdigest()[0:5])
         )
 
     @property
-    def flux_expression(self):
-        """Forward flux expression
+    def flux_expression(self) -> Union["Variable", None]:
```
synchon: Please use `Optional["Variable"]` for the return type.

src/cobra/core/reaction.py
```diff
@@ -111,31 +139,32 @@ def _set_id_with_model(self, value):
...
+        """
         if self.model is not None:
             return 1.0 * self.forward_variable - 1.0 * self.reverse_variable
         else:
             return None
 
     @property
-    def forward_variable(self):
-        """An optlang variable representing the forward flux
+    def forward_variable(self) -> Union["Variable", None]:
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -149,25 +178,23 @@ def forward_variable(self):
             return None
 
     @property
-    def reverse_variable(self):
-        """An optlang variable representing the reverse flux
+    def reverse_variable(self) -> Union["Variable", None]:
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -149,25 +178,23 @@ def forward_variable(self):
...
         if self.model is not None:
             return self.model.variables[self.reverse_id]
         else:
             return None
 
     @property
-    def objective_coefficient(self):
-        """Get the coefficient for this reaction in a linear
-        objective (float)
+    def objective_coefficient(self) -> float:
```
synchon, line 210: Please document the return type in the docstring.

src/cobra/core/reaction.py
```diff
@@ -177,29 +204,37 @@ def objective_coefficient(self):
         return linear_reaction_coefficients(self.model, [self]).get(self, 0)
 
     @objective_coefficient.setter
-    def objective_coefficient(self, value):
+    def objective_coefficient(self, value: float) -> None:
```
synchon, line 231: Please add a docstring including Parameters and Raises sections.

src/cobra/core/reaction.py
```diff
@@ -177,29 +204,37 @@ def objective_coefficient(self):
...
         if self.model is None:
             raise AttributeError("cannot assign objective to a missing model")
         if self.flux_expression is not None:
             set_objective(self.model, {self: value}, additive=True)
 
-    def __copy__(self):
+    def __copy__(self) -> "Reaction":
```
synchon, line 249: Please add the Returns section in the docstring.

src/cobra/core/reaction.py
```diff
@@ -177,29 +204,37 @@ def objective_coefficient(self):
...
+        """Copy the Reaction."""
         cop = copy(super(Reaction, self))
         return cop
 
-    def __deepcopy__(self, memo):
+    # Unclear what memo should be
+    def __deepcopy__(self, memo) -> "Reaction":
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -177,29 +204,37 @@ def objective_coefficient(self):
...
+        """Copy the reaction with memo."""
         cop = deepcopy(super(Reaction, self), memo)
         return cop
 
     @staticmethod
-    def _check_bounds(lb, ub):
+    def _check_bounds(lb: Union[int, float], ub: Union[int, float]) -> None:
```
synchon: As the input types for `lower_bound` and `upper_bound` are taken as `float`, we can use `float` for the argument types as well. Also, please add a Raises section in the docstring.

src/cobra/core/reaction.py
```diff
@@ -177,29 +204,37 @@ def objective_coefficient(self):
...
-                "bound ({} <= {}).".format(lb, ub)
+                f"The lower bound must be less than or equal to the upper bound "
+                f"({lb} <= {ub})."
             )
 
-    def update_variable_bounds(self):
+    def update_variable_bounds(self) -> None:
+        """Update and correct variable bounds.
+
+        Sets the forward_variable and reverse_varaible bounds based on lower and
```
synchon: A small typo: `variable` in place of `varaible`

src/cobra/core/reaction.py
```diff
@@ -224,8 +259,8 @@ def update_variable_bounds(self):
             )
 
     @property
-    def lower_bound(self):
-        """Get or set the lower bound
+    def lower_bound(self) -> Union[int, float]:
```
synchon: As earlier, please make the return type as `float` and also add a Returns section to the docstring. This method only gets the value, so I believe you can move the sets directive in the docstring to the setter method.

src/cobra/core/reaction.py
```diff
@@ -239,15 +274,15 @@ def lower_bound(self):
 
     @lower_bound.setter
     @resettable
-    def lower_bound(self, value):
+    def lower_bound(self, value: Union[int, float]) -> None:
```
synchon: Please make the argument type as `float` and also add a docstring with the necessary information.

src/cobra/core/reaction.py
```diff
@@ -239,15 +274,15 @@ def lower_bound(self):
...
         # Validate bounds before setting them.
         self._check_bounds(value, self._upper_bound)
         self._lower_bound = value
         self.update_variable_bounds()
 
     @property
-    def upper_bound(self):
-        """Get or set the upper bound
+    def upper_bound(self) -> Union[int, float]:
```
synchon: Same comment as for the lower bound case.

src/cobra/core/reaction.py
```diff
@@ -261,15 +296,15 @@ def upper_bound(self):
 
     @upper_bound.setter
     @resettable
-    def upper_bound(self, value):
+    def upper_bound(self, value: Union[int, float]) -> None:
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -261,15 +296,15 @@ def upper_bound(self):
...
         # Validate bounds before setting them.
         self._check_bounds(self._lower_bound, value)
         self._upper_bound = value
         self.update_variable_bounds()
 
     @property
-    def bounds(self):
-        """Get or set the bounds directly from a tuple
+    def bounds(self) -> Tuple[Union[float, int], Union[float, int]]:
```
synchon: Please adapt the return type and also document it in the docstring.

src/cobra/core/reaction.py
```diff
@@ -282,7 +317,7 @@ def bounds(self):
 
     @bounds.setter
     @resettable
-    def bounds(self, value):
+    def bounds(self, value: Union[Tuple, list]) -> None:
```
synchon: Please add a docstring.

src/cobra/core/reaction.py
```diff
@@ -291,9 +326,9 @@ def bounds(self, value):
         self.update_variable_bounds()
 
     @property
-    def flux(self):
+    def flux(self) -> Union[int, float]:
```
synchon: I believe flux value is always returned as `float`.

src/cobra/core/reaction.py
```diff
@@ -333,24 +368,20 @@ def flux(self):
...
-                ),
-                err,
-            )
+            raise OptimizationError(
+                f"Likely no solution exists. Original solver message: {str(err)}."
+            ) from err
 
     @property
-    def reduced_cost(self):
+    def reduced_cost(self) -> Union[int, float]:
```
synchon: I think reduced cost is always returned as `float`.

src/cobra/core/reaction.py
```diff
@@ -389,23 +420,20 @@ def reduced_cost(self):
...
-            )
+            raise OptimizationError(
+                f"Likely no solution exists. Original solver message: {str(err)}."
+            ) from err
 
     # read-only
     @property
-    def metabolites(self):
+    def metabolites(self) -> Dict:
+        """Get a dictionary of metabolites and coefficients."""
```
synchon: Please add a return type in the docstring.

src/cobra/core/reaction.py
```diff
@@ -504,12 +529,12 @@ def gene_name_reaction_rule(self):
         return self._gpr.to_string(names=names)
 
     @property
-    def gpr(self):
+    def gpr(self) -> GPR:
```
synchon, line 678: Please add a Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -538,8 +563,8 @@ def functional(self):
         return True
 
     @property
-    def x(self):
-        """The flux through the reaction in the most recent solution.
+    def x(self) -> Union[int, float]:
```
synchon: The return type can be `float`.

src/cobra/core/reaction.py
```diff
@@ -548,8 +573,8 @@ def x(self):
         return self.flux
 
     @property
-    def y(self):
-        """The reduced cost of the reaction in the most recent solution.
+    def y(self) -> Union[int, float]:
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -558,33 +583,41 @@ def y(self):
...
+        Returns
+        -------
+        bool
+
         """
         return self._lower_bound < 0 < self._upper_bound
 
     @reversibility.setter
-    def reversibility(self, value):
+    def reversibility(self, value: bool) -> None:
```
synchon, line 774: Please add a docstring.

src/cobra/core/reaction.py
```diff
@@ -558,33 +583,41 @@ def y(self):
...
+        Returns
+        -------
+        bool:   `True` if the reaction has either no products or reactants.
         """
         return len(self.metabolites) == 1 and not (self.reactants and self.products)
 
     @property
-    def model(self):
-        """returns the model the reaction is a part of"""
+    def model(self) -> "Model":
```
synchon: Please add a Returns section in the docstring.

src/cobra/core/reaction.py
```diff
@@ -593,8 +626,8 @@ def _update_awareness(self):
         for x in self._genes:
             x._reaction.add(self)
 
-    def remove_from_model(self, remove_orphans=False):
-        """Removes the reaction from a model.
+    def remove_from_model(self, remove_orphans: Optional[bool] = False) -> None:
```
synchon: Please make the type for `remove_orphans` as `bool`.

src/cobra/core/reaction.py
```diff
@@ -609,15 +642,16 @@ def remove_from_model(self, remove_orphans=False):
         """
         self._model.remove_reactions([self], remove_orphans=remove_orphans)
 
-    def delete(self, remove_orphans=False):
-        """Removes the reaction from a model.
+    def delete(self, remove_orphans: Optional[bool] = False) -> None:
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -631,8 +665,10 @@ def delete(self, remove_orphans=False):
         )
         self.remove_from_model(remove_orphans=remove_orphans)
 
-    def __setstate__(self, state):
-        """Probably not necessary to set _model as the cobra.Model that
+    def __setstate__(self, state: Dict) -> None:
```
synchon, line 861: Please add a Parameters section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -653,14 +689,14 @@ def __setstate__(self, state):
...
+            x._model = self._model
             x._reaction.add(self)
         for x in state["_genes"]:
-            setattr(x, "_model", self._model)
+            x._model = self._model
             x._reaction.add(self)
 
-    def copy(self):
-        """Copy a reaction
+    def copy(self) -> "Reaction":
```
synchon, line 896: Please add a Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -682,13 +718,19 @@ def copy(self):
...
+        """Add two reactions and return a new one.
 
         The stoichiometry will be the combined stoichiometry of the two
         reactions, and the gene reaction rule will be both rules combined by an
         and. All other attributes (i.e. reaction bounds) will match those of
-        the first reaction
+        the first reaction.
+
+        Does not modify in place.
+
```
synchon, line 932: Please add a Parameters section here.

src/cobra/core/reaction.py
```diff
@@ -701,38 +743,82 @@ def __add__(self, other):
...
-    def __iadd__(self, other):
+    def __iadd__(self, other: "Reaction") -> "Reaction":
+        """Add two reactions in place and return the modified first one.
 
+        The stoichiometry will be the combined stoichiometry of the two
+        reactions, and the gene reaction rule will be both rules combined by an
+        and. All other attributes (i.e. reaction bounds) will match those of
+        the first reaction.
+
+        Modifies in place.
```
synchon, line 961: Please add Parameters and Returns sections here.

src/cobra/core/reaction.py
```diff
@@ -701,38 +743,82 @@ def __add__(self, other):
...
+    def __sub__(self, other: "Reaction") -> "Reaction":
+        """Subtract two reactions and return a new one.
+
+        The stoichiometry will be the subtracted stoichiometry of the two
+        reactions, and the gene_reaction_rule will be the gene_reaction_rule of the
+        first reaction. All other attributes (i.e. reaction bounds) will match those of
+        the first reaction.
+
+        Does not modify in place. The name will still be that of the first reaction.
+
```
synchon, line 994: Please add a Parameters section here.

src/cobra/core/reaction.py
```diff
@@ -701,38 +743,82 @@ def __add__(self, other):
...
+    def __isub__(self, other: "Reaction") -> "Reaction":
+        """Subtract metabolites of one reaction from another in place.
+
+        The stoichiometry will be the metabolites of self minus the metabolites
+         of the other. All other attributes including gene_reaction_rule
+         (i.e. reaction bounds) will match those of
+        the first reaction.
+
+        Modifies in place and changes the original reaction.
+
```
synchon, line 1017: Please add a Parameters section here.

src/cobra/core/reaction.py
```diff
@@ -701,38 +743,82 @@ def __add__(self, other):
...
+        Returns
+        -------
+        Reaction - self with the subtracted metabolites.
+        """
         self.subtract_metabolites(other._metabolites, combine=True)
         return self
 
-    def __imul__(self, coefficient):
-        """Scale coefficients in a reaction by a given value
+    def __imul__(self, coefficient: Union[int, float]) -> "Reaction":
```
synchon: Please enforce `float` for the type annotation and the docstring as well.

src/cobra/core/reaction.py
```diff
@@ -751,24 +837,43 @@ def __imul__(self, coefficient):
 
         return self
 
-    def __mul__(self, coefficient):
+    def __mul__(self, coefficient: Union[int, float]) -> "Reaction":
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -751,24 +837,43 @@ def __imul__(self, coefficient):
...
+        Reaction
+            Returns a new reaction, identical to the original except coefficients.
+        """
         new = self.copy()
         new *= coefficient
         return new
 
     @property
-    def reactants(self):
+    def reactants(self) -> List[Metabolite]:
```
synchon, line 1090: Please add a Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -751,24 +837,43 @@ def __imul__(self, coefficient):
...
         """Return a list of reactants for the reaction."""
         return [k for k, v in self._metabolites.items() if v < 0]
 
     @property
-    def products(self):
-        """Return a list of products for the reaction"""
+    def products(self) -> List[Metabolite]:
```
synchon, line 1101: Please add a Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -751,24 +837,43 @@ def __imul__(self, coefficient):
...
+        """Return a list of products for the reaction."""
         return [k for k, v in self._metabolites.items() if v >= 0]
 
-    def get_coefficient(self, metabolite_id):
-        """
-        Return the stoichiometric coefficient of a metabolite.
+    def get_coefficient(
+        self, metabolite_id: Union[str, Metabolite]
+    ) -> Union[int, float]:
```
synchon: I believe the return is always a `float`, can you please check it? Also, please add a Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -781,20 +886,32 @@ def get_coefficient(self, metabolite_id):
         _id_to_metabolites = {m.id: m for m in self._metabolites}
         return self._metabolites[_id_to_metabolites[metabolite_id]]
 
-    def get_coefficients(self, metabolite_ids):
-        """
-        Return the stoichiometric coefficients for a list of metabolites.
+    def get_coefficients(
+        self, metabolite_ids: Iterable[Union[str, Metabolite]]
+    ) -> Iterator[Union[int, float]]:
```
synchon: I think the return type can be `Iterable[float]`.

src/cobra/core/reaction.py
```diff
@@ -781,20 +886,32 @@ def get_coefficient(self, metabolite_id):
...
+        map: Iterable
+            Returns the result of map function, which is a map object (an Iterable).
+
         """
         return map(self.get_coefficient, metabolite_ids)
 
-    def add_metabolites(self, metabolites_to_add, combine=True, reversibly=True):
+    def add_metabolites(
+        self,
+        metabolites_to_add: Dict[Metabolite, Union[int, float]],
```
synchon: Please enforce type `float` for the values.

src/cobra/core/reaction.py
```diff
@@ -781,20 +886,32 @@ def get_coefficient(self, metabolite_id):
...
+        combine: Optional[bool] = True,
```
synchon: Please make the type as `bool` and document the optional behaviour in the docstring.

src/cobra/core/reaction.py
```diff
@@ -781,20 +886,32 @@ def get_coefficient(self, metabolite_id):
...
+        reversibly: Optional[bool] = True,
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -914,7 +1030,12 @@ def add_metabolites(self, metabolites_to_add, combine=True, reversibly=True):
                     )
                 )
 
-    def subtract_metabolites(self, metabolites, combine=True, reversibly=True):
+    def subtract_metabolites(
+        self,
+        metabolites: Dict[Metabolite, Union[int, float]],
```
synchon: The type for the value can be `float`.

src/cobra/core/reaction.py
```diff
@@ -914,7 +1030,12 @@ def add_metabolites(self, metabolites_to_add, combine=True, reversibly=True):
...
+        combine: Optional[bool] = True,
```
synchon: Please make the type as `bool` and document the optional behaviour in the docstring.

src/cobra/core/reaction.py
```diff
@@ -914,7 +1030,12 @@ def add_metabolites(self, metabolites_to_add, combine=True, reversibly=True):
...
+        reversibly: Optional[bool] = True,
```
synchon: Same as above.

src/cobra/core/reaction.py
```diff
@@ -950,18 +1071,20 @@ def subtract_metabolites(self, metabolites, combine=True, reversibly=True):
         )
 
     @property
-    def reaction(self):
-        """Human readable reaction string"""
+    def reaction(self) -> str:
+        """Return Human readable reaction string."""
```
synchon: Please add a Returns section in the docstring.

src/cobra/core/reaction.py
```diff
@@ -950,18 +1071,20 @@ def subtract_metabolites(self, metabolites, combine=True, reversibly=True):
...
         return self.build_reaction_string()
 
     @reaction.setter
-    def reaction(self, value):
-        return self.build_reaction_from_string(value)
+    def reaction(self, value: str):
```
synchon: Please add a docstring and the return type annotation.

src/cobra/core/reaction.py
```diff
@@ -950,18 +1071,20 @@ def subtract_metabolites(self, metabolites, combine=True, reversibly=True):
...
+        self.build_reaction_from_string(value)
 
-    def build_reaction_string(self, use_metabolite_names=False):
-        """Generate a human readable reaction string"""
+    def build_reaction_string(
+        self, use_metabolite_names: Optional[bool] = False
```
synchon: Please use `bool` as the type for `use_metabolite_names`.

src/cobra/core/reaction.py
```diff
@@ -950,18 +1071,20 @@ def subtract_metabolites(self, metabolites, combine=True, reversibly=True):
...
+    ) -> str:
+        """Generate a human readable reaction string."""
```
synchon: Please add Parameters and Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -988,37 +1111,40 @@ def format(number):
...
 
-        returns a dict of {element: amount} for unbalanced elements.
-        "charge" is treated as an element in this dict
-        This should be empty for balanced reactions.
+        Returns
+        -------
+        dict
+            a dict of {element: amount} for unbalanced elements.
+            "charge" is treated as an element in this dict
+            This should be empty for balanced reactions.
```
synchon, line 1393: Please add a Raises section here.

src/cobra/core/reaction.py
```diff
@@ -988,37 +1111,40 @@ def format(number):
         reaction_string += " + ".join(product_bits)
         return reaction_string
 
-    def check_mass_balance(self):
-        """Compute mass and charge balance for the reaction
+    def check_mass_balance(self) -> Dict[str, Union[int, float]]:
```
synchon: I think the value for the return can be of type `float`.

src/cobra/core/reaction.py
```diff
@@ -988,37 +1111,40 @@ def format(number):
...
             for element, amount in metabolite.elements.items():
                 reaction_element_dict[element] += coefficient * amount
         # filter out 0 values
         return {k: v for k, v in reaction_element_dict.items() if v != 0}
 
     @property
-    def compartments(self):
-        """lists compartments the metabolites are in"""
+    def compartments(self) -> Set:
+        """Return set of compartments the metabolites are in."""
```
synchon: Please add a Returns section to the docstring.

src/cobra/core/reaction.py
```diff
@@ -988,37 +1111,40 @@ def format(number):
...
         return {
             met.compartment for met in self._metabolites if met.compartment is not None
         }
 
-    def get_compartments(self):
-        """lists compartments the metabolites are in"""
+    def get_compartments(self) -> list:
```
synchon, line 1424: Maybe add a Returns section to the docstring? Can be kept as it is as it has a deprecation warning on it.

src/cobra/core/reaction.py
```diff
@@ -1041,20 +1167,20 @@ def _dissociate_gene(self, cobra_gene):
...
-        reaction_str,
-        verbose=True,
-        fwd_arrow=None,
-        rev_arrow=None,
-        reversible_arrow=None,
-        term_split="+",
-    ):
-        """Builds reaction from reaction equation reaction_str using parser
+        reaction_str: str,
+        verbose: Optional[bool] = True,
```
synchon: Please use `bool` for the type.

src/cobra/core/reaction.py
```diff
@@ -1041,20 +1167,20 @@ def _dissociate_gene(self, cobra_gene):
...
+        fwd_arrow: Optional[AnyStr] = None,
+        rev_arrow: Optional[AnyStr] = None,
+        reversible_arrow: Optional[AnyStr] = None,
+        term_split: Optional[str] = "+",
```
synchon: Please use `str` for the type.

src/cobra/core/reaction.py
```diff
@@ -1070,11 +1196,11 @@ def build_reaction_from_string(
...
+        fwd_arrow : AnyStr
             for forward irreversible reaction arrows
-        rev_arrow : re.compile
+        rev_arrow : AnyStr
             for backward irreversible reaction arrows
-        reversible_arrow : re.compile
+        reversible_arrow : AnyStr
             for reversible reaction arrows
         term_split : string
             dividing individual metabolite entries
```
synchon: Please add necessary information for a Raises section here.

src/cobra/core/reaction.py
```diff
@@ -1041,20 +1167,20 @@ def _dissociate_gene(self, cobra_gene):
...
-        verbose=True,
-        fwd_arrow=None,
-        rev_arrow=None,
-        reversible_arrow=None,
-        term_split="+",
-    ):
-        """Builds reaction from reaction equation reaction_str using parser
+        reaction_str: str,
+        verbose: Optional[bool] = True,
+        fwd_arrow: Optional[AnyStr] = None,
```
synchon, line 1468: I believe `re.compile()` generates a regular expression object and `AnyStr` is not accurate for that.

src/cobra/core/reaction.py
```diff
@@ -1041,20 +1167,20 @@ def _dissociate_gene(self, cobra_gene):
...
+        rev_arrow: Optional[AnyStr] = None,
```
synchon, line 1469: Same as above.

src/cobra/core/reaction.py
```diff
@@ -1041,20 +1167,20 @@ def _dissociate_gene(self, cobra_gene):
...
+        reversible_arrow: Optional[AnyStr] = None,
```
synchon, line 1470: Same as above.

src/cobra/core/reaction.py
```diff
@@ -1186,42 +1316,38 @@ def summary(self, solution=None, fva=None):
             fva=fva,
         )
 
-    def __str__(self):
-        return "{id}: {stoichiometry}".format(
-            id=self.id, stoichiometry=self.build_reaction_string()
-        )
+    def __str__(self) -> str:
```
synchon, line 1615: Please add a Returns section for the docstring.

src/cobra/core/reaction.py
```diff
@@ -1186,42 +1316,38 @@ def summary(self, solution=None, fva=None):
...
+        """Return reaction id and reaction as string."""
+        return f"{self.id}: {self.build_reaction_string()}"
 
-    def _repr_html_(self):
-        return """
+    def _repr_html_(self) -> str:
```
synchon, line 1625: Same as above.

src/cobra/core/reaction.py
```diff
@@ -1186,42 +1316,38 @@ def summary(self, solution=None, fva=None):
...
+                <td><strong>Reaction identifier</strong></td><td>{format_long_string(
+            self.id, 100)}</td>
             </tr><tr>
-                <td><strong>Name</strong></td><td>{name}</td>
+                <td><strong>Name</strong></td><td>{format_long_string(
+            self.name, 100)}</td>
             </tr><tr>
                 <td><strong>Memory address</strong></td>
-                <td>{address}</td>
+                <td>{f"0x0{id(self):x}"}</td>
```
synchon: Please make it as: `{id(self):#x}`.

synchon (changes requested):
I have tried my best to go through everything. As the file is large, I might have missed a couple of things, so feel free to add those enhancements.

src/cobra/core/reaction.py
akaviaLab: You mean not Optional for empty strings?
Another maintainer said that default of None should be Optional. Do you agree?

src/cobra/core/reaction.py
akaviaLab: Done. Can you help me figure out what memo should be?

src/cobra/core/reaction.py
akaviaLab: I'm not sure, but I made it a float.

src/cobra/core/reaction.py
akaviaLab: The code uses re.compile(re.escape(fwd_arrow)) (see line 1213 below, line 1501 in the revised code, line 1087 in the original code). re.escape accepts a pattern, either str or bytes (binary str). AnyStr is equal to Union[str, bytes].
I can also set it as str, but it definitely shouldn't be Pattern. Try re.escape(re.compile('-->')) and you'll get an error.

akaviaLab (commented):
Thank you very much for your detailed comments. I tried to respond to all of them.
Can you please help me with memo (in deepcopy)? I can't figure out what it is supposed to be?

src/cobra/core/reaction.py
```diff
@@ -60,22 +69,32 @@ class Reaction(Object):
...
+    id : str, default None
+        The identifier to associate with this reaction (default None).
+    name : str, optional, default ""
+        A human readable name for the reaction (default "").
+    subsystem : str, optional, default ""
+        Subsystem where the reaction is meant to occur (default "").
+    lower_bound : float, default 0
+        The lower flux bound (default 0).
+    upper_bound : float, optional, default None
+        The upper flux bound (default None).
```
Midnighter: This is a new style and not in line with numpydoc https://numpydoc.readthedocs.io/en/latest/format.html#parameters, default values should be in the description.

```suggestion
    id : str or None, optional
        The identifier to associate with this reaction (default None).
    name : str, optional
        A human readable name for the reaction (default "").
    subsystem : str, optional
        Subsystem where the reaction is meant to occur (default "").
    lower_bound : float, optional
        The lower flux bound (default 0).
    upper_bound : float or None, optional
        The upper flux bound (default None).
```

src/cobra/core/reaction.py
Midnighter: It's a `dict` https://docs.python.org/3/library/copy.html#copy.deepcopy

```suggestion
    def __deepcopy__(self, memo: dict) -> "Reaction":
```

src/cobra/core/reaction.py
```diff
@@ -149,57 +190,121 @@ def forward_variable(self):
...
         return cop
 
-    def __deepcopy__(self, memo):
+    # Unclear what memo should be
+    def __deepcopy__(self, memo) -> "Reaction":
+        """Copy the reaction with memo.
+
+        Parameters
+        ----------
+        memo
```
Midnighter:
```suggestion
        memo: dict
            Automatically passed parameter.
```

src/cobra/core/reaction.py
synchon: If the default value is `""` (empty string), the type should be `str`. And, if the default value is `None`, then the type becomes `Optional[...]`.

src/cobra/core/reaction.py
```diff
@@ -224,65 +329,117 @@ def update_variable_bounds(self):
...
+        -------
+        tuple: lower_bound, upper_bound
+            A tuple of floats, representing the lower and upper bound.
         """
         return self.lower_bound, self.upper_bound
 
     @bounds.setter
     @resettable
-    def bounds(self, value):
+    def bounds(self, value: Union[Tuple[float, float], List[float]]) -> None:
```
Midnighter: It would be more concise to say that a [Sequence](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) interface needs to be supported.

```suggestion
    def bounds(self, value: Sequence[float]) -> None:
```

but if you want to highlight that the usual case is a pair that's fine, too:

```suggestion
    def bounds(self, value: Union[Tuple[float, float], Sequence[float]]) -> None:
```

src/cobra/core/reaction.py
```diff
@@ -224,65 +329,117 @@ def update_variable_bounds(self):
...
+        """Set the bounds directly, using a tuple or list.
+
+        Parameters
+        ----------
+        value: tuple or list
```
Midnighter:
```suggestion
        value: tuple or sequence
```

src/cobra/core/reaction.py
synchon: According to https://docs.python.org/3/library/copy.html , `memo` should be a dictionary, so the type can be `Dict`. A generic `Dict` without key and value type is fine in my opinion.

src/cobra/core/reaction.py
synchon: Okay let's keep it like that for now, and if required, we can change it later.

src/cobra/core/reaction.py
synchon: Okay, yes you are correct. My bad, I missed the usage part. Let's keep it `AnyStr`.

src/cobra/core/reaction.py
```diff
@@ -333,28 +495,29 @@ def flux(self):
             check_solver_status(self._model.solver.status)
             return self.forward_variable.primal - self.reverse_variable.primal
         except AttributeError:
-            raise RuntimeError("reaction '{}' is not part of a model".format(self.id))
+            raise RuntimeError(f"reaction '{self.id}' is not part of a model")
         # Due to below all-catch, which sucks, need to reraise these.
         except (RuntimeError, OptimizationError) as err:
-            raise_with_traceback(err)
+            raise err.with_traceback()
```
Midnighter: More modern syntax:

```suggestion
            raise
```

one could also consider raising some kind of cobrapy error.

synchon:
Please rebase it on the latest `devel` after you push your final changes.

src/cobra/core/reaction.py
```diff
@@ -389,23 +552,29 @@ def reduced_cost(self):
             check_solver_status(self._model.solver.status)
             return self.forward_variable.dual - self.reverse_variable.dual
         except AttributeError:
-            raise RuntimeError("reaction '{}' is not part of a model".format(self.id))
+            raise RuntimeError(f"reaction '{self.id}' is not part of a model")
         # Due to below all-catch, which sucks, need to reraise these.
         except (RuntimeError, OptimizationError) as err:
-            raise_with_traceback(err)
+            raise err.with_traceback()
```
Midnighter: As above

```suggestion
            raise
```

src/cobra/core/reaction.py
```diff
@@ -538,53 +715,97 @@ def functional(self):
         return True
 
     @property
-    def x(self):
-        """The flux through the reaction in the most recent solution.
+    def x(self) -> float:
```
Midnighter, line 723: These have been deprecated for some time. Maybe it's time to remove them?

src/cobra/core/reaction.py
```diff
@@ -538,53 +715,97 @@ def functional(self):
...
+        .. deprecated ::
+        Use reaction.flux instead.
         """
         warn("Please use reaction.flux instead.", DeprecationWarning)
         return self.flux
 
     @property
-    def y(self):
-        """The reduced cost of the reaction in the most recent solution.
+    def y(self) -> float:
```
Midnighter, line 741: Time to remove `.y`?

src/cobra/core/reaction.py
```diff
@@ -1066,19 +1472,23 @@ def build_reaction_from_string(
 
         Parameters
         ----------
-        reaction_str : string
+        reaction_str : str
             a string containing a reaction formula (equation)
-        verbose: bool
+        verbose: bool, default True
             setting verbosity of function
```
Midnighter:
```suggestion
        verbose: bool, optional
            Set the verbosity of the function (default `True`).
```

Midnighter (changes requested):
Couple more changes otherwise it looks great. Thanks for the big update 🙂

src/cobra/core/reaction.py
synchon: Maybe a separate PR to get rid of deprecations?

src/cobra/core/reaction.py
akaviaLab: I read [Raising exceptions](https://python-future.org/compatible_idioms.html#raising-exceptions) and it seems to indicate that Python 3 requires err.with_traceback()
Should I import traceback or something?

akaviaLab:
> Please rebase it on the latest `devel` after you push your final changes.

Resolved conflicts, which failed.
Rebasing and force pushing to avoid having 6 copies of the commits

src/cobra/core/reaction.py
Midnighter: Separate PR

src/cobra/core/reaction.py
Midnighter: Take a look at the official tutorial. I think the futures docs don't give the most recent impression since that is focused on compatible 2 & 3 syntax.

https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

> If no expressions are present, [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) re-raises the exception that is currently being handled, which is also known as the active exception. If there isn’t currently an active exception, a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) exception is raised indicating that this is an error.

src/cobra/core/reaction.py
```diff
@@ -1,44 +1,52 @@
...
-from cobra.core.gene import GPR, Gene
-from cobra.core.metabolite import Metabolite
-from cobra.core.object import Object
-from cobra.exceptions import OptimizationError
-from cobra.manipulation import remove_genes
-from cobra.util.context import get_context, resettable
-from cobra.util.solver import (
+if TYPE_CHECKING:
+    from optlang.interface import Variable
+    from . import Model, Solution
```
synchon: Apologies for not pointing out earlier: for imports under `TYPE_CHECKING`, you will need the full import path like so: `from cobra import Model, Solution`.

src/cobra/core/reaction.py
```diff
@@ -1,44 +1,52 @@
...
+    from ..summary import ReactionSummary
```
synchon: `from cobra.summary import ReactionSummary`

src/cobra/core/reaction.py
```diff
@@ -61,22 +71,32 @@ class Reaction(Object):
...
+        The identifier to associate with this reaction (default None).
+    name : str, optional
+        A human readable name for the reaction (default "").
+    subsystem : str, optional
+        Subsystem where the reaction is meant to occur (default "").
     lower_bound : float
-        The lower flux bound
-    upper_bound : float
-        The upper flux bound
+        The lower flux bound (default 0).
```
synchon: `... (default 0.0).`

src/cobra/core/reaction.py
```diff
@@ -150,57 +192,122 @@ def forward_variable(self):
...
 
         Assuming that the objective of the associated model is summation of
         fluxes from a set of reactions, the coefficient for each reaction
         can be obtained individually using this property. A more general way
         is to use the `model.objective` property directly.
+
+        Returns
+        ------
+        float
+            Linear coefficient if this reaction has any, or 0 otherwise.
```
synchon: ... 0.0 otherwise.

src/cobra/core/reaction.py
```diff
@@ -458,31 +629,28 @@ def _update_genes_from_gpr(self, new_gene_names: set = None) -> None:
...
         Returns
         -------
-        string
+        str
 
         """
         return self._gpr.to_string()
 
     @gene_reaction_rule.setter
     def gene_reaction_rule(self, new_rule: str):
```
synchon: Please add the return type.

src/cobra/core/reaction.py
```diff
@@ -812,14 +1162,22 @@ def add_metabolites(self, metabolites_to_add, combine=True, reversibly=True):
...
-            Describes behavior a metabolite already exists in the reaction.
+            Describes behavior if a metabolite already exists in the reaction (default
+            True).
             True causes the coefficients to be added.
             False causes the coefficient to be replaced.
 
         reversibly : bool
             Whether to add the change to the context to make the change
-            reversibly or not (primarily intended for internal use).
+            reversibly or not (primarily intended for internal use). Default is True.
```
synchon, line 1172: ... (default True).

src/cobra/core/reaction.py
```diff
@@ -937,14 +1299,14 @@ def subtract_metabolites(self, metabolites, combine=True, reversibly=True):
...
 
         combine : bool
-            Describes behavior a metabolite already exists in the reaction.
+            Describes behavior if a metabolite already exists in the reaction (default
+            True).
             True causes the coefficients to be added.
             False causes the coefficient to be replaced.
 
-        reversibly : bool
+        reversibly : bool, defalt True
```
synchon: Please move the `default True` to the description as `... (default True).`

src/cobra/core/reaction.py
```diff
@@ -1069,19 +1482,23 @@ def build_reaction_from_string(
...
-            for reversible reaction arrows
-        term_split : string
-            dividing individual metabolite entries
+            setting verbosity of function (default True)
+        fwd_arrow : AnyStr, optional
+            Str or bytes that encode forward irreversible reaction arrows (default None).
+        rev_arrow : AnyStr, optional
+            Str or bytes that encode backward irreversible reaction arrows (default None).
+        reversible_arrow : AnyStr, optional
+            Str or bytes that encode reversible reaction arrows (defualt None).
```
synchon: A small typo: `... (default True).`

src/cobra/test/test_core/test_core_reaction.py
```diff
@@ -53,8 +53,11 @@ def test_gpr_uppercase():
 
 
 @pytest.mark.parametrize("input_gpr", ["(a1 or a2", "(forT or "])
-def test_gpr_malformed(input_gpr):
-    # Malformed GPR strings will lead to empty GPRs with no genes
+def test_gpr_malformed(input_gpr: str) -> None:
+    """Test ability to deal with malformed GPR.
+
+    Malformed GPR strings will lead to empty GPRs with no genes.
```
synchon, line 59: Please add a Parameters section here.

synchon (changes requested):
A few more changes.

src/cobra/core/reaction.py
akaviaLab: Is this a stylistic choice or a syntax issue?
Because it seems to work like this.

src/cobra/core/reaction.py
akaviaLab: Same question as above.

Midnighter (approved):

src/cobra/core/reaction.py
synchon: So, static type checkers like `mypy` need to import `cobra` just like `pytest` does for testing. `TYPE_CHECKING` is respected by static type checkers. From my experience of using `mypy`, it reports error. Since, we don't have a static type checking in CI as of now, it will pass the checks. It will be good to have it for future.

src/cobra/core/reaction.py
synchon: See above comment.

src/cobra/core/reaction.py
akaviaLab: Done

synchon:
Thank you for your contribution!

synchon (approved):

Midnighter:
I don't follow exactly why you made those changes in the last commit.

akaviaLab:
Hi @Midnighter - The fist change is based on the conversation linked in https://github.com/opencobra/cobrapy/pull/1171#discussion_r825337513, where I understood @synchon would like these last changes to happen.
The change to err is because my linter was complaining that it was not used.
The third change is just line formatting.

Midnighter:
That makes sense. Thank you.

PR #1173, opened 2022-03-08
Refactor/core group

akaviaLab:
updated group.py and test_group.py to Python 3.6+ compliant code.

### DIFF
--- a/src/cobra/core/group.py
+++ b/src/cobra/core/group.py
@@ -1,9 +1,6 @@
-# -*- coding: utf-8 -*-
-
 """Define the group class."""
 
-from __future__ import absolute_import
-
+from typing import Iterable, Optional
 from warnings import warn
 
 from cobra.core.dictlist import DictList
@@ -44,7 +41,14 @@ class Group(Object):
 
     KIND_TYPES = ("collection", "classification", "partonomy")
 
-    def __init__(self, id, name="", members=None, kind=None):
+    def __init__(
+        self,
+        id: str,
+        name: Optional[str] = "",
+        members: Optional[Iterable] = None,
+        kind: Optional[str] = None,
+    ):
+        """Initialize the group object."""
         Object.__init__(self, id, name)
 
         self._members = DictList() if members is None else DictList(members)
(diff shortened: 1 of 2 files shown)


Midnighter:
As always, thank you for your contributions to cobrapy. I'm a bit overwhelmed with all this activity 🙂 It's way more than what I usually spend on cobrapy. @synchon can you maybe review these code updates? You've done a lot of them yourself and I would really appreciate the help.

synchon:
@Midnighter yeah sure, I can review the updates.

src/cobra/core/group.py
```diff
@@ -1,9 +1,6 @@
-# -*- coding: utf-8 -*-
-
 """Define the group class."""
 
-from __future__ import absolute_import
-
+from typing import Iterable, Optional
 from warnings import warn
 
 from cobra.core.dictlist import DictList
```
synchon: Please make the import as `from .dictlist import DictList` . Also, you can use this "relative import" for other cases as well.

src/cobra/core/group.py
```diff
@@ -44,7 +41,14 @@ class Group(Object):
 
     KIND_TYPES = ("collection", "classification", "partonomy")
 
-    def __init__(self, id, name="", members=None, kind=None):
+    def __init__(
+        self,
+        id: str,
+        name: Optional[str] = "",
```
synchon: As the default value is not `None`, please use `str` as the type and mention that the default value is `""` in the docstring.

src/cobra/core/group.py
```diff
@@ -54,55 +58,52 @@ def __init__(self, id, name="", members=None, kind=None):
...
-    def __len__(self):
+    def __len__(self) -> int:
+        """Get length of group."""
         return len(self._members)
 
     # read-only
     @property
-    def members(self):
+    def members(self) -> DictList:
+        """Get members of the group."""
```
synchon: Would be great if you could please add a detailed docstring stating the arguments and returns for it.

src/cobra/core/group.py
```diff
@@ -54,55 +58,52 @@ def __init__(self, id, name="", members=None, kind=None):
...
         return self._members
 
     @property
-    def kind(self):
+    def kind(self) -> str:
+        """Return the group kind."""
```
synchon: Same as the previous comment.

src/cobra/core/group.py
```diff
@@ -54,55 +58,52 @@ def __init__(self, id, name="", members=None, kind=None):
...
         return self._kind
 
     @kind.setter
-    def kind(self, kind):
+    def kind(self, kind: str) -> None:
+        """Set the group kind."""
```
synchon: Same as the previous comment along with the exception it raises.

synchon (changes requested):
Just a few comments before it's ready. Also, would be nice if you could please run `tox -e flake8` to fix the `flake8` complaints.

src/cobra/core/group.py
akaviaLab: I'm a bit confused. There are no arguments, and it returns the members.
Do you want me to write something like
```
Returns
-------
DictList
A dictlist containing the members of the group
```

synchon: Yes, sorry about the "arguments" part, but I wanted it to be a general comment. What you write above is correct for this case.

synchon:
Looks like the PR is not rebased on the latest `devel`. Can you please rebase it on the latest `devel`?

akaviaLab:
Should be rebased now. The commits appear four times because that's what rebasing does on my computer. Gah.

synchon (approved):
Looks good. Thank you for the contribution!

PR #1175, opened 2022-03-10
Refactor/core species

akaviaLab:
Update species.py to Python 3.6+
Pretty straightforward, since this file is small - remove future, add docstrings and type

### DIFF
--- a/src/cobra/core/species.py
+++ b/src/cobra/core/species.py
@@ -1,14 +1,20 @@
-# -*- coding: utf-8 -*-
+"""Define the Species class, used as a base for Gene and Metabolite."""
 
-from __future__ import absolute_import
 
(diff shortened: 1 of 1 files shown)


src/cobra/core/species.py
```diff
@@ -19,35 +25,79 @@ class Species(Object):
...
+        Returns
+        -------
+        Species
+            A copy of the species.
         """
         return deepcopy(self)
 
     @property
-    def model(self):
+    def model(self) -> Union[None, "Model"]:
```
Midnighter:
```suggestion
    def model(self) -> Optional["Model"]:
```

Midnighter (approved):
Cheers 🙂

PR #1176, opened 2022-03-10
update object.py to Python 3.6+

akaviaLab:
Remove future, object dependencies, added type checking and documentation

### DIFF
--- a/src/cobra/core/object.py
+++ b/src/cobra/core/object.py
@@ -1,31 +1,53 @@
-# -*- coding: utf-8 -*-
+"""Define base Object class in Cobra."""
 
-from __future__ import absolute_import
+from typing import Optional
 
 
-class Object(object):
-    """Defines common behavior of object in cobra.core"""
+class Object:
+    """Defines common behavior of object in cobra.core."""
 
-    def __init__(self, id=None, name=""):
-        """A simple object with an identifier
+    def __init__(self, id: Optional[str] = None, name: str = "") -> None:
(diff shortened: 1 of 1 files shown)


src/cobra/core/object.py
```diff
@@ -35,29 +57,78 @@ def id(self, value):
...
-        return "<%s %s at 0x%x>" % (self.__class__.__name__, self.id, id(self))
+    def __repr__(self) -> str:
+        """Return string representation of Object, with class.
+
+        Returns
+        -------
+        str
+            Composed of class.name, id and hexadecimal of id.
+        """
+        return f"<{self.__class__.__name__} {self.id} at 0x{id(self):x}>"
```
Midnighter:
```suggestion
        return f"<{self.__class__.__name__} {self.id} at {id(self):#x}>"
```

src/cobra/core/object.py
```diff
@@ -1,31 +1,53 @@
...
+        Objects will have notes and _annotation as dicitionaries, initialized as empty
+        dictionaries.
         """
         self._id = id
         self.name = name
 
-        self.notes = {}
-        self._annotation = {}
+        self.notes = dict()
+        self._annotation = dict()
```
Midnighter: Prefer the literal instance when possible.

```suggestion
        self.notes = {}
        self._annotation = {}
```

Midnighter (approved):
👍🏼

PR #1178, opened 2022-03-12
Change safety command, adjust timing

Midnighter:
* Tox configuration for safety didn't work out the way I planned so I propose to just always ignore the numpy vulnerabilities
* We've had a lot of CI action lately and problems with queuing (especially Mac nodes) so I propose
    * to increase the workflow time to the maximum possible 360 minutes
    * limit the test suite step to an hour

My reason being that I can see GLPK getting stuck on specific solutions somewhat randomly. 60 minutes is already generous. The tests should run in a few minutes really.


synchon (approved):
Okay let's try this out and see if this resolves the issue.

PR #1181, opened 2022-03-13
make model.copy() handle GPRs correctly

cdiener:
* [X] tests added/passed

This ensures that GPRs are handled correctly in `model.copy()`. Prior to this the copy would leave a dangling reference to the old GPR body.

I also fixed one banchmark test for GRPs which was set up to fail and way too slow.

### DIFF
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -4,6 +4,8 @@
 
 ## Fixes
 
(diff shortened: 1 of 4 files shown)


Midnighter (approved):
Cheers, thank you for the quick fix.

PR #1183, opened 2022-03-14
Fix copying

akaviaLab:
* [X] fix #1179 
Copying should use reaction.update_genes_from_gpr() instead of the private ._genes.
update_genes_from_gpr() should have context.

### DIFF
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -1118,7 +1118,7 @@ class Model(Object):
             for gene in self.genes:
                 gene._reaction.clear()
             for rxn in self.reactions:
-                rxn._update_genes_from_gpr()
+                rxn.update_genes_from_gpr()
                 for met in rxn._metabolites:
                     met._reaction.add(rxn)
 
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -17,7 +17,11 @@ from cobra.core.metabolite import Metabolite
 from cobra.core.object import Object
 from cobra.core.reaction import Reaction
 from cobra.core.solution import get_solution
-from cobra.medium import find_boundary_types, find_external_compartment, sbo_terms
+from cobra.medium import (
+    find_boundary_types,
+    find_external_compartment,
+    sbo_terms,
+)
 from cobra.util.context import HistoryManager, get_context, resettable
 from cobra.util.solver import (
     add_cons_vars_to_problem,
@@ -358,7 +362,7 @@ class Model(Object):
             new.genes.append(new_gene)
 
         new.reactions = DictList()
-        do_not_copy_by_ref = {"_model", "_metabolites", "_genes"}
+        do_not_copy_by_ref = {"_model", "_metabolites", "_genes", "_gpr"}
         for reaction in self.reactions:
             new_reaction = reaction.__class__()
             for attr, value in reaction.__dict__.items():
@@ -371,10 +375,7 @@ class Model(Object):
                 new_met = new.metabolites.get_by_id(metabolite.id)
                 new_reaction._metabolites[new_met] = stoic
                 new_met._reaction.add(new_reaction)
-            for gene in reaction._genes:
-                new_gene = new.genes.get_by_id(gene.id)
-                new_reaction._genes.add(new_gene)
-                new_gene._reaction.add(new_reaction)
+            new_reaction.gpr = reaction.gpr.copy()
 
         new.groups = DictList()
         do_not_copy_by_ref = {"_model", "_members"}
@@ -1120,8 +1121,6 @@ class Model(Object):
                 rxn._update_genes_from_gpr()
                 for met in rxn._metabolites:
                     met._reaction.add(rxn)
-                for gene in rxn._genes:
-                    gene._reaction.add(rxn)
 
         # point _model to self
         for l in (self.reactions, self.genes, self.metabolites, self.groups):
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -17,11 +17,7 @@ from cobra.core.metabolite import Metabolite
 from cobra.core.object import Object
 from cobra.core.reaction import Reaction
 from cobra.core.solution import get_solution
-from cobra.medium import (
-    find_boundary_types,
-    find_external_compartment,
-    sbo_terms,
-)
+from cobra.medium import find_boundary_types, find_external_compartment, sbo_terms
 from cobra.util.context import HistoryManager, get_context, resettable
 from cobra.util.solver import (
     add_cons_vars_to_problem,
--- a/src/cobra/core/reaction.py
+++ b/src/cobra/core/reaction.py
@@ -597,6 +597,7 @@ class Reaction(Object):
         ----------
         new_gene_names: set
         """
+        context = get_context(self)
         if new_gene_names is None:
             if self._gpr.body is not None:
                 new_gene_names = self._gpr.genes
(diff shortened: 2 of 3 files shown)


src/cobra/core/reaction.py
```diff
@@ -590,13 +590,14 @@ def genes(self) -> FrozenSet:
         """
         return frozenset(self._genes)
 
-    def _update_genes_from_gpr(self, new_gene_names: Optional[Set] = None) -> None:
+    def update_genes_from_gpr(self, new_gene_names: Optional[Set] = None) -> None:
```
cdiener, line 593: Shouldn't `new_gene_names` be a dict? Wouldn't you need the old name and new name for each gene here to rename them?

src/cobra/core/reaction.py
```diff
@@ -908,16 +916,13 @@ def copy(self) -> "Reaction":
...
-        for i in self._genes:
-            i._model = None
         # now we can copy
         new_reaction = deepcopy(self)
+        new_reaction.gpr = self.gpr.copy()
         # restore the references
         self._model = model
         for i in self._metabolites:
             i._model = model
-        for i in self._genes:
```
cdiener, line 919: This has to stay for SBML compatibility. We/the SBML spec do explicitly allow genes to be part of a reaction even if the reaction has no GPR (and even if they don't appear in the GPR).

src/cobra/core/reaction.py
```diff
@@ -609,13 +610,20 @@ def _update_genes_from_gpr(self, new_gene_names: Optional[Set] = None) -> None:
             model_genes = self._model.genes
             self._genes = set()
             for g_id in new_gene_names:
```
cdiener, line 612: Wait, so it will add genes even if they don't appear in the GPR? That does not seem to fit the name of the function...

src/cobra/core/reaction.py
```diff
@@ -908,16 +916,13 @@ def copy(self) -> "Reaction":
         self._model = None
         for i in self._metabolites:
             i._model = None
-        for i in self._genes:
-            i._model = None
         # now we can copy
         new_reaction = deepcopy(self)
+        new_reaction.gpr = self.gpr.copy()
```
cdiener, line 921: Not necessary. The deepcopy above already guarantees that `Reaction.gpr` was copied as well. This even applies to a normal `copy()` with the PR that was merged before.

Also I think the slightly more pythonic way for those things is to overwrite `Object.__copy__` instead of adding a `Object.copy` function.

src/cobra/core/model.py
```diff
@@ -358,7 +362,7 @@ def copy(self):
             new.genes.append(new_gene)
 
         new.reactions = DictList()
-        do_not_copy_by_ref = {"_model", "_metabolites", "_genes"}
+        do_not_copy_by_ref = {"_model", "_metabolites", "_genes", "_gpr"}
```
cdiener, line 361: not necessary, was already fixed in the previous PR. The default copy on a ref now works for the GPR class.

src/cobra/core/model.py
```diff
@@ -371,10 +375,7 @@ def copy(self):
                 new_met = new.metabolites.get_by_id(metabolite.id)
                 new_reaction._metabolites[new_met] = stoic
                 new_met._reaction.add(new_reaction)
-            for gene in reaction._genes:
-                new_gene = new.genes.get_by_id(gene.id)
-                new_reaction._genes.add(new_gene)
-                new_gene._reaction.add(new_reaction)
+            new_reaction.gpr = reaction.gpr.copy()
```
cdiener, line 374: Not necessary, already fixed in the previous PR. See the tests that pass already.

cdiener (changes requested):
So, I don't dully understand what this Pr is supposed to do. Could you add some more explanation to the PR test and describe the behavior that you want to achieve?

src/cobra/core/reaction.py
akaviaLab: No. This is a bit confusing, because it is a long revamp of the gene_reaction_rule setter.
Before I used GPR() as class, gene_reaction_rule setter would parse the gene_rule, and then use split and such like to get the gene_names. Then it would copy the old_names, and compare it to gene_names.
If the reaction had a model, it would either point to genes in the model, or add new genes that were not in the model. If the reaction didn't have a model, it would just create new genes as Gene(gene_id). Then genes in old_genes, not in new_genes would be removed from the reaction (and possibly the model).

update_genes_from_gpr() tries to do the roughly the same thing - update reaction._genes with new_genes and remove old genes. Now that you mention it, new_gene_names should probably NOT be a parameter, since it is taken from GPR() (and then if new_gene_names is None: will always be true, so it can be removed).

src/cobra/core/reaction.py
akaviaLab: assigning reaction.gpr to a value also forces update_genes_from_gpr(). I can make it explicit, by having the copy happen in the attributes, and then forcing update_genes_from_gpr() manually.

src/cobra/core/model.py
akaviaLab: Right, got it.

src/cobra/core/reaction.py
akaviaLab: Hmm.
I think one discussion with @Midnighter gave me the impression that GPR should be the only thing updated, so gene_reaction_rule and genes are always derived from it. That led to the function update_genes_from_gpr()
If you set up a model with a reaction that has genes and no GPR, when you run update_genes_from_gpr() you'll wipe out the genes field of that reaction. update_genes_from_gpr is called in various time points, including renaming genes and others...

So maybe we should think on how to do update without wiping some genes? Do you have any test case for this SBML compatibility?

src/cobra/core/reaction.py
cdiener: Okay got it. Doing it explicitly seems to make more sense than having it as a side effect of the assignment.

src/cobra/core/reaction.py
cdiener: I am fairly certain @Midnighter did not want to break SBML compatibility. I think adding genes from the GPR that don't exist yet is fine and removing the old gene when renaming is fine as well. But it should never delete genes otherwise. Otherwise we would read SBML models incorrectly (deleting genes that may appear in the SBML version silently).

src/cobra/core/reaction.py
cdiener: Yeah, I agree that the gene names should be taken from the GPR only here and not via an argument. I think the behavior looks good though. I did not see it deleting genes here. However it tries to add genes even if they already exist in the reaction genes. So it seems that may be skipped probably.

src/cobra/core/reaction.py
cdiener: Though you only delete the genes from the reaction, not the model, right? This may be okay but it leads to some odd behavior. I tracked it in #1185 to get some more input.

src/cobra/core/reaction.py
akaviaLab: It adds the genes to the current model if the reaction has a model, and these genes exist, but aren't in the reaction. Unless I misunderstood you?

src/cobra/core/reaction.py
akaviaLab: Well. Not exactly.
If the gene has no reactions after removal from this reaction, the function tries to remove it from the model. I could change that behavior, but see my comments on #1185

akaviaLab:
This should be addressed by #1187

PR #1187, opened 2022-03-15
Make genes and GPR consistent with SBML spec and add context manager to GRPs

akaviaLab:
* [X] fix #1183 #1185 #1157
* [X] description of feature
This removes the function gene.remove_from_model() since that function was old and weird
* [X] description of fix
Fixes the problems with copying, fixes inconsistent and inappropriate removal of genes (#1185) and started adding context to reaction._update_genes_from_gpr()

I haven't been able to add context to reaction.gene_reaction_rule (setter). If someone can add to that please, I'd appreciate it.

## New behavior

As a reference here the behavior with those changes.

<details>

```
In [1]: from cobra.test import create_test_model

In [2]: mod = create_test_model("textbook")

In [3]: mod.reactions[0].genes
Out[3]: frozenset({<Gene b0351 at 0x7f6d17a1b4f0>, <Gene b1241 at 0x7f6d17a1b460>})

In [4]: mod.reactions[0].gpr
Out[4]: cobra.core.gene.GPR('b0351 or b1241')

In [5]: mod.reactions[0].gene_reaction_rule = "b0351"

In [6]: mod.reactions[0].genes
Out[6]: frozenset({<Gene b0351 at 0x7f6d17a1b4f0>})

In [7]: mod.reactions[0].gpr
Out[7]: cobra.core.gene.GPR('b0351')

In [8]: mod.genes.b1241
Out[8]: <Gene b1241 at 0x7f6d17a1b460>

In [9]: mod.genes.b1241.reactions
Out[9]: frozenset({<Reaction ALCD2x at 0x7f6d179d2f40>})

In [10]: mod.reactions[0]
Out[10]: <Reaction ACALD at 0x7f6d17a1b490>

In [11]:                                                                                                                                                                      
Do you really want to exit ([y]/n)? y
 ~/code/cobrapy   fix_1185  ipython
Python 3.9.10 (main, Feb 22 2022, 13:54:07) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from cobra.test import create_test_model

In [2]: mod = create_test_model("textbook")

In [3]: mod.reactions[0]
Out[3]: <Reaction ACALD at 0x7f5921403550>

In [4]: mod.genes.b1241.reactions
Out[4]: 
frozenset({<Reaction ACALD at 0x7f5921403550>,
           <Reaction ALCD2x at 0x7f59213bc880>})

[...]

PR #1188, opened 2022-03-16
fix: close and join the pool on exit

Midnighter:
* [ ] fix #(issue number)
* [ ] tests added/passed

### DIFF
--- a/src/cobra/util/process_pool.py
+++ b/src/cobra/util/process_pool.py
@@ -79,8 +79,14 @@ class ProcessPool:
         exc_tb: Optional[TracebackType],
     ) -> Optional[bool]:
         """Clean up resources when leaving a context."""
+        # The `multiprocessing.Pool.__exit__` only terminates pool processes. For a
+        # clean exit, we close the pool and join the pool processes first.
+        try:
+            self._pool.close()
+            self._pool.join()
+        finally:
+            self._clean_up()
         result = self._pool.__exit__(exc_type, exc_val, exc_tb)
-        self._clean_up()
         return result
 
     def close(self) -> None:
--- a/src/cobra/flux_analysis/deletion.py
+++ b/src/cobra/flux_analysis/deletion.py
@@ -228,12 +228,13 @@ def _multi_deletion(
             )[entity]
             chunk_size = len(args) // processes
 
-            with ProcessPool(
-                processes, initializer=_init_worker, initargs=(model,)
-            ) as pool:
-                results = extract_knockout_results(
-                    pool.imap_unordered(worker, args, chunksize=chunk_size)
-                )
+            if __name__ == "__main__":
+                with ProcessPool(
+                    processes, initializer=_init_worker, initargs=(model,)
+                ) as pool:
+                    results = extract_knockout_results(
+                        pool.imap_unordered(worker, args, chunksize=chunk_size)
+                    )
         else:
             worker = dict(gene=_gene_deletion, reaction=_reaction_deletion)[entity]
             results = extract_knockout_results(map(partial(worker, model), args))
--- a/src/cobra/flux_analysis/variability.py
+++ b/src/cobra/flux_analysis/variability.py
@@ -236,15 +236,16 @@ def flux_variability_analysis(
                 # objective direction for all reactions. This creates a
                 # slight overhead but seems the most clean.
                 chunk_size = len(reaction_ids) // processes
-                with ProcessPool(
-                    processes,
-                    initializer=_init_worker,
-                    initargs=(model, loopless, what[:3]),
-                ) as pool:
-                    for rxn_id, value in pool.imap_unordered(
-                        _fva_step, reaction_ids, chunksize=chunk_size
-                    ):
-                        fva_result.at[rxn_id, what] = value
+                if __name__ == "__main__":
+                    with ProcessPool(
+                        processes,
+                        initializer=_init_worker,
+                        initargs=(model, loopless, what[:3]),
+                    ) as pool:
+                        for rxn_id, value in pool.imap_unordered(
+                            _fva_step, reaction_ids, chunksize=chunk_size
+                        ):
+                            fva_result.at[rxn_id, what] = value
             else:
                 _init_worker(model, loopless, what[:3])
                 for rxn_id, value in map(_fva_step, reaction_ids):
--- a/src/cobra/sampling/optgp.py
+++ b/src/cobra/sampling/optgp.py
@@ -168,10 +168,11 @@ class OptGPSampler(HRSampler):
             # limit errors, something weird going on with multiprocessing
             args = list(zip([n_process] * self.processes, range(self.processes)))
 
-            with ProcessPool(
-                self.processes, initializer=mp_init, initargs=(self,)
-            ) as pool:
-                results = pool.map(_sample_chain, args, chunksize=1)
+            if __name__ == "__main__":
+                with ProcessPool(
+                    self.processes, initializer=mp_init, initargs=(self,)
+                ) as pool:
+                    results = pool.map(_sample_chain, args, chunksize=1)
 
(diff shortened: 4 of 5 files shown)


synchon:
Looks like the problem with CI is still there. Maybe re-run the pipeline for the ones that failed and see if that helps?

Midnighter:
Funny that it's the Ubuntu ones that timed out this time rather than the Mac jobs.

synchon:
Yes that's the weird part and makes me believe that either our time limits are too strict as @cdiener points out in #1187 , or something funny is going on Github's workflow management end.

src/cobra/util/process_pool.py
```diff
@@ -79,8 +79,14 @@ def __exit__(
         exc_tb: Optional[TracebackType],
     ) -> Optional[bool]:
         """Clean up resources when leaving a context."""
+        # The `multiprocessing.Pool.__exit__` only terminates pool processes. For a
+        # clean exit, we close the pool and join the pool processes first.
+        try:
+            self._pool.close()
```
cdiener, line 85: Don't you need to get the result before you close and join?

cdiener, line 85: Might also wrap it in the "if name is main" clause to make sure.

cdiener (commented):
Not sure what is happening. Just some suggestions that worked for me.

src/cobra/util/process_pool.py
cdiener: NVM. That's the exit result. Not sure how terminate plays with a pool that has already been closed and joined though.

Midnighter: > Not sure how terminate plays with a pool that has already been closed and joined though.

Not 100% sure either but at least it doesn't seem to fail.

Midnighter: I'm not sure I follow you. `__name__` will always be `cobra.util.process_pool` here, won't it?

Midnighter:
> makes me believe that either our time limits are too strict

The limits work differently now. The overall workflow has 6 hours to complete which is the maximum. The individual pytest step has 1 hour to complete. Which should be plenty considering that it finishes in less than 10 minutes normally. Maybe it's rather related to what @cdiener found in https://github.com/opencobra/cobrapy/issues/1189#issuecomment-1069498379 ?

cdiener:
I don't think so. Those tests run fast on my machine and they did run fine on linux before the changes. The best way to validate it's indeed a deadlock is to run pytest without pytest-cov. If that is suddenly fast it's due to the deadlock.

synchon:
So, re-triggering the 2 failed jobs fixes one case and the other one goes into the old state. I am a bit more confident about the fact that this is something related to Github's workflow queuing strategy.

cdiener:
I did have similar behavior in micom though and I could fix it by changing the pool code.

synchon:
Re-triggering solves the problem with the last failing job. My guess would be the runner running out of computing resources, most probably RAM.

cdiener:
RAM could be because of the memory leak. We can try to set maxtasksperchild to something intermediate.

synchon:
> I did have similar behavior in micom though and I could fix it by changing the pool code.

By any chance, is `micom`'s strategy matrix simpler than `cobrapy`'s?

cdiener:
It's only 3.7-3.9 apart from that it's the same.

synchon:
> RAM could be because of the memory leak. We can try to set maxtasksperchild to something intermediate.

After quite some community forum hops, I think it's due to the way Azure does it. Some users report that after running out of computing resources, Azure does not trigger an immediate cancel and also does not report the issue. This is a broad guess and might be something else.

What I am confused about is that if there is a memory leak, how did the previous runs on older PRs run well? One thought is that the strategy matrix previously was a bit simpler and thus the runner was able to complete it.

cdiener:
Or they made the instances smaller. Hard to say...

synchon:
> It's only 3.7-3.9 apart from that it's the same.

So essentially, we run: 2 versions x 3 platforms more and I think we are allotted a fixed set and amount of computing resources for each workflow.

synchon:
> Or they made the instances smaller. Hard to say...

Highly possible. I, somehow, have had a bitter experience with Azure resources in the past. But, it might be just me.

src/cobra/util/process_pool.py
```diff
@@ -79,8 +79,14 @@ def __exit__(
         exc_tb: Optional[TracebackType],
     ) -> Optional[bool]:
         """Clean up resources when leaving a context."""
+        # The `multiprocessing.Pool.__exit__` only terminates pool processes. For a
+        # clean exit, we close the pool and join the pool processes first.
+        try:
```
synchon, line 84: Does the code in the `try` block raise exceptions? I'm thinking whether we need the `try ... finally` block at all.

src/cobra/util/process_pool.py
cdiener: That's what I thought too, but no. The `__name__` variable gets filled on execution. If you have it outside of a function what you say would apply, but within a function it gets assigned when the function is called. See [here](https://github.com/micom-dev/micom/blob/d8190304695103df10c399a079f35d89d284cf6a/micom/workflows/core.py#L48). That code works fine and actually runs in jupyter etc.

src/cobra/util/process_pool.py
Midnighter: I can't remember why I originally put it there (it was in the `close` method). My guess is that I really wanted to clean up the pickle file no matter what. I'm also no longer sure that these changes are needed. 🤷🏼‍♂️

synchon: I see. 😅

src/cobra/util/process_pool.py
Midnighter: So you want the `.close` and `.join` only to be called when this runs as `"__main__"` or what is supposed to be in the main guard? Currently, it always runs which doesn't seem like a bad idea?

cdiener:
The constant need to restart the runners is becoming a bit tedious. We should try to merge this to see if it improves things or run everything with a single core in the tests to see if that avoids the issue.

Midnighter:
Need one approving review then, please. (I prefer a real review, of course.)

src/cobra/util/process_pool.py
cdiener: Probably more important for creating the pool. If that has no main guard then all subprocesses may spawn their own pool and send workloads.

cdiener (approved):

Midnighter:
I didn't address your comment about the `__main__` guard yet. Will try to do that before merging.

src/cobra/util/process_pool.py
cdiener: Reading up join blocks the main process until all subprocesses return. So there could be some undefined behavior if a subprocess itself calls join.

Midnighter: Reading your linked code again. It seems to me that our functions calling the process pool should include the main guard. Do you agree?

cdiener: Yep, agreed.

cdiener (approved):

cdiener:
I did some testing and just putting the parallel tests in main guards seems to help: https://github.com/cdiener/cobrapy/tree/test/deadlock. Will restart the CI in this branch a few times to see if I can trigger a deadlock again. 

EDIT: Okay finished testing and can confirm the following.

- the tests stalling the CI are exclusively tests using the pool
- tests in pytest-cov are called outside of the main process (so main guards in the code base itself will lead to tests failing)
- having the pool join/close *and* having the tests in main guards does fix the issue

cdiener:
Should I try to open a new PR with the mentioned fixes or merge them in here @Midnighter?

akaviaLab:
I think this can be closed now.

PR #1193, opened 2022-03-25
initial attempt at compartments

akaviaLab:
Very initial attempt at compartments. Just to see if I'm heading in the right direction. Will not be committed as is.

### DIFF
--- a/src/cobra/core/compartment.py
+++ b/src/cobra/core/compartment.py
@@ -1,0 +1,209 @@
+"""Define the group class."""
+
+from typing import Iterable, Optional, Union, List, FrozenSet
+from warnings import warn
+
+from .dictlist import DictList
+from .object import Object
+from .. import Metabolite, Gene, Reaction, Model
+
+
+class Compartment(Object):
+    """
+    Manage groups via this implementation of the SBML group specification.
+
+    `Compartment` is a class for holding information regarding a bounded space in
+    which species are located. You can think of it as a location in a model,
+    usually representing organelles like the Endoplasmic Reticulum (ER).
+
+    Parameters
+    ----------
+    id : str
+        The identifier to associate with this group
+    name : str, optional
+        A human readable name for the group
+    members : iterable, optional
+        A DictList containing references to cobra.Model-associated objects
+        that belong to the group. Members should be metabolites or genes, where
+        reactions are inferred from metabolites.
+    compartment_type : str, optional
+        SBML Level 2 Versions 2–4 provide the compartment type as a grouping construct
+        that can be used to establish a relationship between multiple Compartment
+        objects. A CompartmentType object only has an identity, and this identity can
+        only be used to indicate that particular Compartment objects in the model
+        belong to this type. This may be useful for conveying a modeling intention,
+        such as when a model contains many similar compartments, either by their
+        biological function or the reactions they carry. Without a compartment type
+        construct, it would be impossible within SBML itself to indicate that all of
+        the compartments share an underlying conceptual relationship because each
+        SBML compartment must be given a unique and separate identity. A
+        CompartmentType has no mathematical meaning in SBML—it has no effect on
+        a model's mathematical interpretation. Simulators and other numerical analysis
+        software may ignore CompartmentType definitions and references in a model.
+    dimensions: float, optional
+        Compartments can have dimensions defined, if they are volume (3 dimensions) or
+        2 (a two-dimensional compartment, a surface, like a membrane). In theory, this
+        can be 1 or 0 dimensions, and even partial dimensions, but that will needlessly
+        complicate the math. The number of dimensions influences the size and units
+        used.
+    """
+    def __init__(
+        self,
+        id: str,
+        name: str = "",
+        members: Optional[Iterable] = None,
+        compartment_type: Optional[str] = None,
+        dimensions: Optional[float] = None,
+    ):
+        """Initialize the group object.
+
+         id : str
+            The identifier to associate with this group
+        name : str, optional
+            A human readable name for the group
+        members : iterable, optional
+            A DictList containing references to cobra.Model-associated objects
+            that belong to the group.
+        kind : {"collection", "classification", "partonomy"}, optional
+            The kind of group, as specified for the Groups feature in the SBML
+            level 3 package specification.
+        """
+        Object.__init__(self, id, name)
+
+        self._members = DictList() if members is None else DictList(members)
+        self._compartment_type = None
+        self.compartment_type = "" if compartment_type is None else compartment_type
+        # self.model is None or refers to the cobra.Model that
+        # contains self
+        self._dimensions = dimensions
+        self._model = None
+
+    def __len__(self) -> int:
+        """Get length of group.
+
+        Returns
+        -------
+        int
+            An int with the length of the group.
+
+        """
+        return len(self._members)
+
+    # read-only
+    @property
+    def members(self) -> DictList:
+        """Get members of the group.
+
+        Returns
+        -------
+        DictList
+            A dictlist containing the members of the group.
+        """
+        return self._members
+
+    @property
+    def compartment_type(self) -> str:
+        """Return the compartment type.
+
+        Returns
+        -------
+        str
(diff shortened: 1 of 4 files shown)


cdiener:
I think for efficiency it would be nice if it could derive from the Group class already implemented in core/group.py. You could add new attributes specific to the compartment.

src/cobra/core/compartment.py
```diff
@@ -4,11 +4,11 @@
...
-from .object import Object
-from .. import Metabolite, Gene, Reaction, Model
+from .group import Group
+from .. import Metabolite, Reaction, Model
 
 
-class Compartment(Object):
+class Compartment(Group):
     """
     Manage groups via this implementation of the SBML group specification.
```
matthiaskoenig: This is probably wrong. First sentence should describe `Compartment` not `Group`. Better Documentation of class needed.

src/cobra/core/compartment.py
```diff
@@ -26,35 +26,15 @@ class Compartment(Object):
...
-        self,
-        id: str,
-        name: str = "",
-        members: Optional[Iterable] = None,
-        compartment_type: Optional[str] = None,
-        dimensions: Optional[float] = None,
-    ):
+    def __init__(self, id: str, name: str = "", members: Optional[Iterable] = None,
+                 dimensions: Optional[float] = None):
         """Initialize the group object.
```
matthiaskoenig: "Initialize the Compartment object."

src/cobra/core/compartment.py
```diff
@@ -64,69 +44,32 @@ def __init__(
...
-        Object.__init__(self, id, name)
+        dimensions: float, optional
+            Compartments can have dimensions defined, if they are volume (3 dimensions) or
+            2 (a two-dimensional compartment, a surface, like a membrane). In theory, this
+            can be 1 or 0 dimensions, and even partial dimensions. The number of
+            dimensions influences the size and units used.
 
+        Raises
+        ------
+        TypeError if given anything other than metaoblites or reactions.
```
matthiaskoenig, line 56: Typo: metabolites

src/cobra/core/compartment.py
```diff
@@ -64,69 +44,32 @@ def __init__(
...
+        """
+        super().__init__(id, name, members)
+        for x in members:
+            if not isinstance(x, (Metabolite, Reaction)):
+                raise(TypeError, f"Compartments should have only "
```
matthiaskoenig: Better language `Compartments should only have metabolites or reactions as members`

src/cobra/core/compartment.py
```diff
@@ -137,7 +80,7 @@ def add_members(self, new_members: Union[str, Metabolite, Reaction,
 
         Raises
         ------
-        TypeError - if given any object other than Metaoblite or Gene
+        TypeError - if given any object other than Metaoblite or Reaction
```
matthiaskoenig, line 84: typo "Metabolite"

src/cobra/core/compartment.py
```diff
@@ -137,7 +80,7 @@ def add_members(self, new_members: Union[str, Metabolite, Reaction,
...
         """
         if isinstance(new_members, str) or hasattr(new_members, "id"):
```
matthiaskoenig, line 86: Code duplication here. Same code below again. This is not very robust and the warning message is not informative. Be more precise, e.g. "members should be provided as list`.

src/cobra/core/compartment.py
```diff
@@ -180,25 +123,64 @@ def metabolites(self) -> DictList[Metabolite]:
         return self._members.query(lambda x: isinstance(x, Metabolite))
 
     @property
-    def reactions(self) -> Optional[FrozenSet[Reaction]]:
+    def inferred_reactions(self) -> FrozenSet[Reaction]:
```
matthiaskoenig, line 127: Personally I always prefer to use names, so that code completion shows me the options below each other. I.e. `reactions_inferred` and `reactions_assigned` are already much better. Also be more precise. There could be many ways for inferral, so I would just state the method. 
I.e. better names could be: `reactions_from_metabolites` and `reactions_from_manual`

src/cobra/core/compartment.py
```diff
@@ -180,25 +123,64 @@ def metabolites(self) -> DictList[Metabolite]:
...
+
         Returns
         -------
         FrozenSet of cobra.Reactions
             Reactions that have metabolites that belong to this compartment.
         """
-        direct_set = set(self._members.query(lambda x: isinstance(x, Reaction)))
         rxn_set = set()
         for met in self.metabolites:
             rxn_set.update(met._reactions)
```
matthiaskoenig, line 145: This seems just wrong. I understand what you are doing is: "Adding all reactions with at least 1 metabolite in the compartment". What you should be doing is "Add only reactions with all metabolites in the compartment!". Please check this!

src/cobra/core/compartment.py
```diff
@@ -180,25 +123,64 @@ def metabolites(self) -> DictList[Metabolite]:
...
+
+        Returns
+        -------
+        FrozenSet of cobra.Reactions
+            Reactions that were assigned to this compartment, if any.
+        """
+        return frozenset(self._members.query(lambda x: isinstance(x, Reaction)))
+
+    @property
+    def reactions(self) -> FrozenSet[Reaction]:
```
matthiaskoenig: Don't do this! This will just confuse everybody and create so many issues. Keep the two sets separated, never mix them !!!

matthiaskoenig (changes requested):

src/cobra/core/compartment.py
```diff
@@ -64,69 +44,32 @@ def __init__(
         members : iterable, optional
             A DictList containing references to cobra.Model-associated objects
             that belong to the group.
-        kind : {"collection", "classification", "partonomy"}, optional
-            The kind of group, as specified for the Groups feature in the SBML
-            level 3 package specification.
-        """
-        Object.__init__(self, id, name)
+        dimensions: float, optional
```
Midnighter, line 48: I was curious in what cases are non-integer dimensions used, do you know?

src/cobra/core/compartment.py
```diff
@@ -4,11 +4,11 @@
...
 
 from .dictlist import DictList
-from .object import Object
-from .. import Metabolite, Gene, Reaction, Model
+from .group import Group
+from .. import Metabolite, Reaction, Model
 
 
-class Compartment(Object):
+class Compartment(Group):
```
Midnighter, line 11: I don't know if inheriting from `Group` is really desirable or we should just have a similar class interface.

Midnighter: Another point in favor is that `.members` is a bit awkward here. We rather want attributes `.reactions` and `.metabolites`.

src/cobra/core/compartment.py
```diff
@@ -180,25 +123,64 @@ def metabolites(self) -> DictList[Metabolite]:
...
+        FrozenSet of cobra.Reactions
+            Reactions that belong to this compartment, both assigned and inferred.
+        """
+        assigned_reactions = set(self.assigned_reactions)
+        return frozenset(assigned_reactions.union(self.inferred_reactions))
 
-    def __contains__(self, member: Union[Metabolite, Gene]):
-        return member.compartment is self
+    def __contains__(self, member: Union[Metabolite, Reaction]):
+        return self.members.__contains__(member)
```
Midnighter, line 168: If we keep members then more pythonic would be
```suggestion
        return member in self.members
```

Midnighter (commented):
Some initial observations. I think we should nail down a few more details on paper before actually coding it up.

src/cobra/core/compartment.py
matthiaskoenig: For some models fractional dimensions exist, often fractal dimensions.
This can occur for instance in the context of vessel geometries/trees where you can have effective geometries between 2 and 3. I am sure there are other examples. I would say for almost every application in constraint based modeling this will be an integer. For compatibility with the SBML standard we could leave this a float, otherwise we have to cast this in reading/writing.

PR #1195, opened 2022-03-27
Fix bounds setting in matlab reader

### LINKED ISSUE #1194: Unable to simulate Harvey and Harvetta models using cobrapy (Thiele, Ines, et al.)
I tried to work on the Harvetta model, in which the LB and UB on the reactions are -1,000,000 and 1,000,000. I was not even able to load the .mat file of the model using jupyter notebbok due to this error which pointed at the reaction whose LB and UB were 62168.8 and 93253.2. It said that 62168.8 cannot be greater than 1000.
**It looks like cobrapy is using the default upper bound of 1000 for the comparison*

Screenshot 2022-03-27 at 12 26 49 AM

cdiener:
* [X] fix #1194 
* [X] tests added/passed

This fixes a bug where bound setting can thrown an error in reading matlab models due to the bounds validation clashing with the configuration defaults bounds. For instance, if the rxn bounds in the matlab file are 10000, 20000 this would crash due the lower bounds being set first and exceeding the default upper bound of 1000. 

This seems to also affect the JSON, dict and potentially the SBML (?) loader. So we should check there as well.  

This PR includes an additional fix to strip whitespace from IDs in order to fix #1194.

### DIFF
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -6,6 +6,9 @@
 
 `model.copy()` will now correctly copy GPRs.
 
+Fixed an error where matlab models can not be read if their bounds exceed the configuration
+default in some cases.
+
 ## Other
 
 ## Deprecated features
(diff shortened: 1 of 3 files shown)


Midnighter (approved):
Great, glad to see a use for `.bounds` 🙂  and good work.

Issue #653, opened 2018-01-21
[Feature Request] Store collections of evidence annotation


ChristianLieven:
#### Problem description
The system of confidence scores which has been proposed in the past is simple albeit overly-simplified and difficult-to-verify. Using an evidence ontology coupled directly with cross-references to the appropriate publications would truly make metabolic models into the knowledge bases that they are always said to be.

Please consider these threads for more details:
https://github.com/opencobra/schema/issues/4
https://github.com/SBRG/ModelPolisher/issues/5

cdiener:
I would have expected that particular way (using annotations) to work. Could you give an example of an SBML model (*with FBC*) that has one of those annotations but is not read by our parser?

JuBra:
I was in the same situation wanting to update a model with the existing experimental information. As I did not really liked the way of developing the model using a spreadsheet, I wrote an editor based on cobrapy that allows you to edit and check both the model and the experimental information. Have a look at it here: https://github.com/JuBra/GEMEditor

What I basically came up with is the division of the information in two types:

1. Evidence items (see [wiki](https://github.com/JuBra/GEMEditor/wiki/Evidences))
2. Model tests (see [wiki](https://github.com/JuBra/GEMEditor/wiki/Test-cases))

All this information is stored along with the model in the sbml file, however within a different namespace.
This way I hoped to ensure the evolution of the model by being able to automatically check the latest changes against all information that has been collected before while allowing the use of the model with any software. 

I really like the idea of developing a general standard for how to represent the experimental information in the model though.

Midnighter:
>  I wrote an editor based on cobrapy that allows you to edit and check both the model and the experimental information. Have a look at it here: https://github.com/JuBra/GEMEditor

@matthiaskoenig is that closer to what you were looking for?

matthiaskoenig:
This would basically work today from a parser perspective, i.e. if one encodes evidence on species, genes, compartments and reactions via RDF annotations these should be usable. 
One problem is evidence on GPRs because there is no GPR class on which one could store annotations.

So basically, I would say this issue will be solved with rewriting the SBML parser. The underlying issue (which is not part of this here) is "how to store evidence for a model", which cannot be solved here, but is part of the reconstruction.

matthiaskoenig:
Hi all, 
I changed the topic of this issue from
'Extend parser functionality to include collections of evidence annotation' -> 'Store collections of evidence annotation'
and removed the SBML tag.

[...]

PR #1197, opened 2022-03-31
Improve mat

### LINKED ISSUE #1040: Improve the reading of MATLAB models
### Checklist

### Description

We support reading MATLAB (`.mat`) models but we only read a very minimal set of info which leads to a loss of information. The full set of available information can be [found here](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/notes/COBRAModelFields.md).

Currently dropped:
1. names for metabolites and reactions
2. all annotations for Models, Metabolites, Reactions and Genes
3. extra variables

Furthermore, we don't sanitize IDs which can lead to introducing invalid IDs that will crash optlang. Since cobrapy is no pretty standard-compliant in writing SBMl due to @matthiaskoenig's contributions it would be nice to allow full conversion of MATLAB to SBML. The inverse would be handy too but has a lower priority for me. 

## Solution

Expand the parser to include 1-3 mentioned above.

You could convert with the cobratoolbox but this solution is only accessible to individuals with a Matlab license and may not support all features supported by cobrapy (for instance SBML groups).

## Additional context

I feel this is necessary since many source models are still only published in MATLAB formats. This applies to Recon3 as well as Harvetta and Harvey. The proposal will make it possible to generate modern and standardized SBML versions of those with cobrapy.

If this is fully supported we might consider dropping the Matlab bridge as this has been shown to be unmaintainable due to infeasibility of testing and lacking maintainers. We would rather argue for using the `.mat` format for communication between cobrapy and the cobratoolbox.

akaviaLab:
* [X] fix #1040 
Improve reading of matlab models including annotations. See https://github.com/opencobra/cobratoolbox/blob/master/docs/source/notes/COBRAModelFields.md for fields added.
Fields I didn't add, and I'm not sure what they mean
all gene annotations (genes don't have annotations in cobrapy yet)
model.C, model.E. model.e*, model.D, model.d*

I haven't modified tests yet. Not sure what's necessary, but I'm thinking about making a copy of load_model which will load mat models from BIGG/BioModels, and then maybe make a test of load_model, load_model_mat and compare them.

I haven't modified the writing of mat models. If this code makes sense, I can write the inverse function.
Improve mat reading


cdiener:
That's awesome, thanks so much!

I think all cobrapy species have annotations, including genes:

```

In [3]: mod.genes[0].annotation
Out[3]: {}
```

Empty in this case but present.

cdiener:
With the [annotated Recon3D](https://www.vmh.life/files/reconstructions/Recon/3D.01/Recon3D_301.zip) it errors:

```
In [5]: rec = load_matlab_model("Recon3DModel_301.mat")
KeyError                                  Traceback (most recent call last)
Input In [5], in <cell line: 1>()
----> 1 rec = load_matlab_model("Recon3DModel_301.mat")

File ~/code/cobrapy/src/cobra/io/mat.py:231, in load_matlab_model(infile_path, variable_name, inf)
    229 for possible_name in possible_names:
    230     try:
--> 231         return from_mat_struct(data[possible_name], model_id=possible_name, inf=inf)
    232     except ValueError:
    233         # TODO: use custom cobra exception to handle exception
    234         pass

File ~/code/cobrapy/src/cobra/io/mat.py:496, in from_mat_struct(mat_struct, model_id, inf)
    494         new_metabolite.formula = met_formulas[i]
    495     new_metabolites.append(new_metabolite)
--> 496 mat_annotations(new_metabolites, m, _d_replace=D_MET)
    497 model.add_metabolites(new_metabolites)
    499 new_genes = []

File ~/code/cobrapy/src/cobra/io/mat.py:368, in mat_annotations(target_list, mat_struct, _d_replace)
    364 annotation_matlab = list(
    365     set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
    366 )
    367 annotation_lists = [[None]] * len(annotation_matlab)
--> 368 annotation_providers = [
    369     D_REPLACE[_d_replace + "_REV"][x] for x in annotation_matlab
    370 ]
    371 for i in range(len(annotation_matlab)):
    372     annotation_lists[i] = _cell_to_str_list(mat_struct[annotation_matlab[i]][0, 0])

File ~/code/cobrapy/src/cobra/io/mat.py:369, in <listcomp>(.0)
    364 annotation_matlab = list(
    365     set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
    366 )
    367 annotation_lists = [[None]] * len(annotation_matlab)
    368 annotation_providers = [
--> 369     D_REPLACE[_d_replace + "_REV"][x] for x in annotation_matlab
    370 ]
    371 for i in range(len(annotation_matlab)):
    372     annotation_lists[i] = _cell_to_str_list(mat_struct[annotation_matlab[i]][0, 0])

KeyError: 'metKEGGID'
```

I think you don't want the `+ "_REV"` in line 369.

akaviaLab:
Fixed.
If you can figure out what everything lower than 
model.E	m x evars 
on the cobratoolbox page means, and if it is relevant to cobrapy, I'd appreciate it. I think cobrapy doesn't support any of these fields, but if I'm wrong, please let me know.

cdiener:
Oh yeah, those are additional variables used for enzyme capacity models. Not sure whether we should read them. They can be represented with optlang, but cobrapy has no methods for enzyme capacity models at this point. We may want to support model.C, model.ctrs, and model.ctrNames, which would just be added as additional constraints to the model. But this should probably be its own PR. What we may consider in this PR is to add writing annotations to Matlab as well, or did you want this in a separate PR (would be fine with me)?

cdiener:
Some of the properties don't seem to get parsed correctly. For instance, EC numbers:

```
In [97]: m["Recon3DModel"]["rxnECNumbers"][0][0][12]
Out[97]: array([array(['1.4.3.6'], dtype='<U7')], dtype=object)

In [98]: m["Recon3DModel"]["rxns"][0][0][12]
Out[98]: array([array(['13DAMPPOX'], dtype='<U9')], dtype=object)

In [99]: rec.reactions[12].annotation
Out[99]: {'pubmed': '15035803,8920635,9653080'}

In [100]: rec.reactions[12]
Out[100]: <Reaction 13DAMPPOX at 0x7fb42ae47ca0>
```

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+        annotation_lists[i] = _cell_to_str_list(mat_struct[annotation_matlab[i]][0, 0])
+        if annotation_matlab[i] == "rxnReferences":
+            # noinspection PyUnresolvedReferences
+            annotation_lists[i] = [
+                x.replace("PMID:", "") if x else None for x in annotation_lists[i]
+            ]
+        if annotation_matlab[i] == "rxnECNumbers":
+            # if there are more than one ec code, turn them to a comma separated str
+            # noinspection PyUnresolvedReferences,PyTypeChecker
+            annotation_lists[i] = [
```
cdiener: Don't join them but keep them as lists. COBRAPY annoation value scan be lists for a single provider like `rx.annotation["ec-code"] = ["1.1.1.1", "1.2.1.1"]` for instance. Everything else won't be handled correctly by the SBML writer.

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+    annotation_matlab = list(
+        set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
+    )
+    annotation_lists = [[None]] * len(annotation_matlab)
+    annotation_providers = [D_REPLACE[_d_replace][x] for x in annotation_matlab]
+    for i in range(len(annotation_matlab)):
+        annotation_lists[i] = _cell_to_str_list(mat_struct[annotation_matlab[i]][0, 0])
+        if annotation_matlab[i] == "rxnReferences":
+            # noinspection PyUnresolvedReferences
+            annotation_lists[i] = [
```
cdiener: Need to split them apart and save them as list as well.

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+            # noinspection PyUnresolvedReferences,PyTypeChecker
+            annotation_lists[i] = [
+                ", ".join([y.strip() for y in x.split("or")])
+                if x and "or" in x
+                else None
+                for x in annotation_lists[i]
+            ]
+        if annotation_matlab[i] == "metCHEBIID":
+            # if there are more than one ec code, turn them to a comma separated str
+            # noinspection PyUnresolvedReferences,PyTypeChecker
```
cdiener: split them into a list

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+    annotation_lists = [[None]] * len(annotation_matlab)
+    annotation_providers = [D_REPLACE[_d_replace][x] for x in annotation_matlab]
+    for i in range(len(annotation_matlab)):
+        annotation_lists[i] = _cell_to_str_list(mat_struct[annotation_matlab[i]][0, 0])
+        if annotation_matlab[i] == "rxnReferences":
+            # noinspection PyUnresolvedReferences
+            annotation_lists[i] = [
+                x.replace("PMID:", "") if x else None for x in annotation_lists[i]
+            ]
+        if annotation_matlab[i] == "rxnECNumbers":
```
cdiener: elif

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+            # if there are more than one ec code, turn them to a comma separated str
+            # noinspection PyUnresolvedReferences,PyTypeChecker
+            annotation_lists[i] = [
+                ", ".join([y.strip() for y in x.split("or")])
+                if x and "or" in x
+                else None
+                for x in annotation_lists[i]
+            ]
+        if annotation_matlab[i] == "metCHEBIID":
```
cdiener: elif

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+    _d_replace: str
+        A string that points to the dictionary of converstions between MATLAB and
+        providers. Default D_MET (for metabolite).
+    """
+    annotation_matlab = list(
+        set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
+    )
+    annotation_lists = [[None]] * len(annotation_matlab)
+    annotation_providers = [D_REPLACE[_d_replace][x] for x in annotation_matlab]
+    for i in range(len(annotation_matlab)):
```
cdiener: use more descriptive names for the running variable, and rather make annotation_lists a dict and iterate over the names like

```
annotations = dict.from_keys(annotation_providers, None)
....
for name in annotation_matlab:
....
```

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+        A darray that includes the data imported from matlab file.
+    _d_replace: str
+        A string that points to the dictionary of converstions between MATLAB and
+        providers. Default D_MET (for metabolite).
+    """
+    annotation_matlab = list(
+        set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
+    )
+    annotation_lists = [[None]] * len(annotation_matlab)
+    annotation_providers = [D_REPLACE[_d_replace][x] for x in annotation_matlab]
```
cdiener: just calling it "providers" is probably sufficient.

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+            # noinspection PyUnresolvedReferences,PyTypeChecker
+            annotation_lists[i] = [
+                x.replace("CHEBI:", "") if x else None for x in annotation_lists[i]
+            ]
+    for i in range(len(target_list)):
+        annotation_dict = {
+            annotation_providers[j]: annotation_lists[j][i]
+            for j in range(len(annotation_providers))
+            if annotation_lists[j][i]
+        }
```
cdiener: Could be simplified to:
```
obj.annotation = {
    prov: annotation_lists[prov][i]
    for prov in annotation_providers
    if annotation_lists[prov][i]
}
```

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+        if len(annotation_dict):
```
cdiener: Unnecessary, if the dict is empty you can still set it. The if probably generates more overhead.

src/cobra/io/mat.py
```diff
@@ -266,11 +432,14 @@ def from_mat_struct(
...
         raise ValueError("Invalid MATLAB struct.")
 
-    if "c" in m.dtype.names:
-        c_vec = m["c"][0, 0]
-    else:
-        c_vec = None
-        warn("Objective vector `c` not found.")
+    if "metCharge" in m.dtype.names and "metCharges" not in m.dtype.names:
+        warn(
+            "This model seems to have metCharge instead of metCharges field. Will use"
```
cdiener: missing a space at the end of the first string part

src/cobra/io/mat.py
```diff
@@ -284,76 +453,123 @@ def from_mat_struct(
...
+            new_reaction.gene_reaction_rule = rxn_gene_rules[i]
         new_reactions.append(new_reaction)
+    mat_annotations(new_reactions, m, _d_replace=D_REACTION)
     model.add_reactions(new_reactions)
-    set_objective(model, coefficients)
 
-    coo = scipy_sparse.coo_matrix(m["S"][0, 0])
-    for i, j, v in zip(coo.row, coo.col, coo.data):
-        model.reactions[j].add_metabolites({model.metabolites[i]: v})
+    csc = scipy_sparse.csc_matrix(m["S"][0, 0])
```
cdiener, line 845: We should try out what happens if we do this before running `model.add_reactions(new_reactions). There are some optimizations that kick in if you add reactions with assigned stoichiometry in bulk.

src/cobra/io/mat.py
```diff
@@ -232,18 +342,74 @@ def create_mat_dict(model: Model) -> OrderedDict:
...
+                else None
+                for x in annotation_lists[i]
+            ]
+        if annotation_matlab[i] == "metCHEBIID":
+            # if there are more than one ec code, turn them to a comma separated str
+            # noinspection PyUnresolvedReferences,PyTypeChecker
+            annotation_lists[i] = [
+                x.replace("CHEBI:", "") if x else None for x in annotation_lists[i]
+            ]
+    for i in range(len(target_list)):
```
cdiener: Can use enumerate: `for i, obj in enumerate(target_list):`

cdiener (changes requested):
Looking pretty good. Some improvements to variable names and simplifications.

src/cobra/io/mat.py
akaviaLab: So single values should also be in a list?

akaviaLab (commented):
Okay. Will work on writing to matlab later.

akaviaLab:
> Oh yeah, those are additional variables used for enzyme capacity models. Not sure whether we should read them. They can be represented with optlang, but cobrapy has no methods for enzyme capacity models at this point. We may want to support model.C, model.ctrs, and model.ctrNames, which would just be added as additional constraints to the model. But this should probably be its own PR. What we may consider in this PR is to add writing annotations to Matlab as well, or did you want this in a separate PR (would be fine with me)?

If we can add these as constraints, you can tell me what each one means or how to represent it in optlang, so they can be read in. It can be this PR.
The enzyme capacity that we don't have methods in cobrapy should probably be ignored for now. 
As for writing annotations to matlab - I'm working on it.

src/cobra/io/mat.py
akaviaLab: That is (with the other suggestions you made) much much faster

akaviaLab:
Some thoughts before mission creep and/or adding stuff to this PR that doesn't belong

I've been working on comparing RECON3D SBML and RECON3D mat reading to see what's missing, and I have some helper functions I'm working on and so forth.
1) Would you like me to try and add a function to compare models? I have a kludgy function I can tidy up. I'm using it to see where the differences are, so I can focus on what to fix. Separate PR is also an option.
2) Would you like me to improve SBML reading? For example, the RECON3D SBML (which I think was written by cobratoolbox) has things in Notes that should really be in annotations. I can add some changes to SBML notes reading/writing to make sure these kind of annotations end up in annotations, or I can ignore them.
3) It looks like cobratoolbox uses subsystems for groups, and only uses groups for reactions for now. Would you like me to make the matlab reader build groups based on subsystems? What about writing?

cdiener:
> 1. Would you like me to try and add a function to compare models? I have a kludgy function I can tidy up. I'm using it to see where the differences are, so I can focus on what to fix. Separate PR is also an option.

I would say separate PR and maybe have it in the utils submodule?

> 2. Would you like me to improve SBML reading? For example, the RECON3D SBML (which I think was written by cobratoolbox) has things in Notes that should really be in annotations. I can add some changes to SBML notes reading/writing to make sure these kind of annotations end up in annotations, or I can ignore them.

We had that discussion before but Notes in SBML are specifically meant to not be parsed and are only there for a human to look at them. No machine readable information should be stored there. So ignoring those is fine. This should be fixed on the cobratoolbox side when creating those models.

> 3. It looks like cobratoolbox uses subsystems for groups, and only uses groups for reactions for now. Would you like me to make the matlab reader build groups based on subsystems? What about writing?

Yeah, the subsystem attribute itself is a leftover from legacy cobra models. Groups would definitely be the modern version for this. Probably better in a separate PR but it makes sense to read them as groups and write them from groups with specific names.

cdiener:
Oh and single annotations don't have to be in a list right now, but they can be. And maybe leave the writing to a new PR because that one is already large.

akaviaLab:
Oh and single annotations don't have to be in a list right now, but they can be. 

Can single annotations be in a list? I wrote the code in a way that both single and multiple annotations will go into a list.

The PR has writing. If you want me to separate it to a different PR, I could, but I did it before I saw your comment.

cdiener:
Yes a list is fine. You mean like `annotations = {"ec-number": ["1.2.3.4"]}`, right? That is fine.

And no need to separate the PR. I thought you hadn't started on writing yet, but it's all good.

akaviaLab:
Yes. That example is exactly how it is.

src/cobra/io/mat.py
```diff
@@ -681,6 +684,11 @@ def from_mat_struct(
         met_comps = [comps[i] for i in met_comp_index]
         met_comp_names = [comp_names[i] for i in met_comp_index]
     else:
+        logger.warning(
```
cdiener, line 731: I think it would be cool if that warning also reports the names of all unique compartments. That way one can see immediately if there are wonky ones.

akaviaLab: Like this new commit?

akaviaLab:
Not sure which additional tests I should add without making the test/data directory very big, or more complicated functions.

src/cobra/io/mat.py
```diff
@@ -218,6 +602,8 @@ def create_mat_dict(model: Model) -> OrderedDict:
     mat["grRules"] = _cell(rxns.list_attr("gene_reaction_rule"))
     mat["rxns"] = _cell(rxns.list_attr("id"))
     mat["rxnNames"] = _cell(rxns.list_attr("name"))
+    annotations_to_mat(mat, rxns.list_attr("annotation"), D_REACTION_REV)
+    notes_to_mat(mat, rxns.list_attr("notes"), D_REACTION_NOTES_REV)
```
cdiener: Notes in COBRA models should not be parsed. See the comments before. They also do not have to be a structured format and should generally be strings.

src/cobra/io/mat.py
cdiener: Yep, that looks good. I would drop the temporary assignment and just do `f"{', '.join(set(met_comps))}"`.

src/cobra/io/mat.py
```diff
@@ -179,6 +353,214 @@ def create_mat_metabolite_id(model: Model) -> str:
...
+                    v = ", ".join(v)
+                if provider_key not in providers_used:
+                    continue
+                annotation_cells_to_be[annotation_matlab[provider_key]][i] = v
+    for annotation_key, _list in annotation_cells_to_be.items():
+        if annotation_key not in mat_dict:
+            mat_dict[annotation_key] = _cell(_list)
+
+
+def notes_to_mat(
```
cdiener, line 523: So notes from cobra model should never be parsed, and they should definitely never clash with, change, or modify annotations. They were used in legacy models for annotations, but this is not supported anymore with MIRIAM and other standards. So those should only be written as strings into the matlab metNotes or rxnNotes fields with some arbitrary formattting like several lines of "key: value" for instance. Int the future notes will probably become strings and can be written as is.

src/cobra/io/mat.py
```diff
@@ -179,6 +353,214 @@ def create_mat_metabolite_id(model: Model) -> str:
...
+    _d_replace: str
+        A string that points to the dictionary of converstions between MATLAB and
+        notes. Default D_REACTION_NOTES (for reactions).
+    """
+    annotation_matlab = list(
+        set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
+    )
+    note_providers = [D_REPLACE[_d_replace][x] for x in annotation_matlab]
+    notes = dict.fromkeys(note_providers, None)
+    _pumbed_re = re.compile("PMID: ?(\\d+),?")
```
cdiener: typo, should be `_pubmed_re`. Also, it's already a variable limited to the function scope so the `_` prefix is not required.

src/cobra/io/mat.py
```diff
@@ -179,6 +353,214 @@ def create_mat_metabolite_id(model: Model) -> str:
...
+                [y.strip() for y in _each_cell.split(", ")] if _each_cell else None
+                for _each_cell in _cell_to_str_list(mat_struct[mat_key][0, 0])
+            ]
+    for i, obj in enumerate(target_list):
+        obj.annotation = {
+            prov: annotations[prov][i] for prov in providers if annotations[prov][i]
+        }
+
+
+def mat_parse_notes(
```
cdiener, line 402: Not really required but nice to have I guess. Though I would argue, they should still go into the annotation field and not into notes. Notes should be taken from the matlab metNotes and rxnNotes fields only. This would work with the SBML writer as well which may issue warnings that the annotations have no MIRIAM identifier but will not crash.

src/cobra/io/mat.py
```diff
@@ -284,139 +673,148 @@ def from_mat_struct(
...
+    mat_parse_notes(new_reactions, m, _d_replace=D_REACTION_NOTES)
 
+    csc = scipy_sparse.csc_matrix(m["S"][0, 0])
+    for i in range(csc.shape[1]):
+        stoic_dict = {
+            model.metabolites[j]: csc[j, i] for j in csc.getcol(i).nonzero()[0]
+        }
+        new_reactions[i].add_metabolites(stoic_dict)
 
-def model_to_pymatbridge(
```
cdiener, line 380: Good for me to deactivate that since we have been looking for someone to test this for 2+ years now and nobody has stepped up. We should keep in mind that this requires a new release though since it removes existing functionality.

src/cobra/io/mat.py
```diff
@@ -284,139 +673,148 @@ def from_mat_struct(
...
-    matlab : pymatbridge.Matlab, optional
-        The MATLAB workspace to which the variable will be sent. If None,
-        the variable will be sent to the same environment used in IPython
-        magics.
+    if "c" in m.dtype.names:
+        c_vec = _cell_to_float_list(m["c"][0, 0])
+        coefficients = dict(zip(new_reactions, c_vec))
+        set_objective(model, coefficients)
+    else:
+        warn("Objective vector `c` not found.")
```
cdiener: I would transfer all wanrings to logger warnings since those can be manager better and play nice with multiprocessing (thread safe).

cdiener (changes requested):
Looks fantastic. Some changes to how notes are handled to make it more future-proof. Regarding the tests we may want to look for a small Matlab model that has as many features/annotations as possible. I suspect the low coverage is due to many annotations not being present in the models, so those parser sections are not tested. Or we could try to take one of the well-annotated SBML models like textbook and write it, then read it...

src/cobra/io/mat.py
akaviaLab: Done

src/cobra/io/mat.py
akaviaLab: Typo corrected. I feel better keeping it starting with _, since there are multiple variants of it. Each one is limited in scope, but that tells me it should be limited when reading the code.

src/cobra/io/mat.py
akaviaLab: Done

src/cobra/io/mat.py
akaviaLab: Okay

src/cobra/io/mat.py
akaviaLab: Clarifying.
There are situations where the rxnReferences in matlab has something like
'PMID:9028. Also when we read this book by Blah et al, 2001, published Random House' This can't be an identifier, so I'm not moving it into annotations, but notes. I don't want to throw it away, since it is info

The mat_to_notes function takes the notes from three fields (rxnReferences as above, rxnConfidenceScores, rxnNotes), throws away anything that can go into annotations (that is in identifiers), and keeps the rest.
It is in a dictionary, because Confidence Scores are still kept as a distinct entity, even if COBRA objects dont' have it. Otherwise, we will lose data on round trip back and forth to mat, and some people care about Confidence Scores. 
SBML reader still reads notes into a dictionary as well.
It could be a string, but then read-write matlab would loose information, including confidence scores which people care about.

src/cobra/io/mat.py
akaviaLab: Okay.
This is again, not to lose information on round trips.
Something needs to happen with confidence scores, in order not to lose it.
The weird references go into notes from rxnReferences, and then they go back so they will be written in "rxnReferences". The whole _tmp_list2 section is for that.

akaviaLab (commented):
Corrected where I could.
SBML still reads NOTES into dictionary, which is partially necessary because there are notes like "Confidence Scores" that people want to keep distinct. I don't care, but I think people still do (see #653). And I think the cobratoolbox still uses confidence scores.
If there is a global decision, it should be global and apply to SBML and all other formats as well.

src/cobra/io/mat.py
cdiener: I mean, it would already lose information that does not fit the identifiers that can be mapped to matlab. If you write them to the Matlab notes it would not get lost but put in a different place (and one that is consistent with cobrapy/SBML). But the point of the notes field is that it should not contain information that you want to keep or have parsed, that is what annotations are for.

src/cobra/io/mat.py
cdiener: Oh god, yeah I understand now. I guess it makes sense to have those in the notes then. But I would leave them there. The issue with confidence scores saved outside of notes or MIRIAM annotations is that there is no way to interpret them. For instance, what do confidences of 100, 0, and -10 mean if they appear? Is a confidence of 3 in model A the same as in model B? So in the end, that information is hard to use outside a standardized system.

cdiener:
> SBML still reads NOTES into dictionary, which is partially necessary because there are notes like "Confidence Scores" that people want to keep distinct. I don't care, but I think people still do (see https://github.com/opencobra/cobrapy/issues/653). And I think the cobratoolbox still uses confidence scores.
If there is a global decision, it should be global and apply to SBML and all other formats as well.

This has more to do with cobrapy default types and incomplete support of SBML annotation namespaces. SBML already allows for custom annotations, but we don't support that yet. I think at some point, there were also plans for key-value types in FBC that would take this function. Notes in SBML are just strings with no formatting requirements, and we also don't support that  well yet. But we probably will at some point, and then all those instances that assume or require that notes are a dict will have to be adjusted.

akaviaLab:
Lots of comments, so I'm trying to have one reply. I'm writing what I understand. Feel free to correct me.

1) Should rxnNotes go into cobra.notes? I think the answer is yes.
2) Should cobra.notes go out into rxnNotes? Yes, because otherwise reading/writing a mat file will lose info. 
Missing annotation keys can be added later. I covered the ones in the cobratoolbox git page. If someone says "I have my annotation which isn't read" we can add the key.
3) Are notes still a dictionary? Yes, for now. If not, what should they be?
4) What to do with mat fields that absolutely do not match the identifiers format?
Like confidence scores or full references that do not match pubmed at all?
If they need to go in annotation instead of notes, I would give them an annotation key that starts with _ (which signifies won't work with identifiers)
So it could be something like
annotation = {
  '_confidenceScores' = 
  '_rxnReferences' = 
}
And then they can be written out to mat files and ignored otherwise.

If they are going into notes, what is the format (question 3).

cdiener:
>  1. Should rxnNotes go into cobra.notes? I think the answer is yes.

Yep, should go into `Reaction.notes`. Since they are currently a dictionary you could choose a key. May make sense to keep the source around, so it could be `rxn.notes["cobratoolbox"] = rxnNotes[i]`.

>  2. Should cobra.notes go out into rxnNotes? Yes, because otherwise reading/writing a mat file will lose info.

Yes, exactly.

>  Missing annotation keys can be added later. I covered the ones in the cobratoolbox git page. If someone says "I have my annotation which isn't read" we can add the key.

You mean reading from cobratoolbox? I think what you covered here is pretty complete. I mean writing COBRAPY notes to matlab, but this would be covered by (2) if we go that direction.

>  3. Are notes still a dictionary? Yes, for now. If not, what should they be?

Yes, see above for one idea. But any key should be fine.

> 4. What to do with mat fields that absolutely do not match the identifiers format?
> Like confidence scores or full references that do not match pubmed at all?
> If they need to go in annotation instead of notes, I would give them an annotation 
> key that starts with _ (which signifies won't work with identifiers)
> So it could be something like ...

I think it's okay to write them to notes, since they will need human intervention to be interpreted. That means giving up a perfect round trip, but that is fine. If you put them into annotations they **will be lost when saving to SBML**. So if you want them to be conserved you can keep them in the notes for now. 

> If they are going into notes, what is the format (question 3).

Using the matlab identifiers is fine, like `rxn.notes["confidence"] = ..., etc`. The only thing where I suggested a different route is writing them into matlab, where I would just save everything in the `rxnNotes` even if that means that confidence scores change their location in the matlab model to make it easier on us. If notes ever get converted to string they would enter the notes as `<p>confidence: 3</p>\n...`, but hopefully we would be able to write them to an actual key-value store at this point that is supported by SBML and our IO.

---

I am open for other ideas though. Let me know what you think.

akaviaLab:
Okay.
If notes are still a dictionary, it isn't more complicated to write them to matlab fields, and less disappearing data.

Confidence Scores are put into a key called 'Confidence Level' when reading from SBML, so that's why it is identical. So mat -> cobra -> sbml will write them in a format that can be read by sbml. cobratoolbox already has a specific field, so there's no point in putting them in rxnNotes and leaving rxnConfidenceScores empty or missing.

I can put rxnReferences that aren't MIRIAM compliant into cobrapy.notes, and then write them into cobratoolbox.rxnNotes.
cobratoolbox.rxnNotes -> cobrapy.notes["NOTES"] but it could be cobrapy.notes["rxnNotes"]

I'll add a comment that when SBML supports note strings, cobrapy should be revised, including mat reader/writer.

akaviaLab:
Changes should hopefully answer this.
For tests, thinking about making a function to download MAT models from BIGG, and then a test to download MAT and SBML, and compare them. If that sounds like a good idea, is there a way to convert BytesIO to open File-like object? I'm asking since scipy_io.loadmat needs an str or open file-like handle. If you don't think load_matlab_web() function should be in this PR, I'll find a different test.

cdiener:
In general you can use the pytest tempfile fixture to write it out then read it. BiGG wouldn't work though since the mat models are actually generated with cobrapy so they miss all the annotations not supported before. Also we want the tests to be fast if possible. So the smallest model with the most annotations we can find.

akaviaLab:
Okay then.
Maybe I'll add compressed recon3.mat if it isn't too big.

akaviaLab:
Adding test cases, and nearly done.
Question - models that don't have compartment in the met_id (like raven.mat) that match with 
if all(var in m.dtype.names for var in ["metComps", "comps", "compNames"]):

Are rewritten (because of create_mat_metabolite_id) as metabolites that do have compartment in the met id.
For example, in raven.mat
'm156' is rewritten as 'm156[Cytoplasm]'

It means that models that don't have compartments in met_ids do not make a seamless read, write, reread transfer.
Should this be happening? Should I change raven.mat or change the functions so it is rewritten the same as it was?

This seems to be a quirk of RAVEN format, so I don't know if it is even used apart from test cases.

cdiener:
Yeah, I agree with you that this is odd. I would also favor maintaining IDs the same. There isn't really a rule that compartment has to be in a metabolite ID. It's a convention that many  IDs follow, but you usually also want a real attribute for a compartment. I would vote for removing that substitution and using the original ID. My argument would be that it would be surprising if I wrote a COBRApy model to Matlab and suddenly all my metabolites got renamed.

akaviaLab:
It looks like pepole who create mat files are sloppy and/or the behavior of the cobratoolbox changed over time. Specifically, the field names for annotations keep varying a bit on the capitalization
metChEBIID, metCHEBIID, and metChebiID are things that I've seen

Likewise for other fields.

Question - do we want cobrapy to be tolerant of this? If not, I'll go with the official description in https://github.com/opencobra/cobratoolbox/blob/master/docs/source/notes/COBRAModelFields.md

If so, do you have any idea on how to do case-insensitive matching?
Right now, I run 
annotation_matlab = list(
        set(mat_struct.dtype.names).intersection(D_REPLACE[_d_replace].keys())
    )
Any ideas on how to do case-insensitive matching equivalent to this? I could do regular expression, but that seems like it would be complicated.

cdiener:
Yeah, they run no validations at all it seems. You could convert all keys in the mat struct to lower case and match on those only, maybe?

akaviaLab:
Okay. I'll output what is supposed to be the official version.

akaviaLab:
All done. Including tests for all kinds of weird cases.

src/cobra/test/test_io/test_mat.py
```diff
@@ -68,10 +69,257 @@ def test_save_matlab_model(
...
+    # str or file-stream object, hence the str conversion
+    io.save_matlab_model(mini_mat_model, str(mini_output_file))
+    io.save_matlab_model(raven_mat_model, str(raven_output_file))
+    mini_mat_model_reload = io.load_matlab_model(str(mini_output_file))
+    raven_mat_model_reload = io.load_matlab_model(str(raven_output_file))
+    assert compare_models(mini_mat_model, mini_mat_model_reload) is None
+    assert compare_models(raven_mat_model, raven_mat_model_reload) is None
+
+
+def _fix_xml_annotation_to_identifiers(model: "Model") -> None:
```
cdiener, line 100: That was a great catch. The annotations in those SBML models are indeed wrong. Though, rather than fix it in the test ad hoc we should probably fix those in a different PR in the models themselves.

src/cobra/test/test_io/test_mat.py
```diff
@@ -68,10 +69,257 @@ def test_save_matlab_model(
...
+        if "sbo" in rxn.annotation.keys():
+            rxn.annotation["SBO"] = rxn.annotation.pop("sbo")
+        if len(rxn.annotation):
+            for annot, val in rxn.annotation.items():
+                if isinstance(val, str):
+                    rxn.annotation[annot] = [val]
+    for gene in model.genes:
+        if len(gene.annotation):
+            if "ncbigi" in gene.annotation.keys():
+                gene.annotation["ncbiprotein"] = gene.annotation.pop("ncbigi")
```
cdiener, line 139: I thinks this should be `gene.annotation["ncbigene"] = ...`

src/cobra/test/test_io/test_mat.py
```diff
@@ -68,10 +69,257 @@ def test_save_matlab_model(
...
+        "kegg.compound": ["C00022"],
+    }
+    for met in mat_model.metabolites.list_attr("id"):
+        assert (
+            mat_wrong_caps_model.metabolites.get_by_id(met).annotation
+            == mat_model.metabolites.get_by_id(met).annotation
+        )
+
+
+"""
```
cdiener: What is this supposed to do? Is this just there to generate the model for the tests? In that case you can remove it here. You can create a Gist if you want and link to it in a comment.

src/cobra/test/test_io/test_sbml.py
```diff
@@ -205,7 +205,9 @@ def test_read_2(self, io_trial: Tuple[str, Model, Model, Model]) -> None:
             A tuple containing name, reference_model, test_model, rewritten_model.
         """
         name, reference_model, test_model, _ = io_trial
-        if name in ["fbc1", "mat", "cobra", "raven-mat"]:
+        if name in [
```
cdiener: Why are you changing SBML tests here?

src/cobra/io/mat.py
```diff
@@ -71,6 +179,73 @@ def _cell(x: Iterable[str]) -> np.ndarray:
     return np.array(x_no_none, dtype=object)
 
 
+def _cell_to_str_list(
+    m_cell: np.ndarray,
+    empty_value: Optional[str] = None,
+    _re: Optional[Pattern] = None,
```
cdiener: Don't use underscore naming in function arguments they are local anyway. Rather try to give them a clear and interpretable name.

src/cobra/io/mat.py
```diff
@@ -158,25 +334,239 @@ def save_matlab_model(
...
+    """
+    struct_names = [x.casefold() for x in mat_struct.dtype.names]
+    matlab_field_dict = {
+        x.casefold(): D_REPLACE[_d_replace][x] for x in D_REPLACE[_d_replace].keys()
+    }
+    _caseunfold = {x.casefold(): x for x in mat_struct.dtype.names}
+    annotation_matlab = list(set(struct_names).intersection(matlab_field_dict.keys()))
+    providers = [matlab_field_dict[x] for x in annotation_matlab]
+    annotations = dict.fromkeys(providers, None)
+    _pubmed_re = re.compile("PMID: ?(\\d+),?")
```
cdiener: i think those are constants and can go on top with the other precompiled regexes.

src/cobra/io/mat.py
```diff
@@ -158,25 +334,239 @@ def save_matlab_model(
...
+        A darray that includes the data imported from matlab file.
+    _d_replace: str
+        A string that points to the dictionary of converstions between MATLAB and
+        providers. Default D_MET (for metabolite).
+    """
+    struct_names = [x.casefold() for x in mat_struct.dtype.names]
+    matlab_field_dict = {
+        x.casefold(): D_REPLACE[_d_replace][x] for x in D_REPLACE[_d_replace].keys()
+    }
+    _caseunfold = {x.casefold(): x for x in mat_struct.dtype.names}
```
cdiener: No need to use underscore for local variables. They are not exported anyways.

src/cobra/io/mat.py
```diff
@@ -158,25 +334,239 @@ def save_matlab_model(
...
+        }
+
+    # TODO - When cobrapy.notes are revised not be a dictionary (possibly when
+    #  annotations are fully SBML compliant, revise this function.
+
+
+def mat_parse_notes(
+    target_list: List[Object],
+    mat_struct: np.ndarray,
+    _d_replace: str = D_REACTION_NOTES,
```
cdiener: Drope the underscore. Also what does the "D_" stand for? Maybe improve the naming.

release-notes/next-release.md
```diff
@@ -15,6 +18,8 @@ fail more comprehensibly.
 
 ## Other
 
+Removed pymatlib direct transfer of models to matlab process.
```
cdiener: This has to go in backwards incompatible changes.

cdiener (changes requested):
Okay, only a bunch of formatting/variable naming changes. The code is getting really hard to review at this point due to large number of commits. So I would probably recommend to leave most of those for a clean-up PR afterwards. The only thing for now would be to remove the code in the docstring that generates the wrong key model. Then we can merge and follow up with a PR to fix the wrong annotations in SBML models and another one for the variable name cleanup.

src/cobra/io/mat.py
akaviaLab: Done

src/cobra/io/mat.py
akaviaLab: Dropped and improved. The D_ was for DICT_ so that's what it is now.

src/cobra/test/test_io/test_mat.py
akaviaLab: I can do that, but then the test would fail. Remove it temporarily and return it when the models are improved?

src/cobra/test/test_io/test_mat.py
akaviaLab: Well, the model has ncbigi, not ncbigene, which is an identifier in Genbank (abbreviated as GI) for proteins. It wouldn't make sense for it to be ncbigene.

src/cobra/test/test_io/test_sbml.py
akaviaLab: Not sure why I was doing that. It doesn't fail, but I can change it back.

src/cobra/test/test_io/test_mat.py
akaviaLab: Done

akaviaLab (commented):
Fixed. Some questions remaining.

src/cobra/test/test_io/test_mat.py
cdiener: Leave it in and remove it in the next PR.

akaviaLab:
@cdiener - reverted the test_sbml.py changes
Will change the models and everything in a later PR. I'll add a note to the new PR to be merged after this one.
Yuck - this is not just changing the XML files, but changing how they are read. I'll open a discussion for that.

src/cobra/test/test_io/test_mat.py
akaviaLab: The comparison function doesn't pick up on the differences (I was using a more strict one for development), so I can remove the fixing function. The later PR will fix the models to work even with strict comparison.
OR I can just stop changing this PR and do it all in a later one.

cdiener:
Actually, my bad. I read up on it and I think the lower case providers in SBML are now the de facto standard and are allowed by identifiers.org. So annotations["sbo"] is fine and the SBML parser will consolidate into this format (so transform from "SBO"). But we can discuss this more in another PR that addresses this because it also needs some adaptations for IDs.

akaviaLab:
@cdiener - I think we're done. Would you like another developer to review this?
I have the various improvements and cleanup ready to go in another PR once this is merged.

cdiener (approved):
Let's get this in.

PR #1200, opened 2022-04-09
Remove depreceated manipulation

### LINKED ISSUE #530: Remove unused functions from manipulation module
This will be applicable when we delete the old solver interface.

Functions like `delete_model_genes` and `undelete_model_genes` are pretty much superseded by the context managers and only used by the old versions of gene deletion routines.

akaviaLab:
* [X] fix #530 
Removed deprecated manipulation functions including undelete_model_genes, get_compiled_gene_reaction_rules, find_gene_knockout_reactions
Removed related functions from test_delete.py
Removed trimmed field from model
* [X] tests modified
Modified test_delete_genes to perform the same tests with the new functions.

### DIFF
--- a/src/cobra/manipulation/delete.py
+++ b/src/cobra/manipulation/delete.py
@@ -1,14 +1,16 @@
 """Provide functions for pruning reactions, metabolites and genes."""
-from ast import And, BoolOp, Module, Name, NodeTransformer
+import logging
+from ast import And, BoolOp, Name, NodeTransformer
 from functools import partial
-from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple, Union
-from warnings import warn
+from typing import TYPE_CHECKING, List, Optional, Set, Tuple, Union
 
 from cobra.util import get_context
 
 
 if TYPE_CHECKING:
     from cobra import Gene, Metabolite, Model, Reaction
+
+logger = logging.getLogger(__name__)
 
 
 def prune_unused_metabolites(model: "Model") -> Tuple["Model", List["Metabolite"]]:
@@ -80,43 +82,17 @@ def knock_out_model_genes(
     reaction_list: list[cobra.Reaction]
         A list of cobra.Reactions that had the bounds turned to zero.
     """
-    orig_bounds = model.reactions.list_attr('bounds')
+    orig_bounds = model.reactions.list_attr("bounds")
     for gene in gene_list:
         if isinstance(gene, str):
             gene = model.genes.get_by_id(gene)
         gene.knock_out()
-    new_bounds = model.reactions.list_attr('bounds')
+    new_bounds = model.reactions.list_attr("bounds")
     reaction_list = list()
     for i, rxn in enumerate(model.reactions):
         if orig_bounds[i] != new_bounds[i]:
             reaction_list.append(rxn)
     return reaction_list
-
-
-def undelete_model_genes(model: "Model") -> None:
-    """Undo the effects of a call to `delete_model_genes` in place.
-
-    Parameters
-    ----------
-    model: cobra.Model
-        The model which will be modified in place.
-
-    """
-    if model._trimmed_genes is not None:
-        for x in model._trimmed_genes:
-            x.functional = True
-
-    if model._trimmed_reactions is not None:
-        for (
-            the_reaction,
-            (lower_bound, upper_bound),
-        ) in model._trimmed_reactions.items():
-            the_reaction.lower_bound = lower_bound
-            the_reaction.upper_bound = upper_bound
-
-    model._trimmed_genes = []
-    model._trimmed_reactions = {}
-    model._trimmed = False
 
 
 def delete_model_genes(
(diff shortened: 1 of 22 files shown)


akaviaLab:
@Midnighter @synchon @cdiener - can I get a review of this from one of you please?
Thank you

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +60,40 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
-    return [
-        r
-        for r in potential_reactions
-        if not compiled_gene_reaction_rules[r].eval(gene_set)
-    ]
+        gene.knock_out()
+    new_bounds = model.reactions.list_attr("bounds")
+    reaction_list = list()
+    for i, rxn in enumerate(model.reactions):
+        if orig_bounds[i] != new_bounds[i]:
```
cdiener, line 94: Exact comparison with floating point numbers is never a good idea.

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +60,40 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
-    warn(
-        "The function `get_compiled_gene_reaction_rules` has outlived its purpose. "
-        "It will be removed soon.",
-        DeprecationWarning,
-    )
-    return {r: r.gpr for r in model.reactions}
-
-
-def find_gene_knockout_reactions(
+def knock_out_model_genes(
```
cdiener, line 63: This looks like a major API change and not what the PR title implies. Can you add some more comment to the PR description why that was necessary?

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +60,40 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
     model: "Model",
-    gene_list: List["Gene"],
-    compiled_gene_reaction_rules: Optional[Dict["Reaction", Module]] = None,
+    gene_list: Union[List["Gene"], Set["Gene"], List[str], Set[str]],
 ) -> List["Reaction"]:
-    """Identify reactions which will be disabled when genes are knocked out.
+    """Temporarily remove the effect of genes in `gene_list`.
```
cdiener, line 67: Maybe better to say "Disable the genes in `gene_list`." It's a bit unclear what "temporarily removing the effect" of a gene would mean.

cdiener:
I was hoping we could fix the CI deadlocks before that, sorry for the delay.

src/cobra/manipulation/delete.py
akaviaLab: Sure.
My reasoning was - delete_model_genes() is marked as deprecated, and should be removed. However, it might be possible people want to set all reactions of a gene to zero. 
I couldn't see any other function doing it, so I added a new function that will do so in a context specific manner. 
If there is one that does it since remove_model_genes removes genes from model, I missed it and would be happy to point to it.

src/cobra/manipulation/delete.py
akaviaLab: So math.isclose(), maybe with the solver defined tolerance?
Something else?
https://docs.python.org/3/whatsnew/3.5.html#pep-485-a-function-for-testing-approximate-equality

src/cobra/manipulation/delete.py
akaviaLab: Will fix once other comments are answered, as to not spam the github actors

src/cobra/manipulation/delete.py
cdiener: All reactions of a gene are already knocked out with a normal `Gene.knockout`. Did you mean knockout several genes in a model? The question there would if we really need a helper since `[g.knockout() for g in gene_list()]` is already pretty concise and will work with the context manager.

akaviaLab: You're right. Several genes in a model.
My function is simply a helper which returns a list, and so is functionally equivalent to delete_model_genes() only in a context aware manner and with updated code. 
Do we want to have this helper for people who aren't that good with writing python code? If not , I can just delete it.

src/cobra/manipulation/delete.py
Midnighter: Since numpy is a direct dependency you can also use `numpy.isclose` if you like. And I don't think the solver tolerance has an impact on bound equivalence unless I overlooked something.

src/cobra/manipulation/delete.py
akaviaLab: @cdiener - any decision on this?
Keep or remove?

src/cobra/manipulation/delete.py
akaviaLab: Again. The problem of updates. I can rebase it on the new devel so there won't be a million commits and re-commits.

akaviaLab:
see #1208

PR #1204, opened 2022-04-18
fix bug in GPR.from_string()

akaviaLab:
Fixed some bugs in GPR().from_string() where it was using the unmodified string, leading to errors with GPRs that should work. Made GPRs that have empty parenthesis fail more comprehensibly.
* [X] tests added/passed
test_gpr_wrong_input()
test_gpr_that_needs_two_replacements()

### DIFF
--- a/src/cobra/core/gene.py
+++ b/src/cobra/core/gene.py
@@ -253,46 +253,6 @@ class Gene(Species):
             if not reaction.functional:
                 reaction.bounds = (0, 0)
 
-    def remove_from_model(
-        self, model: "Model" = None, make_dependent_reactions_nonfunctional: bool = True
-    ) -> None:
-        """Remove the association of gene from a model.
-
-        Parameters
-        ----------
-        model : cobra model
-           The model to remove the gene from
-        make_dependent_reactions_nonfunctional : bool
-           If True then replace the gene with 'False' in the gene
-           association, else replace the gene with 'True'
-
-
-        .. deprecated :: 0.4
-            Use cobra.manipulation.delete_model_genes to simulate knockouts
-            and cobra.manipulation.remove_genes to remove genes from
-            the model.
-
-        """
-        logger.warning(
-            "Use cobra.manipulation.remove_genes instead to remove genes "
-            "from the model."
-        )
-        logger.warning(
-            "Use cobra.manipulation.delete_model_genes to simulate knockouts."
-        )
-        if model is not None:
-            if model != self._model:
-                raise Exception(
-                    f"{repr(self)} is a member of {repr(self._model)}, "
-                    f"not {repr(model)}"
-                )
-        if self._model is None:
-            raise Exception(f"{repr(self)} is not in a model")
-        from cobra.manipulation import remove_genes
-
-        remove_genes(self._model, [self])
-        self._model = None
-
     def _repr_html_(self):
         return f"""
         <table>
--- a/src/cobra/core/gene.py
+++ b/src/cobra/core/gene.py
@@ -1,5 +1,6 @@
 """Provide functions for dealing with genes and gene product rules (GPR)."""
 
+import logging
 import re
 from ast import (
     AST,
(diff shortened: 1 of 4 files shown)


src/cobra/core/gene.py
```diff
@@ -181,10 +183,9 @@ def parse_gpr(str_expr: str) -> Tuple:
     Use GPR(string_gpr=str_expr) in the future. Because of the GPR() class,
     this function will be removed.
     """
-    warn(
+    logger.warning(
```
Midnighter: I think an actual deprecation warning was actually the right thing to do here. What made you change your mind?

src/cobra/core/gene.py
```diff
@@ -271,11 +272,13 @@ def remove_from_model(
...
         """
-        warn(
+        logger.warning(
             "Use cobra.manipulation.remove_genes instead to remove genes "
             "from the model."
         )
-        warn("Use cobra.manipulation.delete_model_genes to simulate knockouts.")
+        logger.warning(
+            "Use cobra.manipulation.delete_model_genes to simulate knockouts."
+        )
```
Midnighter: Maybe we should execute on this deprecation rather than fixing the code here?

src/cobra/core/gene.py
```diff
@@ -780,9 +787,8 @@ def eval_gpr(expr: Union[Expression, GPR], knockouts: Union[DictList, set]) -> b
         True if the gene reaction rule is true with the given knockouts
         otherwise false
     """
-    warn(
+    logger.warning(
```
Midnighter: IMO, was good as a deprecation warning.

src/cobra/core/gene.py
```diff
@@ -815,9 +821,8 @@ def ast2str(expr: Union[Expression, GPR], level: int = 0, names: dict = None) ->
     Use GPR.to_string(names=) in the future. Because of the GPR() class,
     this function will be removed.
     """
-    warn(
+    logger.warning(
```
Midnighter: IMO, was good as a deprecation warning.

src/cobra/test/test_core/test_gpr.py
```diff
@@ -208,39 +208,38 @@ def test_wrong_input_gpr_error(test_input: Union[list, set]) -> None:
...
 def test_wrong_input_gpr_warning(test_input: str) -> None:
     with pytest.warns(SyntaxWarning):
         gpr1 = GPR.from_string(test_input)
         assert gpr1.body is None
         assert len(gpr1.genes) == 0
 
 
+def test_gpr_that_needs_two_replacements() -> None:
+    gpr1 = GPR.from_string(
+        "(591001.3.peg.1891 AND 591001.3.peg.1892 " "AND 591001.3.peg.1893)"
```
Midnighter:
```suggestion
        "(591001.3.peg.1891 AND 591001.3.peg.1892 AND 591001.3.peg.1893)"
```

Midnighter (changes requested):
I like that you introduce logging in general. For some of them I suggest to stick to the deprecation warnings, though.

src/cobra/core/gene.py
akaviaLab: I was under the impression that logger.warning should be used instead of warning. I'll change it back.

src/cobra/core/gene.py
akaviaLab: You mean just raise a warning and do nothing? If not, what do you mean? I don't understand this comment.

src/cobra/core/gene.py
akaviaLab: Will fix

akaviaLab (commented):
Okay. Will change according to comments.

src/cobra/core/gene.py
Midnighter: Sorry for not being clear. I mean that since the deprecation warning has been around since version 0.4, maybe it's time to just remove the code rather than continue to refactor it?

src/cobra/core/gene.py
Midnighter: I don't have a good rule for when to use which but to my mind the `DeprecationWarning` and `SyntaxWarning` are good cases to raise as objects.

akaviaLab: The SyntaxWarning I have it as both logging and warning, to be thorough.

cdiener (approved):
Looks good to me. Don't forget to add a note in the release notes.

Midnighter (approved):

PR #1205, opened 2022-04-19
Fix deadlocks in pytest-cov runs

cdiener:
* [ ] fixes deadlocks in the CI
* [X] tests added/passed

This combines the changes in #1188 with main guards for all the parallel tests which will avoid deadlocks in pytest-cov.

### DIFF
--- a/src/cobra/sampling/optgp.py
+++ b/src/cobra/sampling/optgp.py
@@ -169,13 +169,10 @@ class OptGPSampler(HRSampler):
             # limit errors, something weird going on with multiprocessing
             args = list(zip([n_process] * self.processes, range(self.processes)))
 
-            if __name__ == "__main__":
-                with ProcessPool(
-                    self.processes, initializer=mp_init, initargs=(self,)
-                ) as pool:
-                    results = pool.map(_sample_chain, args, chunksize=1)
-            else:
-                return None
+            with ProcessPool(
+                self.processes, initializer=mp_init, initargs=(self,)
+            ) as pool:
+                results = pool.map(_sample_chain, args, chunksize=1)
 
             chains = np.vstack([r[1] for r in results])
             self.retries += sum(r[0] for r in results)
--- a/src/cobra/sampling/optgp.py
+++ b/src/cobra/sampling/optgp.py
@@ -173,6 +173,8 @@ class OptGPSampler(HRSampler):
                     self.processes, initializer=mp_init, initargs=(self,)
                 ) as pool:
                     results = pool.map(_sample_chain, args, chunksize=1)
+            else:
+                results = []
 
             chains = np.vstack([r[1] for r in results])
             self.retries += sum(r[0] for r in results)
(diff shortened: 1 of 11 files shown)


src/cobra/sampling/optgp.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy as np
 import pandas as pd
+from pydantic import NoneIsAllowedError
```
synchon: Not sure where this import is being used.

synchon (commented):
Just a minor comment about the import. Looks great and the CI works perfectly. Thanks!

Midnighter:
I had one doubt when playing around with this myself. Are you sure those tests are still properly run? Or do they pass because they are essentially skipped due to the main guard?

cdiener:
I mean they are actually passing, so I would assume they do. I am not 100% certain if they get picked up by coverage.

Midnighter:
Well,

```py
def test_foo():
    pass
```

is a valid passing test case. So tests where the `__name__` is not `"__main__"` would also simply pass.

Midnighter:
I just ran tests on devel and for this branch, coverage looks the same and timings are equivalent. So I guess it works correctly.

Midnighter:
When you address @synchon 's comment, I think it's good to go in.

Midnighter:
Actually, just to be sure I raised an exception within one of those if blocks in the tests and pytest skips right over it.

cdiener:
Yeah, you're right! This is so weird because it is exactly what pytest-cov says one should do in their own docs: https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html .

cdiener:
That makes it difficult since it is definitely a deadlock, at least when running pytest-cov. I did see some other projects have the same issue, and some of that seem to stem from [open bugs in Python itself](https://github.com/pytest-dev/pytest-cov/issues/492). The only other alternative I see is trying to separate tests and coverage and skip multiprocessing tests in coverage only...

Midnighter:
Hmm, should we just add `# pragma: no cover` to those test cases that use the `ProcessPool`?

cdiener:
The question is if it would still run it in the coverage process and exclude it afterwards, or not run it at all. There is also a chance that pytest itself might deadlock, even though I think that is unlikely.

cdiener:
Though I would like some workaround for now as this breaks our releaser. I also tested `pragma: no cover` and `pytest.mark.no_cover` and both will still run the code (just not record coverage) so it wouldn't fix our issue. We could do it with a `TEST_MP` environment variable and `pytest.mark.skipif` directive...

cdiener:
Here is an implementation based on the previous command. It seems to work.

synchon:
Looks cleaner and works. 🎉

synchon (approved):
Thanks for the effort! 🙏🏻

Midnighter (approved):
Feels like a pretty convoluted solution but I don't have a better idea. Thanks for working this out 👍🏼

cdiener:
Yeah, fixing the source issue would have been better. Maybe it will fix itself in the future...

PR #1206, opened 2022-04-25
add compare.py which contains functions for comparing.

akaviaLab:
* [X] description of feature
Added functions for comparing models, metabolites, reactions, groups, genes. This is an adaptation of a function I used to compare XML and MAT models of the same file.

I'm not sure what tests to add for this. Maybe we can replace the test.test_io.conftest.py compare_mdoels() function with this one.

* [ ] tests added/passed

### DIFF
--- a/src/cobra/util/compare.py
+++ b/src/cobra/util/compare.py
@@ -1,6 +1,8 @@
 """Comparing models, reactions, metabolites, genes and groups."""
 
-from typing import TYPE_CHECKING, Dict, Optional, Tuple, TypeVar
+from typing import TYPE_CHECKING, Dict, Optional, Tuple, TypeVar, Callable
+
+from cobra import DictList
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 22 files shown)


src/cobra/util/compare.py
```diff
@@ -0,0 +1,363 @@
...
+    same = {o for o in shared_keys if d1[o] == d2[o]}
+    return {"added": added, "removed": removed, "modified": modified, "same": same}
+
+
+def compare_state(
+    obj1: "TObject", obj2: "TObject", ignore_keys: Optional[set] = None
+) -> Tuple[bool, Dict]:
+    """Cmpare two cobra Objects (including subclasses).
+
+    Useful for Metaboite, and Gene Comparison.
```
matthiaskoenig, line 66: typo Metaboite

Midnighter:
I'm a bit unsure if this should be in the cobrapy package, part of a cookbook, or possibly a cobra-tools package or so. Let's discuss at the dev meeting!

cdiener:
Since the test suite is now separated, maybe this could also live there? There is some value to have it for testing purposes...

akaviaLab:
I'd be happy to move it to the testing suite. This really helped me working on MAT import to make sure it is the same as SBML.
If I move it to the testing suite, how thorough should the testing of the compare function be? I'm working on testing that it gives the appropriate results, but I'm not sure that is appropriate/necessary if it lives in the testing suite.

akaviaLab:
Would you like me to move the compare function to tests? I can do that.

PR #1208, opened 2022-04-27
Remove deprecated manipulation2

akaviaLab:
* [X] fix https://github.com/opencobra/cobrapy/issues/530
Removed deprecated manipulation functions including undelete_model_genes, get_compiled_gene_reaction_rules, find_gene_knockout_reactions
Removed related functions from test_delete.py
Removed trimmed field from model
Added knock_out_model_genes()
* [X] tests modified
Modified test_delete_genes to perform the same tests with the new functions.
* [X} add an entry to the [next release](https://github.com/opencobra/cobrapy/release-notes/next-release.md)

This is a cleaner version of #1200 

One decision that needs to be made is about `knock_out_model_genes() which is a simple helper function for people who want to knock out lots of genes, and aren't good with writing the list comprehension on their own.


src/cobra/manipulation/delete.py
```diff
@@ -57,113 +61,43 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
+        gene.knock_out()
+    new_bounds = model.reactions.list_attr("bounds")
+    reaction_list = list()
+    for i, rxn in enumerate(model.reactions):
+        if not (
+            np.isclose(orig_bounds[i][0], new_bounds[i][0])
+            and np.isclose(orig_bounds[i][1], new_bounds[i][1])
+        ):
+            reaction_list.append(rxn)
+    return reaction_list
```
Midnighter: Even though it increases the computational work, I think a version more in line with the object-oriented design is:

```suggestion
    return [rxn for rxn in itertools.chain.from_iterable((gene.reactions for gene in gene_list)) if not rxn.functional]
```

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +61,43 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
 
     Returns
     -------
-    list of cobra.Reaction
-       The list of cobra.Reaction objects which will be disabled.
-
-    .. deprecated:: 0.22.1
-        Internal function that has outlived its purpose.
-
+    reaction_list: list[cobra.Reaction]
```
Midnighter:
```suggestion
    list[cobra.Reaction]
```

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +61,43 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
-    )
-    return {r: r.gpr for r in model.reactions}
-
-
-def find_gene_knockout_reactions(
+def knock_out_model_genes(
     model: "Model",
-    gene_list: List["Gene"],
-    compiled_gene_reaction_rules: Optional[Dict["Reaction", Module]] = None,
+    gene_list: Union[List["Gene"], Set["Gene"], List[str], Set[str]],
```
Midnighter: When using `get_by_any`:

```suggestion
    gene_list: Iterable[Union["Gene", int, str]],
```

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +61,43 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
-    warn(
-        "The function `find_gene_knockout_reactions` has outlived its purpose. "
-        "It will be removed in the next minor version (0.23.0).",
-        DeprecationWarning,
-    )
-    potential_reactions = set()
+    orig_bounds = model.reactions.list_attr("bounds")
     for gene in gene_list:
         if isinstance(gene, str):
             gene = model.genes.get_by_id(gene)
```
Midnighter:
```suggestion
    for gene in model.genes.get_by_any(set(gene_list)):
```

src/cobra/manipulation/delete.py
```diff
@@ -185,62 +119,39 @@ def delete_model_genes(
...
         If True, then any previous deletions will be maintained in the
-        model (default True).
+        model (default True). Unused, ignored.
     disable_orphans: bool, optional
         If True, then orphan reactions will be disabled. Currently, this
-        is not implemented (default False).
+        is not implemented (default False). Unused, ignored.
+
+    .. deprecated :: 0.25
+        Use cobra.manipulation.delete_model_genes to simulate knockouts
```
Midnighter:
```suggestion
        Use cobra.manipulation.knock_out_model_genes to simulate knockouts
```

Midnighter (changes requested):
Some minor suggestions, otherwise this looks good. I think the `knock_out_model_genes` function makes sense as it conveniently returns the list of affected reactions. I'm not super happy with the name but it's descriptive and I don't have a better suggestion. Good work 👍🏼

src/cobra/manipulation/delete.py
akaviaLab: Why are there two parenthesis?

akaviaLab (commented):
One question about the list comprehension left

src/cobra/manipulation/delete.py
akaviaLab: Also, why set? In the get_by_any above?

akaviaLab: Not sure gene.reactions as written would work, since it can accept strings

akaviaLab: Pushed a version with list comprehension that works. You can pick if the most recent one [
24cbb5e] makes sense , or [66d0428](https://github.com/opencobra/cobrapy/pull/1208/commits/66d04288a2e7abe6456ca64c73f19b104f5f0276) makes more sense.

Midnighter: Double parenthesis was outer for the function call and the inner for a generator statement. This was assuming that `gene_list` would be the output of `get_by_any`, sorry, my comment wasn't clear.

I used a set to avoid duplicates which are in principle possible with a list and `get_by_any`.

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +59,35 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
-        compiled_gene_reaction_rules = {r: r.gpr for r in potential_reactions}
-
-    return [
-        r
-        for r in potential_reactions
-        if not compiled_gene_reaction_rules[r].eval(gene_set)
-    ]
+    gene_list = model.genes.get_by_any(gene_list)
+    rxn_set = set()
+    for gene in model.genes.get_by_any(gene_list):
```
Midnighter, line 86: You are replacing the `gene_list` above so `get_by_any` is not necessary again.

src/cobra/manipulation/delete.py
```diff
@@ -57,113 +59,35 @@ def prune_unused_reactions(model: "Model") -> Tuple["Model", List["Reaction"]]:
...
+        gene.knock_out()
+        rxn_set.update(gene.reactions)
+    return [rxn for rxn in list(rxn_set) if not rxn.functional]
```
Midnighter: Converting the set to a list should not be necessary since you can iterate over a set just fine.

src/cobra/manipulation/delete.py
akaviaLab: for gene in model.genes.get_by_any(set(gene_list)): fails due to type error

>       return the_id in self._dict
E       TypeError: unhashable type: 'set'

../manipulation/delete.py:86: in knock_out_model_genes
    for gene in model.genes.get_by_any(set(gene_list)):
../core/dictlist.py:108: in get_by_any
    return [get_item(item) for item in iterable]
../core/dictlist.py:108: in <listcomp>
    return [get_item(item) for item in iterable]
../core/dictlist.py:101: in get_item
../core/dictlist.py:379: TypeError

akaviaLab: I think the current version [https://github.com/opencobra/cobrapy/pull/1208/commits/ae1fdbe5beea58cc453253111914a1bd27e7b750] is the best, and object-oriented. Fixed other comments.

Midnighter (approved):

PR #1216, opened 2022-05-05
refactor: flake8 fixes for conftest files

synchon:
* [ ] fix #(issue number)
* [x] tests added/passed

This PR resolves `flake8` issues for `conftest.py` files required for `pytest`. Also, has type annotations and code style upgrades wherever required.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,17 +1,16 @@
 """Define global fixtures."""
 
-from os.path import join
 from pathlib import Path
 from pickle import load as _load
+from typing import List, Tuple
 
-from cobra import Metabolite, Model, Reaction
+from cobra import Metabolite, Model, Reaction, Solution
 from cobra.io import read_sbml_model
 from cobra.util import solver as sutil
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -1,103 +1,125 @@
...
+from cobra import Metabolite, Model, Reaction, Solution
 from cobra.io import read_sbml_model
 from cobra.util import solver as sutil
 
 
 try:
     import pytest
-    import pytest_benchmark
 except ImportError:
     pytest = None
```
Midnighter: I think you can just convert this whole thing to a simple `import pytest`. We run our test suite with pytest, we have to assume it's installed.

Midnighter (approved):
So how will you celebrate when all code passes flake8? 🙂

tests/conftest.py
synchon: I agree.

synchon:
A good beer probably lol! 😂

PR #1220, opened 2022-05-10
Mat cleanup

akaviaLab:
This does some minor changes to SBML, mostly that annotations (and others) will be loaded as lists. It also changes the example models a bit. The goal is to make sure that XML that is written to MATLAB, then reread will be identical to the original SBML loaded.
Fixes some of the FIXMEs in sbml.py saying "Should be al list"
* [x] tests added/passed


akaviaLab:
@cdiener - This is the updates we were discussing on gitter.
One question - looking at TestCobraIO, there are some that are expected to fail - if I comment the xfail section out, the only one that fails is "fbc1". Would you like me to modify these sections? I'm not sure why they exist or why is it supposed to fail.

cdiener:
Hi, thanks for the fixes. One comment before I start reviewing. For annotations, both the `{"provider": "value"}` and `{"provider": ["value", ...]}` formats are valid and for single values we probably even prefer the first version since it's a bit more concise. So there is no need to change every single-value annotation to the second form. It's okay if the mat reader returns `{"provider": ["value"]}` annotations for now. I guess you changed it to check for equality of models?

akaviaLab:
Yes, I changed it for that reason.
Also, I don't like where you have a situation that a variable is sometimes a list, and sometimes a str - every function that tries to access this variable will have to check str/list and process accordingly.

akaviaLab:
Might become irrelevant if we update the annotations, but we can decide for now.

akaviaLab:
@cdiener Ping. Can you please review this?

cdiener:
I'm still somewhat reluctant to change the test models' annotation format here mostly because the idea is to have the test models manually vetted and than adapt the code base to read them well. Tests and/or the code base still has to be able to recognize that `{"provider": "annotation"}` is the same as `{"provider": ["annotation"]}`.

akaviaLab:
Okay. We can discuss this in detail, but there are several changes to the annotation files (which perhaps should be in a separate PR)
1) Removing some spaces which were confusing Matlab/XML comparison and don't have any other effects
2) Converting ncbigi, which is not a valid identifiers.org identifier to ncbiprotein, which is.
3) Replacing some wrong annoation, using bigg.model annoation for metabolite for some reactions instead of the actual bigg.models reaction annoation
4)Removing unused compartments from the notes.xml file, since Matlab doesn't make compartments that are unused. This led to confusion when comparing XML and Matlab. I can change it back, but since I made the original notes.xml file, I thought it wouldn't matter that much.
Should I move the compare_models to test? If so, can you answer my questions there (#1206) please?

Are these the changes you meant?

cdiener:
All of those cases are fine with me. I would just revert the conversions from single annotations to list. That has to be fixed in the tests that should accept the equivalence I outlined in the previous comment. 

Regarding the comparison it sounded like @Midnighter had some opinion on that so that's why I was waiting on more feedback from him.

akaviaLab:
I'm a bit confused about the list for single annotations.
Right now, before my changes, if there is one annotation for any key, it builds it as a string. If another is added, it converts it to a list. Therefore, every function that accesses annotations has to check if it is a list or a string.
That doesn't make a lot of sense to me - it makes sense to always have it as a list, so the behavior is constant. What is the argument against doing that?
If the annotations are always a list, I don't understand what you mean about the tests.
In my revision of the metadata #1225, I think I made the code so that annotations will be constructed as a list even if given a string, so annotation read will be equal to both list and string as test cases. I don't see how to do it here.

cdiener:
Because by changing the test models to be more convenient for you, you don't change the fact that annotations are still allowed to be strings or lists for single elements. You just remove examples where they are strings from the test suite which then covers less data formats. We need to remain backwards compatible with old models in JSON or YAML format and therefore that case still has to remain supported and tested until we change it to a new annotation format. But like I said it also comes down to only modifying the test models if they are actually wrong and not if it's more convenient to write tests.

cdiener:
To be clear my comments are *only* aimed at the JSON test models you are changing. In the mat parsing you are free to return them as you wish. Also why are you changing almost all other test models in this PR (like pickle and SBML ones as well). What do those have to do with the Matlab interface?

akaviaLab:
There are a lot of changes, and things got mixed.
Let me specify them, and mark which ones make less sense to be in this PR
(although I think some of them are needed in another PR, and some make
sense to be in the same PR). Can you please comment on each one so that we
are on the same page?

c14b30690dd2c73c1ee59a1f284a7df68b190c99
Changes in mat.py (this PR)
modify tests/data/e_coli_core.xml
- ncbigi to ncbiprotein, since ncbigi is not a valid identifiers.org prefix
(Other PR)
- removed excess spaces (shouldn't have done that, should fix code to
ignore them)
- changed invalide identifiers like http://identifiers.org/bigg.reaction/ac
to http://identifiers.org/bigg.reaction/EX_ac_e - ??. Maybe I shouldn't
have done this at all?
annotations.xml
- change invalid ncbgi to ncbiprotein (other PR)
textbook.xml.gz - basically same changes as e_coli_core.xml, will process
according to decision there
sbml.py
- read annotations into list (other PR or not at all)
I saw line 1805 in sbml.py which said # FIXME: always in list, and thought
it would be a good idea
validate.py - fix according to changes in sbml.py
test_annotation.py, test_sbml.py - changed according to sbml.py
update_pickles.py (necessary because of changes in xml files, but should be
moved to other PR)

a9d7f18e44627af90a191e6ae70b35b99820764b
make sbml.py subsystem reading add partonomy, which matches the definition
of SBO:0000633 (see https://sourceforge.net/p/sbo/term-request/113/) -
other PR
update_pickles.py
changes in test_mat.py to match the reading of CHEBI (this PR)

a77f77390851e66d223faab55fe2861a59549202
mat.py - some tidying (this PR)
sbml.py - unify two separate ifs (other PR)
test_sbml.py - unify implicitly concatenated strings (other PR). Doesn't
change behavior, just code

219360ba399718cddbc4096319d6e0997619d370
tests/data/example_notes.xml - remove excess compartments, since excess
compartments are read into SBML, but not MAT. Not sure if this change
should happen, but since I made example_notes.xml I thought it would be fine
src/cobra/io/sbml.py
- read group back into subsystem to keep read/write consistent, also since
mat doesn't understand groups, only subsystem (not sure where this should
be, if at all)
- modify SBO to be in list (like line 1805, above, will modify according to
that)
mat.py - some changes (this PR)

307ec41f8be00c2aafdab90a67300058b7df0cbb
mat.py - correctly read back subsystems into groups from mat (this PR)
sbml.py - set empty values to None instead of '' (other PR?). This makes
sense to me since C++/SBML and Python seem to disagree on what is empty.
SBML does '', while Python does None.

77a1359f5d4eeaa16cf7b62391c88570e18351bf
update_pickles.py and black - needed, but need to decide on other changes

f6244f2f08722bf76e64767a8001de34bf8acbd8
black on test_annotation.py - needed, but will decide on all other changes

One reason these are all the same PR is that some of them would require a
lot of code if they were separate.
For example, if SBML should read things in lists (according to the FIXME
comment) and it is a separate PR, then Mat would have to have excess code
to deal with things that aren't lists. And then when/if SBML becomes list,
mat.py would need to remove excess code.

cdiener:
Okay, I'm onboard with everything except for the following which I'd like to discuss:

> - read annotations into list (other PR or not at all)

In general that make sense, however that would be an API breaking change so we would need to release a new version for that and it would not be backwards compatible. Since there is another PR lined up that will add new annotation formatting and will need a new version as well I would skip it in favour of the FBC3 PR.

> - read group back into subsystem to keep read/write consistent, also since
mat doesn't understand groups, only subsystem (not sure where this should
be, if at all)

Subsystems are deprecated in favor of groups and should only read in the formats that still have them like legacy SBML and Matlab. So I would not convert them here. This info would be dropped in my book since it's data in groups that the Matlab format does not support. I get that this breaks roundtripping and other may think different. My worry would be what would happen if I have "subsystem" group that uses advanced features like nested groups. How should that act?

And the unmentioned changes in `mini.json`. If `mini.json` passes the JSON schema validator in this version, it should not be changed.

akaviaLab:
Other PR for everything else that isn't mat, or this PR?
Like the changes in annotation to make it more identifiers compliant?
The list change will be saved for the appropriate metadata PR.

akaviaLab:
what would happen if I have "subsystem" group that uses advanced features like nested groups. How should that act?

The groups are only exported for reactions. If reactions have the attribute "subsystem" and groups, "subsystem" will win and be exported, "groups" will be dropped.
If there is a reaction that belongs to a group, that reactions is exported with subsystem equal to the group id/name. Any other entities in the group are not exported.

So if you have something like
Group1 = [rxn1, rxn2, Group2]
Group2 = [rxn3, met1]

The subsystems when exported would look like
rxn1 - Group1
rxn2 - Group1
rxn3 - Group2

cdiener:
I think in general it's easier to review several smaller Pars than a huge one. However, since we already discussed a lot of the things here let's just continue as it is. If you want just use a more descriptive title. 

Sorry I misunderstood the subsystem thing. Your solution makes sense.

akaviaLab:
Okay. So I think everything discussed is done.
The annotations into list should be in the metadata PR.

src/cobra/data/mini.json
```diff
@@ -26,9 +26,9 @@
         },
         {
             "annotation": {
-                "ncbigi": [
-                    "GI:1208453",
-                    "GI:1652654"
+                "ncbiprotein": [
```
cdiener, line 29: What's the difference between the mini.json in `src/cobra/data` and the one in `tests/data`. Couldn't the tests just use the first one?

src/cobra/io/mat.py
```diff
@@ -479,42 +518,36 @@ def annotations_to_mat(
...
-                        ]
-                    )
-                elif provider_key == "ec-code":
-                    v = " or ".join(v)
-                else:
-                    v = ", ".join(v)
-                if provider_key not in providers_used:
-                    continue
-                annotation_cells_to_be[annotation_matlab[provider_key]][i] = v
+        for provider_key, v in annotation_list[i].items():
```
cdiener: Better name for `v`. It's not clear from the context what that is.

src/cobra/io/mat.py
```diff
@@ -542,28 +575,26 @@ def notes_to_mat(
...
-                    v = float(v)
-                if not len(annotation_cells_to_be[annotation_matlab[provider_key]][i]):
-                    annotation_cells_to_be[annotation_matlab[provider_key]][i] = v
-                else:
-                    # References that aren't MIRIAM compliant will go to rxnNotes
-                    annotation_cells_to_be[annotation_matlab[provider_key]][i] = (
-                        annotation_cells_to_be[annotation_matlab[provider_key]][i]
-                        + f"; {v}"
-                    )
+        for provider_key, v in note_list[i].items():
```
cdiener: Again, better name for v.

src/cobra/io/mat.py
```diff
@@ -705,7 +753,9 @@ def from_mat_struct(
             new_names[new_names.index(old_field)] = new_field
             m.dtype.names = new_names
 
-    model = Model()
+    model = (
```
cdiener: This formatting looks weird.

src/cobra/io/mat.py
```diff
@@ -856,7 +903,9 @@ def from_mat_struct(
         rxn_group_names = set(rxn_subsystems).difference({None})
         new_groups = []
         for g_name in sorted(rxn_group_names):
-            group_members = model.reactions.query(lambda x: x.subsystem == g_name)
+            group_members = model.reactions.query(
+                lambda x, g_n=g_name: x.subsystem == g_n
```
cdiener: I don't think it makes sense to define default values for a lambda function. The previous one looks better to me.

src/cobra/io/sbml.py
```diff
@@ -598,7 +598,7 @@ def _sbml_to_model(
     if not libsbml.SyntaxChecker.isValidSBMLSId(model_id):
         LOGGER.error(f"'{model_id}' is not a valid SBML 'SId'.")
     cobra_model = Model(model_id)
-    cobra_model.name = model.getName()
+    cobra_model.name = model.getName() or None
```
cdiener, line 601: Above you marked SBML returning a `''` as model name by default as TODO but it seems like this fixes that, right?

cdiener (commented):
Looks good to me. Some minor comments and a doubt about duplicated models files.

src/cobra/io/sbml.py
akaviaLab: Yes. This does fix it. I can remove the TODO.

src/cobra/io/mat.py
akaviaLab: Fixed.

src/cobra/io/mat.py
akaviaLab: Done

src/cobra/io/mat.py
akaviaLab: I'll remove the TODO, and  fix the formatting

src/cobra/io/mat.py
akaviaLab: Okay. My linter complained about it. Reverted.

src/cobra/data/mini.json
akaviaLab: None.
The tests are using the one in tests, because that's how the functions are designed. The ones in src/cobra/data are example files for users installing cobra without development and without tests.
If we want to have tests rely on the files in src/cobra/data we can rewrite tests and update_pickles.py, which I'd be happy to do in a different PR.

cdiener: That sounds good to me :+1:

src/cobra/io/sbml.py
akaviaLab: Removed TODO

release-notes/next-release.md
```diff
@@ -4,6 +4,12 @@
 
 ## Fixes
 
+* Reformatted example files (e_coli_core.xml, mini_cobra.xml, mini.json, mini.yml, textbook.xml.gz) to be more compliant with identifiers.org. ncbigi is not a valid identifiers.org, so it was replaced with ncbiprotein.
+* make sbml.py subsystem reading add partonomy, which matches the definition
+of SBO:0000633 (see https://sourceforge.net/p/sbo/term-request/113/)
+* Correct reading and writing of subysstem in mat.
```
cdiener: small typo

cdiener (changes requested):
Looks good. I'll merge after the typo is fixed.

cdiener (approved):

cdiener:
Awesome, thanks so much for your patience!

PR #1221, opened 2022-05-12
Group members

akaviaLab:
This reverts commit 81717afd

It will fix part of 1214, but not necessarily all of it.


.gitignore
```diff
@@ -92,6 +92,8 @@ ipython_config.py
 .idea/**/usage.statistics.xml
 .idea/**/dictionaries
 .idea/**/shelf
+.idea/**/sonar*
```
cdiener: That seems to be specific to your setup, probably better to just ignore the full `.idea` folder.

cdiener (commented):
Makes sense to me but would be good to get a second opinion as well.

.gitignore
akaviaLab: Yeah. I'm trying, but having problems. If you can ignore this file when merging, that would be great.

akaviaLab:
Ping. Did someone want to review this?

cdiener:
I'll probably merge your MAT PR before this. If you rebase on main afterwards I'll be happy to merge this right after.

cdiener (approved):
:+1:

cdiener:
Looks, like this doesn't have any conflicts and needs no rebase, so I'll just wait for the CI to finish on the previous merge and merge this one after. Thanks!

PR #1225, opened 2022-05-29
Metadata fbc3 group

akaviaLab:
Updated version of #988. Tried to simplify and tidy the code in addition to merging.

Original description follows

- Metadata Classes: A separate directory for handling metadata information is made inside cobra/core directory. Every object derived from SBase can have meta information like annotation (CVTerms), notes, history attached to it.
- The Notes class is holding a simple string containing notes data (XHTML string) and a dictionary, synchronized with the notes string, storing key-value pair of the form <p> key : value </p> present in the notes string. One can only modify this key-value pair data, he can't add new key-value pairs inside notes because notes are not a right place to store these key-value pairs.
- The CVTerms class for storing externally linked resources to each component derived from SBase. This class is maintaining the new format annotation as well as old format annotation simultaneously, and both are kept synchronized with each other. Changing one will modify the other accordingly. This new class for annotation can handle any type of annotation data (be it the case of nested annotation or alternative annotation). It can read the old format as well as the new format annotation data from JSON and other formats. At the time of writing back the model, the new data format is used because it contains the complete data organized in the same way as SBML.
- The History class used for storing the history, validating dates, etc, is now attached to each component derived from SBase.
- The KeyValuePair class for storing key-value pairs, defined by fbc-v3.

The last three metadata objects (i.e CVTerms, History, KeyValuePair) are present inside a single attribute of SBase (Object) class and can be accessed via object.annotation.cvterms, object.annotation.history and object.annotation.key_value_data attributes. Calling simply the annotation attribute (object.annotation) will return the annotation data in old format (making it backward compatible).

**Group to JSON:** The support of the group package is extended to JSON.

**JSON schema v2:** The version2 of JSON schema has been added which defines the new format annotation, history, key-value pair, notes, group package data, user-defined constraints data and basic SBML info.
@akaviaLab  comment - not sure if the v2 schema works or is done. Seems unclear from the code.

Issues Fixed
 fixes issue https://github.com/opencobra/cobrapy/issues/954: The support of group package has been extended to JSON and other formats.
 fixes issue https://github.com/opencobra/cobrapy/issues/856: The infinity values are also enabled for storing bounds.
 fixes issue https://github.com/opencobra/cobrapy/issues/810: The sbml_info storing basic information of SBML is written to JSON to store the basic SBML document information like packages, level, version, notes, annotation attached to the SBML component etc.
 fixes issue https://github.com/opencobra/cobrapy/issues/736: If annotation is present in the form of list of list, it is first modified and then data is read.
 fixes issue https://github.com/opencobra/cobrapy/issues/706: Single annotation resource are now put into a string. While reading from JSON also, if a single resource attribute is in the form of string, it is first fixed to list and then read into the model.
 fixes issue https://github.com/opencobra/cobrapy/issues/695: The notes data is now read into a string and this string is synchronized with a dictionary storing key-value pairs present inside notes string.
 fixes issue https://github.com/opencobra/cobrapy/issues/684: The complete metadata structure (CVTerms, Notes, History) has been redesigned with backward compatibility. Old format data is read, fixed and then written in new format metadata structures. The fbc-v3 "KeyValuePair" class is also a part of this new format annotation. Its corresponding class and its support to JSON and other format has been added. However, the SBML parser for it has to be updated when libsbml adds support for fbc-v3.
 fixes issue https://github.com/opencobra/cobrapy/issues/937: The annotation format has been updated, which is backward compatible too.
 JSON schema v1: The JSON schema version 1 is modified to resolve some pre-existing issues.

Tests


akaviaLab:
Revamped the structure, so the CVTerm and CVTerms are using simplified containers like UserList and not MutableMapping (which lead to complicated code). Need to add some test and code documentation, which I'll do later.

akaviaLab:
@matthiaskoenig - any chance you could review please?

akaviaLab:
closed, see #1237

Issue #684, opened 2018-03-23
Update storage of annotations


matthiaskoenig:
Hi all, 
there are currently some issues with how annotations are stored in cobrapy.

[1] MIRIAM qualifiers are dropped
RDF annotations are triples of the form (subject - predicate - object), with the object being the database which the subject (element in cobrapy) is annotated to. An crucial part of the annotation is the **predicate** which defines the relationship between the annotated element and the annotation.

http://co.mbine.org/standards/qualifiers

It makes a big difference (for all algorithms working with annotations) if something has a `is` relationship or `hasPart`, or `isHomologTo`.

So instead of storing things like
```
{
  'ImaginaryCompDB':'SpecificCompIdentifier',
  "uniprot": ["Q12345", "P12345"]
}
```

We should also store the predicates (otherwise a round trip will change the annotations).
```
{
  ''ImaginaryCompDB': ('is', 'SpecificCompIdentifier'),
  "uniprot": [('hasPart', "Q12345"), ('hasPart', "P12345")]
}
```

[2] Database ids should be identifiers.org collection identifiers
https://www.ebi.ac.uk/miriam/main/collections
http://identifiers.org/documentation
Basically most of the annotations reference a database collections from identifiers.org. The lookup key should be the collection namespace. This is the case for most databases, but is not the case for `SBO` currently which should be `sbo`.
This is a problem because not all users use the sboTerm for sbo annotations, but the alternative use of a RDF annotation to `sbo` collections is also valid. It should not matter if one wrote an SBO term or a SBO annotation, these should be treated the same.

Solution: change `SBO` to `sbo` key in annotation map.

This would affect code working with annotations.

Midnighter:
Sounds very reasonable to me. We want to make a push for a cobrapy 1.0 release after this summer. There will be the GSoC projects, you want to work on a better SBML parser, and we will also contribute in a more focused manner. So then it is okay to introduce breaking changes.

cdiener:
I also think that sounds reasonable. Might also think of a pandas DataFrame for annotations here with columns database, relation and id.

matthiaskoenig:
There needs to be a decision on how annotations should look as a data structure.

Important points for the annotations are:
* easily accessible, i.e. via `Object.annotation` field on first-class cobra objects, i.e., `Model`, `Gene`, `Metabolite`, `Reaction` and to come `Compartment` and `Group`
[...]

PR #1237, opened 2022-06-19
Metadata fbc3

akaviaLab:
Updated version of https://github.com/opencobra/cobrapy/pull/988. Tried to simplify and tidy the code in addition to merging. Modified according to @cdiener requests
- Ignored changes in notes for now
- renamed constructors a bit
- added eq comparisons that build classes if necessary

Original description  of #988 follows

Metadata Classes: A separate directory for handling metadata information is made inside cobra/core directory. Every object derived from SBase can have meta information like annotation (CVTerms), notes, history attached to it.
The CVTerms class for storing externally linked resources to each component derived from SBase. This class is maintaining the new format annotation as well as old format annotation simultaneously, and both are kept synchronized with each other. Changing one will modify the other accordingly. This new class for annotation can handle any type of annotation data (be it the case of nested annotation or alternative annotation). It can read the old format as well as the new format annotation data from JSON and other formats. At the time of writing back the model, the new data format is used because it contains the complete data organized in the same way as SBML.
The History class used for storing the history, validating dates, etc, is now attached to each component derived from SBase.
The KeyValuePair class for storing key-value pairs, defined by fbc-v3.
The last three metadata objects (i.e CVTerms, History, KeyValuePair) are present inside a single attribute of SBase (Object) class and can be accessed via object.annotation.cvterms, object.annotation.history and object.annotation.key_value_data attributes. Calling simply the annotation attribute (object.annotation) will return the annotation data in old format (making it backward compatible).

Group to JSON: The support of the group package is extended to JSON.

JSON schema v2: The version2 of JSON schema has been added which defines the new format annotation, history, key-value pair, notes, group package data, user-defined constraints data and basic SBML info.
@akaviaLab comment - not sure if the v2 schema works or is done. Seems unclear from the code.

Issues Fixed
fixes issue https://github.com/opencobra/cobrapy/issues/954: The support of group package has been extended to JSON and other formats.
fixes issue https://github.com/opencobra/cobrapy/issues/856: The infinity values are also enabled for storing bounds.
fixes issue https://github.com/opencobra/cobrapy/issues/810: The sbml_info storing basic information of SBML is written to JSON to store the basic SBML document information like packages, level, version, notes, annotation attached to the SBML component etc.
fixes issue https://github.com/opencobra/cobrapy/issues/736: If annotation is present in the form of list of list, it is first modified and then data is read.
fixes issue https://github.com/opencobra/cobrapy/issues/706: Single annotation resource are now put into a string. While reading from JSON also, if a single resource attribute is in the form of string, it is first fixed to list and then read into the model.
fixes issue https://github.com/opencobra/cobrapy/issues/684: The complete metadata structure (CVTerms, Notes, History) has been redesigned with backward compatibility. Old format data is read, fixed and then written in new format metadata structures. The fbc-v3 "KeyValuePair" class is also a part of this new format annotation. Its corresponding class and its support to JSON and other format has been added. However, the SBML parser for it has to be updated when libsbml adds support for fbc-v3.
fixes issue https://github.com/opencobra/cobrapy/issues/937: The annotation format has been updated, which is backward compatible too.
JSON schema v1: The JSON schema version 1 is modified to resolve some pre-existing issues.

Tests


akaviaLab:
@cdiener I think this is the reference
http://co.mbine.org/specifications/sbml.level-3.version-2.core.release-2.pdf
http://co.mbine.org/specifications/sbml.level-3.version-1.fbc.version-2.release-1.pdf

cdiener:
@akaviaLab What you linked is FBC version 2. The features implemented here look like a new version of FBC (version 3) but even after some searching I can not find a definition for this version (3).

akaviaLab:
@cdiener
Why do you think it has to be in a new version?
I think the title might be wrong (and it is sbml3, not fbc3), but the
annotation feature (with controlled vocabulary, cvterms) appears in this
link
http://co.mbine.org/specifications/sbml.level-3.version-2.core.release-2.pdf
page 100 and onward

The keyvaluepair could be useful for things that aren't in identifiers and
need to be annotated like
https://synonym.caltech.edu/documents/elaborations/miriam_annotation_syntax/#how-do-i-put-inchi-strings-in-annotations
(Ignore the fact that InChi is now in identifiers, and look at the format).
Right now, it is useful for noting it in COBRA, and we can probably modify
the code for import/export later.

What about the features in the code is not described by the standards I
linked?

cdiener:
Just based on previous discussions where it was mentioned that a new version of FBC would have key-value pairs as those are currently not supported/allowed in SBML. The CV terms (still no fan of the name) do indeed look like they are compatible with the current version of SBML so those could be merged immediately.

akaviaLab:
Cvterms are accesed via structured as requested. Did I misunderstand you?

cdiener:
No, no all good 👍

cdiener:
[This document](https://sourceforge.net/p/sbml/code/26312/tree/trunk/specifications/sbml-level-3/version-1/fbc/spec/main.pdf?format=raw) mentions FBC v3 (section 4) and includes the key-value pairs. However, that seems to be a proposal and not a final consensus. Also, there seems to have been no updates since 2020...

akaviaLab:
Hmm
We can remove the keypairvalues while we discuss the format. Or just remove
the code.

I have no strong feelings about this.

cdiener:
I actually would leave it in. It is very convenient, and we could wait with writing it to SBML until FBC3 is adopted. I think this PR is in a pretty good state. I'll try my best to get to a review.

akaviaLab:
@matthiaskoenig can you help review this?

.gitignore
```diff
@@ -114,14 +116,15 @@ ipython_config.py
...
 # auto-import.
-# .idea/artifacts
-# .idea/compiler.xml
-# .idea/jarRepositories.xml
-# .idea/modules.xml
-# .idea/*.iml
-# .idea/modules
-# *.iml
-# *.ipr
+ .idea/artifacts
```
cdiener: All those `.idea` files need to be removed. Why not just add `.idea/**` to the gitignore?

src/cobra/core/__init__.py
```diff
@@ -2,6 +2,7 @@
 from cobra.core.dictlist import DictList
 from cobra.core.gene import Gene, GPR
 from cobra.core.metabolite import Metabolite
+from cobra.core.metadata import *
```
cdiener: Star import are not good practice. Import what you need explicitly. This will also import any private constants etc into the global namespace.

src/cobra/core/metadata/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Provide functions for loading, saving and modifying metadata annotations."""
+
+from cobra.core.metadata.cvterm import (
```
cdiener: Might be a bit overzealous, but I'm a sucker for good class and variable names and would personally prefer if all the CVterm names and classes get better names like `StandardizedAnnotation[s]` etc.

src/cobra/core/metadata/cvterm.py
```diff
@@ -0,0 +1,652 @@
...
+    bqm_isDerivedFrom = 16
+    bqm_isInstanceOf = 17
+    bqm_hasInstance = 18
+    bqm_unknown = 19
+
+
+class CVTerm:
+    """CVTerm class, representing controlled vocabulary.
+
+    Representation of one CVTerm of an object with theirdependency structure.
```
cdiener: There are some typos here and that docstring is not really helpful unless you are already knees-deep into SBML and FBC syntax. I would try to make this more accessible and also link to the design documents so future devs will understand where that structure comes from.

src/cobra/core/metadata/cvterm.py
```diff
@@ -0,0 +1,652 @@
...
+        """Return the GPR with module, class, and code to recreate it."""
+        return (
+            f"{self.__class__.__module__}.{self.__class__.__qualname__}"
+            f"({self.to_ordered_dict()})"
+        )
+
+
+class CVTermList(UserList):
+    """
+    Representation of all CVTermList of an object in their
```
cdiener: Again, this explanation has to become more accessible. COBRAPY will become the layer between the normal user and the abstract SBML annotation designs so it should be clear what terms mean. When people work with annotations they usually think databass and IDs so I would have no idea what is meant by `external resource` or `qualifier`.

src/cobra/core/metadata/cvterm.py
```diff
@@ -0,0 +1,652 @@
...
+        for datum in self.data:
+            if (
+                datum.external_resources.resources
+                or datum.external_resources.nested_data
+            ):
+                resources.update(datum.external_resources.resource_set)
+        return frozenset(resources)
+
+    @property
+    def qualifiers(self) -> FrozenSet:
```
cdiener: Maybe consolidate this with all the other `query` functions in cobrapy and just accept a function as argument. Also, I think it just needs a single `CVTermList.query`. The passed function can then either match on the qualifier or resource.

src/cobra/core/metadata/cvterm.py
```diff
@@ -0,0 +1,652 @@
...
+            return False
+        if len(self.data) != len(other.data):
+            return False
+        for self_i, other_i in zip(self.data, other.data):
+            if self_i != other_i:
+                return False
+        return True
+
+
+class ExternalResources:
```
cdiener, line 877: If the only permissible data type is a list of URLs why is not called `URLs` or `URIs`?

src/cobra/core/metadata/history.py
```diff
@@ -0,0 +1,332 @@
...
+
+STRTIME_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
+
+
+class History:
+    """History object encoding object provenance.
+
+    Parameters
+    ----------
+    creators : list
```
cdiener, line 20: Wouldn't authors be better. Where do you track individuals that did not create the model but modified it?

src/cobra/core/metadata/history.py
```diff
@@ -0,0 +1,332 @@
...
+        return f"<{self.__class__.__name__} {self.creators}" \
+               f" {self.created_date.datetime}>"
+
+
+class Creator:
+    """Metadata for person who created an object.
+
+    Parameters
+    ----------
+        given_name : str,
```
cdiener: As above, why not "Author". 

And a criticism of the standard itself: This name scheme is centered on western populations and ignores cultural differences in how a person's name is formed. See https://uxmovement.com/forms/why-your-form-only-needs-one-name-field/.

src/cobra/core/metadata/history.py
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+    def __str__(self) -> str:
+        return str(self.to_dict())
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__} {self.given_name} {self.family_name}>"
+
+
+class HistoryDatetime:
```
cdiener: Couldn't this just be a normal datetime object?

src/cobra/core/metadata/keyvaluepairs.py
```diff
@@ -0,0 +1,109 @@
+from collections import UserDict
+from typing import Dict, Iterable, Union
+
+from ...util import format_long_string
+
+
+class KeyValueEntry:
```
cdiener: I would prefer a better name like `CustomAnnotation`.

src/cobra/core/metadata/keyvaluepairs.py
```diff
@@ -0,0 +1,109 @@
...
+    """Single key-value entry.
```
cdiener: This documentation is insufficient. It should at least cover the use case and an example.

src/cobra/core/metadata/keyvaluepairs.py
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+    def __str__(self) -> str:
+        return str(self.to_dict())
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__} ({self.key}, {self.value}, {self.uri})>"
+
+
+class KeyValuePairs(UserDict):
```
cdiener: Insufficient documentation.

src/cobra/core/metadata/metadata.py
```diff
@@ -0,0 +1,218 @@
...
+
+from ..metadata.cvterm import CVTerm, CVTermList
+from ..metadata.history import Creator, History
+from ..metadata.keyvaluepairs import KeyValuePairs
+
+
+class MetaData(MutableMapping):
+    """Meta-data of a cobrapy object.
+
+    Meta-data encodes additional information on an object such as annotations
```
cdiener, line 16: Some more discussion of why that is designed that way and links to FBC design documents would be helpful.

src/cobra/core/metadata/notes.py
```diff
@@ -0,0 +1,206 @@
+import logging
+import re
+from collections.abc import MutableMapping
+from typing import Dict, Iterator
+
+
+logger = logging.getLogger(__name__)
+
+
+class Notes(MutableMapping):
```
cdiener: Do notes have to be XHTML? I thought any string is allowed. And as discussed, I think notes should just be a string with maybe one helper function to interpret the legacy notes for backwards compatibility.

src/cobra/data/mini.json
```diff
@@ -5,1386 +5,1588 @@
     },
     "genes": [
         {
-            "id": "b0755",
+            "id": "G_b0755",
```
cdiener, line 8: Why did the IDs get changed suddenly?

src/cobra/data/mini.yml
```diff
@@ -1,1167 +1,1233 @@
 !!omap
 - metabolites:
   - !!omap
-    - id: 13dpg_c
+    - id: M_13dpg_c
```
cdiener, line 4: Why did the IDs change?

src/cobra/data/mini.json
```diff
@@ -5,1386 +5,1588 @@
     },
```
cdiener, line 5: We should make sure that we have test examples for both the v1 and v2 JSON schema.

src/cobra/io/mat.py
```diff
@@ -242,13 +246,41 @@ def _cell_to_float_list(
...
+        Replace infinite values with defined inf. Default None (don't replace), will be
+        replaced by given value  if given.
 
     Returns
     -------
     List
         A list of processed floats.
     """
-    return [float(x[0]) if np.size(x[0]) else empty_value for x in m_cell]
+
```
cdiener: Aren't those changes in the other PR?

src/cobra/core/metadata/cvterm.py
```diff
@@ -0,0 +1,652 @@
...
+from .helper import URL_IDENTIFIERS_PATTERN, parse_identifiers_uri
+
+
+logger = logging.getLogger(__name__)
+
+
+class Qualifier(Enum):
+    """The possible qualifiers inside a CVTerm."""
+
+    bqb_is = 0
```
cdiener: Drop the `bqb_` prefix.

src/cobra/io/sbml.py
```diff
@@ -1767,43 +1768,83 @@ def _parse_annotations(sbase: libsbml.SBase) -> dict:
...
     # RDF annotation
     cvterms = sbase.getCVTerms()
     if cvterms is None:
         return annotation
 
-    cvterm: "libsbml.CVTerm"
-    for cvterm in cvterms:
-        for k in range(cvterm.getNumResources()):
-            # FIXME: read and store the qualifier
+    def _cvterm_to_cobra(_cvterm: "libsbml.CVTerm") -> CVTerm:
```
cdiener: Not sure about it, but this could also be a classmethod on the `CVTerm` class. Right now this here is a `CVTerm` constructor that is not kept together with the class.

src/cobra/io/schema_v1.json
```diff
@@ -153,5 +153,5 @@
     "metabolites",
     "genes"
   ],
-  "additionalProperties": false
+  "additionalProperties": true
```
cdiener, line 156: What does this do?

src/cobra/io/schema_v2.json
```diff
@@ -0,0 +1,370 @@
...
+          }
+        }
+      }
+    }
+  },
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "pattern": "^[a-zA-Z|_][a-zA-Z0-9_]*$"
```
cdiener, line 114: Hmmm, is that taken from SBML? Not sure if I like it because that is incompatible with the COBRAPY ID format.

documentation_builder/metadata.ipynb
```diff
@@ -0,0 +1,548 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "bVgYSbcPLbwU",
+    "colab_type": "text"
+   },
+   "source": [
+    "# Metadata\n",
+    "\n",
+    "Metadata in the form of annotations, provenance or additional key:value pairs is an important asset for a model.\n",
```
cdiener, line 37: This notebook sounds it is written for people designing standards. Will a newcomer to metabolic modeling know what provenance means? This should become more accessible.

documentation_builder/metadata.ipynb
```diff
@@ -0,0 +1,548 @@
...
+   "metadata": {
+    "collapsed": false
+   }
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "history = History(\n",
```
cdiener, line 89: This looks very clunky with all the constructors, type casting. This would be much more user-friendly with a simpler API.

cdiener (commented):
Okay. I like the general design. There are some discussion points from my side, though some of those may be more about the standard itself, so they may be out of scope. 

Another thing that may be helpful to add in a later PR is a method on Metadata that flattens annotations into a pandas.DataFrame. This may drop some information but will be useful when merging with results or providing supplementary data for a paper. The graph-based formats here are hard to manage when you want to look at annotations for several objects at once.

src/cobra/core/metadata/cvterm.py
akaviaLab: You did notice that some of them are bqm_ prefix, right?
So you think there should be 
is = 0
...
bqm_is = 14

src/cobra/core/metadata/cvterm.py
akaviaLab: That's how it is defined in the standard. Apparently, they can also be URNs, which are not a valid web address.
I can change the documentation, or I can rename it URL or something like that, and people who care about URNs will be deep in the standard anyway.

src/cobra/core/metadata/history.py
akaviaLab: Creator is defined in the standard.
The standard defines given name, and family name. I generally agree with this, but if there was one name field, it would not be exported/imported well with libsbml. If libsbml will be rewritten, I don't mind changing it.

src/cobra/core/metadata/history.py
akaviaLab: Probably. I'll re-look at this code, but this is pretty much as it was in #988.

src/cobra/core/metadata/notes.py
akaviaLab: Right. I should just delete this file. It is no longer valid

src/cobra/io/schema_v2.json
akaviaLab: I think it was. It might be worth having a discussion on JSON standards. 
The IDs for JSON and YAML changed because this patch matches the IDs to SBML.
I can
1) Remove all ID changes
2) Keep them as an option, but have default be no changes
3) Keep them as default.

@matthiaskoenig - you seem to prefer no changes in SBML if I recall correctly. Do you have an opinion on replacement in JSON?

src/cobra/io/mat.py
akaviaLab: Yes, but I needed them for this PR. When that PR is merged, I'll merge devel back into this PR, which would probably remove this change.

documentation_builder/metadata.ipynb
akaviaLab: I'm thinking NamedTuple, which will unfortunately be immutable. Python 3.7 has dataclass, which might be even better, but would force cobrapy to use Python 3.7. What do you think? NamedTuple or dataclass?

src/cobra/core/metadata/history.py
akaviaLab: I was wrong. Creator has the option of being with "NAME" or with both "GIVEN NAME" and "FAMILY NAME"
The problem is that some models have both family and given. I could set cobra to read them both into name, and that the cobra class only uses name. Would that be better?

src/cobra/io/sbml.py
akaviaLab: I don't like that, since then cvterm.py has to import libsbml, which is irrelevant to the functioning of cvterm.

src/cobra/core/metadata/history.py
akaviaLab: Replaced with datetime object.

cdiener:
No need to try fixing 3.6 issues. We should probably drop support for it anyway. https://endoflife.date/python

akaviaLab:
Okay. I'll keep that in mind for next issues. Is this what you meant with the creator for a simpler API?

cdiener:
Yep, pretty much. I would probably allow the dates to be provided directly as string and have __init__ convert it to datetime. And maybe have some things initialized to default dates. For instance, the creation date could just be the current time by default. Might also be helpful if the creator info can be provided in the configuration and is then used by default.

src/cobra/core/metadata/cvterm.py
akaviaLab: ping @cdiener

akaviaLab:
> Yep, pretty much. I would probably allow the dates to be provided directly as string and have **init** convert it to datetime. And maybe have some things initialized to default dates. For instance, the creation date could just be the current time by default. Might also be helpful if the creator info can be provided in the configuration and is then used by default.



> Yep, pretty much. I would probably allow the dates to be provided directly as string and have **init** convert it to datetime. And maybe have some things initialized to default dates. For instance, the creation date could just be the current time by default. Might also be helpful if the creator info can be provided in the configuration and is then used by default.

History accepts str or datetime as a format for the dates.
parse_datetime parses string, and deals with datetime appropriately
So that seems to do what you described already, unless I misunderstood

Creation time now() by default - I can do that.
Creator in config - reasonable, but next PR.

matthiaskoenig:
Hi all,
I could review/update the code.

For clarification with the key/value store. This is basically the new fbc-version3 package which just needs a second implementation (which is this implementation) to be officially accepted. I.e. this is basically official as soon as we merge this in develop. So the key-value pair has to stay in.

A lot of discussion here. Are there any open things which have to be decided or is everything down to implementation details at this point?
Best Matthias

akaviaLab:
Hi Matthias,
As far as I remember, there are still some open discussions
1) Should we rename CVTerm to something like standardized?
2) Should creators have only one name, or stay with family name and given
name? Should one name be an option?
https://github.com/opencobra/cobrapy/pull/1237#discussion_r912185946
3) Renaming the qualifiers, see
https://github.com/opencobra/cobrapy/pull/1237#discussion_r912191819. I
think they should stay as is.
4) What should the JSON schema include? See
https://github.com/opencobra/cobrapy/pull/1237#discussion_r912193394 and
https://github.com/opencobra/cobrapy/discussions/1240#discussioncomment-3088901

The rest is just tidying, code implementation. I'm still working on it, but
if you have comments now that would be helpful.

Thank you,

Uri David

src/cobra/io/schema_v2.json
matthiaskoenig: I would prefer no id changes at all. Id changes just create so many issues. I highly would recommend to get rid of all id changes besides:
- JSON import (IDs are fixed to be valid SIds)
- MAT import (IDS are fixed to be valid SIds)

It would be great if we could have a vote/decision on this because most of our issues are related to the nonsense id replacements. There are hundreds of lines of code to deal with this nonsense which create more and more issues and still don't catch all corner cases.

akaviaLab: What do you mean by SIds?
SBML IDs?
If JSON IDs are fixed to be valid SBML IDs, there must be an id change, because JSON allows IDs that are not SBML compliant. Also, cobrapy allows IDs that are not SBML compliant. If there is a change when reading JSON files, then there are usually changes when writing JSON files, and we're back at the "nonsense id replacement."

We could decide that there are no ID changes when reading JSON/MAT/YAML (as long as the ID is valid for cobra), and then make each format responsible for converting IDs to its own format.
So JSON can have an id that looks like "3dpg", which is not valid SBML ID, and if I read that model into cobra, and then write it as SBML, it will be transformed to meta id "M_3dpg" when writing SBML.

So each format does
File -> Cobrapy format when reading
cobrapy format -> file limitations when writing

akaviaLab: If we have absolutely no id changes, what will happen when there is an ID (in JSON) that is not an SBML valid id?

matthiaskoenig: Yes, SIds are valid SBML ids. This can be checked very easily with a regex.

Like mentioned above the only id changes are on
- JSON import (IDs are fixed to be valid SIds)
- MAT import (IDS are fixed to be valid SIds)
- OTHER_FORMAT_NOT_SBML import (IDS are fixed to be valid SIds)

There would be a single function fixing ids, e.g. `fix_invalid_ids` used in both/all importers to have a consistent behavior across formats. Nothing would be replaced fixed on any exports, because SIds are valid JSON and MAT Ids.

documentation_builder/metadata.ipynb
matthiaskoenig: I agree. This was more for testing/low level documentation.

matthiaskoenig:
> Yep, pretty much. I would probably allow the dates to be provided directly as string and have init convert it to datetime. And maybe have some things initialized to default dates. For instance, the creation date could just be the current time by default. Might also be helpful if the creator info can be provided in the configuration and is then used by default.

Please don't set any current times! This will create diffs on your SBML models and model files every time you run your scripts resulting in large git diffs! I.e. don't set a current time as default. Perhaps an option for that, but please not the default behavior. The HistoryElement can be set on all core objects, so having diffs on all objects every time a script is executed is a bad idea.

akaviaLab:
Right now, the document (libsbml.SBMLDocument),  is set to be created at
current time if there is no document creation time.
Nothing else is created, because having diffs is a bad idea. No entities,
and not the libsbml.Model entity either.

I can remove this, but this doesn't seem harmful to me.

akaviaLab:
Okay. I think if this is what the community wants to do, we should define a
standard of
- What is a valid cobray ID?
- How to deal with weird characters like :, that are currently converted to
ASCII code. We can decide to have something like SBML_COLON, SBML_SEMICOLON
(it might exist, I'm not sure)

And get it done **before** cobrapy 1.0

@cdiener @Midnighter @synchon - any comments?

cdiener:
@matthiaskoenig 

Key-value pairs will be in there. They are super useful and a regularly requested feature.

The current date was only supposed to use current time as default value for the constructor. So it would only be set the first time the history is created and only if the user does not specify a date. EDIT: see reply below. Only set on writin gif no date is set.

The other discussion was whether we should limit the format of IDs for the v2 JSON schema. I don't think there is a good solution here, since the SBML and COBRAPY ID requirements are not compatible. So it will come down to whether we want the JSON schema 100% SBML compliant, with the limitation that not all COBRAPY models can be saved to that format.

The rest is just naming of classes etc. to make it a bit more accessible to devs and users without dep knowledge of SBML abstractions.

akaviaLab:
> The current date was only supposed to use current time as default value for the constructor. So it would only be set the first time the history is created and only if the user does not specify a date. So only when explicitly doing `model.annotation.history = History(...)` the first time and when not providing the date explicitly in the constructor. We could skip this, but I suspect many users will do something like `model.annotation.history = History(created_date=datetime.now(), ...)`. Right now, it will initialize to None if not specified. If you prefer that, we can leave it as is...

Right now, History() is called as part of every annotation, so modifying any default there would lead to *every* entity having a created date. That's a bad idea.
sbml.py has a small section when writing a document - if that document (not the model, but the document) has no creation date, it will be set to current time.

`        elif isinstance(sbase, libsbml.SBMLDocument):
            time = datetime.datetime.now()
            timestr = time.strftime("%Y-%m-%dT%H:%M:%S%z")
            date = libsbml.Date(timestr)
            _check(comp_history.setCreatedDate(date), "set creation date for document")
`

That's can be removed or stay.

matthiaskoenig:
I would highly prefer to have SBML SIds as valid cobrapy ids. 
- These can be serialized to all formats
- These work in optlang and the solver interfaces
- These are valid object identifiers (which is sometimes important)
- These are very simple to validate with the following regex `[a-zA-Z][a-zA-Z0-9_]*`
- Most models already use valid SIds because SBML is the de facto exchange standard

The only exceptions are some databases such as BiGG (no longer maintained, which also provides SBML with valid IDs).

These would allow:
- to get rid of all code for id replacements besides on the importers for JSON/Matlab
- to have no id replacements on SBML and valid cobrapy models (these create a ton of issues when mapping data/models, ...) and create a ton of bugs. 

I strongly advocate for this solution and just enforce valid SIds and move id_replacement in helper functions which would be applied for import besides SBML (and as toolbox for people to fix there invalid models).

**edit**: fixed regex

akaviaLab:
@cdiener @matthiaskoenig 
Is the documentation in af5f3fa for CVTerm and Qualifier better? What the documentation should be in general?

matthiaskoenig:
> @cdiener @matthiaskoenig Is the documentation in [af5f3fa](https://github.com/opencobra/cobrapy/commit/af5f3fa8974a1b514a22d76829c45adcbbfac4a5) for CVTerm and Qualifier better? What the documentation should be in general?

Yes, the documentation is better. But please use a string enum for such enums.
This will allow to parse strings much easier to the actual instances. I.e. something along the lines of
```
class MyEnum(str, Enum):
    state1 = 'state1'
    state2 = 'state2'
```

src/cobra/io/schema_v2.json
cdiener: This would still not affect models build directly with cobrapy objects. And if we substitute IDs there on object creation it would be hard to get backwards compatibility for older code, especially with the JSON v1 schema which is fixed/won't be changed anymore.

However, I see your point for JSON v2 (this discussion) and I think it makes sense to have it as a "fully SBML-compliant without the XML" format. We can maybe just provide a helper function to fix IDs and direct users to that if they want to write it and the IDs are not compliant.

cdiener:
> I would highly prefer to have SBML SIds as valid cobrapy ids.
> 
>     * These can be serialized to all formats
> 
>     * These work in optlang and the solver interfaces
> 
>     * These are valid object identifiers (which is sometimes important)
> 
>     * These are very simple to validate with the following regex `[a-zA-Z][a-zA-Z0-9_]*`
> 
>     * Most models already use valid SIds because SBML is the de facto exchange standard

I think this is hard because of backwards compatibility as mentioned above, and it would be a bit inconsistent for instance when reading models with JSON v1 schema. But I so see your point and agree that all the SBML-like formats IDs should be read as is and there should be helpers to convert them to that format. Point 2 is not fully correct since there are some IDs that are SBML-compliant but don't work in optlang in some solver interfaces (IDs longer 255 chars in GLPK, or metabolites named "st" in Gurobi for instance). So basically we push the defaults toward SBML-compliant IDs without breaking other formats such as Matlab etc.

pascalaldo:
Hi everyone, what are the current plans with this pull request? It contains a great set of features. We're currently looking at updating the BiGG repository and having improved annotation + group support in cobrapy would benefit us greatly.
I get that the changes to the representation of annotations done here could break quite some existing code. But if there are more hurdles, I can allocate time to update and improve the code here.

Midnighter:
That would be fantastic! As I understand it, @akaviaLab has moved on to other topics and will not contribute further.

pascalaldo:
Great, I'll start by merging devel branch changes into here and get all the code working and tests running again.

PR #1246, opened 2022-07-08
Serialize GPRs to string to avoid size explosion

cdiener:
* [x] fixes size explosion when serializing Reaction objects
* [X] tests added/passed

This is a fix for a weird bug. I noticed that the pickled MICOM models suddenly were 10-100x larger than before. This is due to the new GPR class that derives from `ast.Module`. When serializing those objects, genes are saved as `ast.Name` which saves an additional context (it's a bit unclear to me what is stored in there). This context can take up a massive amount of space. For instance for a MICOM model:

```
In [112]: sys.getsizeof(pickle.dumps(com.reactions[8].gpr.body.ctx))
Out[112]: 325105

In [114]: sys.getsizeof(pickle.dumps(com.reactions[8].gpr.to_string()))
Out[114]: 66
```

The context takes up 325KB but the string version can be saved in 66 bytes. This alone was enough to increase the size of MICOM models by 10-100x. This PR overrides the serialization of `cobra.Reaction` to serialize GPRs into strings. That creates a slight overhead when reading models from serialized formats, but limits the storage size. I could not overwrite `GPR.__getstate__` since that derives from `ast.Module` which does not allow this.


cdiener:
Not sure if we want backwards compatibility in loading pickled versions. I could also rebuild the pickle files...

akaviaLab:
Seems to make sense to me.

src/cobra/core/reaction.py
```diff
@@ -882,6 +882,12 @@ def delete(self, remove_orphans: bool = False) -> None:
         )
         self.remove_from_model(remove_orphans=remove_orphans)
 
+    def __getstate__(self) -> dict:
```
synchon: `dict` -> `Dict` for Python <= 3.9. A "Returns" section in the docstring would be perfect.

synchon (changes requested):

src/cobra/core/reaction.py
cdiener: Done

synchon (approved):
Thanks! 😄

PR #1264, opened 2022-08-10
Fix 673

### LINKED ISSUE #673: Adding reactions from another model does not reset the reactions their metabolites refer to
From the mailing list:

> Thank you! That is very helpful. If I add a reaction, the metabolites are also added (good!) but all reactions from the universal model associated with those metabolites are included in the metabolite description, even though they don't exist in the model. Is this the expected behavior?

> For example, here I've only added EX_fru_e, none of the other reactions are in 'model'.

```python
In [310]: model.add_reactions([universe.reactions.EX_fru_e])
In [311]: model.metabolites.fru_e.reactions
Out[311]:
frozenset({<Reaction FRUt2r at 0x7fa918d40630>,
           <Reaction KESTOTTRASEe at 0x7fa90f107048>,
           <Reaction KESTOPTASEe at 0x7fa90f107320>,
           <Reaction LEVANASE_2e at 0x7fa90f124550>,
           <Reaction INULINASEe at 0x7fa90f161d68>,
           <Reaction OLIGOFRUASEe at 0x7fa90f09c550>,
           <Reaction LEVANASE_3e at 0x7fa90f124588>,
           <Reaction EX_fru_e at 0x7fa911398390>,
           <Reaction LEVANASE_4e at 0x7fa90f1245c0>,
           <Reaction KESTOASEe at 0x7fa90f1071d0>})
```

akaviaLab:
* [X] fix #673
Fix #673 by making reactions in species search the model fields, if this species is in a model.
Added checks in the code to copy the reaction if it is in a model that isn't the one it is added to.
Added properties reaction.metabolites as a settable property, added reaction_add, reaction_remove, reaction_clear to species.py, in order to minimize use of ._reaction.
* [X] tests added/passed
test_add_reactions_with_context
test_add_reactions_from_another_model

### DIFF
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -899,7 +899,7 @@ class Model(Object):
             """
             if new_group.id in self.groups:
                 logger.warning(
-                    f"Ignoring group '{new_group.id}'" f" since it already exists."
+                    f"Ignoring group '{new_group.id}' since it already exists."
                 )
                 return False
             return True
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -500,6 +500,9 @@ class Model(Object):
 
         # First check whether the metabolites exist in the model
         metabolite_list = [x for x in metabolite_list if x.id not in self.metabolites]
+        metabolite_list = [
+            x.copy() if x.model and x.model != self else x for x in metabolite_list
+        ]
 
         bad_ids = [
             m for m in metabolite_list if not isinstance(m.id, str) or len(m.id) < 1
@@ -718,8 +721,39 @@ class Model(Object):
                 return False
             return True
 
-        # First check whether the reactions exist in the model.
-        pruned = DictList(filter(existing_filter, reaction_list))
+        # First check whether the reactions exist in the model. Copy reactions that
+        # belong to another model, in order not to remove them (see
+        # https://github.com/opencobra/cobrapy/issues/673#issuecomment-371361476).
+        exist_but_different_stoich = set()
+        other_model = ""
+        for reaction in filter(lambda x: not existing_filter(x), reaction_list):
+            met_id_dict = {
+                met.id: stoich for met, stoich in reaction.metabolites.items()
+            }
+            existing_met_id_dict = {
+                met.id: stoich
+                for met, stoich in self.reactions.get_by_id(
+                    reaction.id
+                ).metabolites.items()
+            }
+            if met_id_dict != existing_met_id_dict:
+                exist_but_different_stoich.add(reaction.id)
+                other_model = reaction.model
+        if len(exist_but_different_stoich):
+            logger.warning(
+                f"The reactions {', '.join(exist_but_different_stoich)} exist"
+                f" in both {self} and {other_model}, but have different "
+                f"stoichiometries in the two mdoels. Please check to see "
+                f"which stoichiometry you'd like to add."
+            )
+        pruned = DictList(
+            [
+                reaction
+                if not reaction._model or reaction._model == self
+                else reaction.copy()
+                for reaction in filter(existing_filter, reaction_list)
+            ]
+        )
 
         context = get_context(self)
 
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -847,13 +847,13 @@ class Model(Object):
                 reaction._model = None
 
                 for met in reaction.metabolites:
-                    if reaction in met.reactions:
+                    if reaction in met._reaction:
                         met.reaction_remove(reaction, context)
                         if remove_orphans and len(met.reactions) == 0:
                             self.remove_metabolites(met)
 
                 for gene in reaction.genes:
-                    if reaction in gene.reactions:
+                    if reaction in gene._reaction:
                         gene.reaction_remove(reaction, context)
 
                         if remove_orphans and len(gene.reactions) == 0:
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -723,34 +723,24 @@ class Model(Object):
             return True
 
         # First check whether the reactions exist in the model. Copy reactions that
-        # belong to another model, in order not to remove them (see
-        # https://github.com/opencobra/cobrapy/issues/673#issuecomment-371361476).
-        exist_but_different_stoich = set()
+        # belong to another model, in order not to remove them from the original model.
+        exist_but_different = set()
         other_model = ""
         for reaction in filter(lambda x: not existing_filter(x), reaction_list):
-            met_id_dict = {
-                met.id: stoich for met, stoich in reaction.metabolites.items()
-            }
-            existing_met_id_dict = {
-                met.id: stoich
-                for met, stoich in self.reactions.get_by_id(
-                    reaction.id
-                ).metabolites.items()
-            }
-            if met_id_dict != existing_met_id_dict:
-                exist_but_different_stoich.add(reaction.id)
-                other_model = reaction.model
-        if len(exist_but_different_stoich):
+            if reaction != self.reactions.get_by_id(reaction.id):
+                exist_but_different.add(reaction.id)
+                other_model = other_model or reaction.model
+        if len(exist_but_different):
             logger.warning(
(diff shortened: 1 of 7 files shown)


src/cobra/core/model.py
```diff
@@ -734,8 +738,39 @@ def existing_filter(rxn: Reaction) -> bool:
...
 
-        # First check whether the reactions exist in the model.
-        pruned = DictList(filter(existing_filter, reaction_list))
+        # First check whether the reactions exist in the model. Copy reactions that
+        # belong to another model, in order not to remove them (see
+        # https://github.com/opencobra/cobrapy/issues/673#issuecomment-371361476).
+        exist_but_different_stoich = set()
+        other_model = ""
+        for reaction in filter(lambda x: not existing_filter(x), reaction_list):
+            met_id_dict = {
```
cdiener: I would refactor this into a utility function on the reaction class like `Reaction.__eq__(self, other)`. Then you can just check with `reaction == existing`. This should also check whether the bounds are the same or at least within `model.tolerance`.

src/cobra/core/model.py
```diff
@@ -744,23 +779,22 @@ def existing_filter(rxn: Reaction) -> bool:
...
-                    stoichiometry = reaction._metabolites.pop(metabolite)
-                    model_metabolite = self.metabolites.get_by_id(metabolite.id)
-                    reaction._metabolites[model_metabolite] = stoichiometry
-                    model_metabolite._reaction.add(reaction)
-                    if context:
-                        context(partial(model_metabolite._reaction.remove, reaction))
+            met_id_dict = {
+                met.id: stoich for met, stoich in reaction.metabolites.items()
+            }
+            met_dict = {met.id: met for met in reaction.metabolites.keys()}
```
cdiener: Just do `for met in reaction.metabolites`.

src/cobra/core/model.py
```diff
@@ -744,23 +779,22 @@ def existing_filter(rxn: Reaction) -> bool:
...
+            mets_to_add = [
```
cdiener, line 756: This seems complicated. Why not just do 

```python
mets_to_add = [
    met for met in reaction.metabolites 
    if met.id is not in self.metabolites.list_attr("id")
]
```

This way you don't need all the objects above.

src/cobra/core/model.py
```diff
@@ -828,21 +862,17 @@ def remove_reactions(
                 self.reactions.remove(reaction)
                 reaction._model = None
 
-                for met in reaction._metabolites:
+                for met in reaction.metabolites:
```
cdiener: This is slower though since it creates a copy of the metabolite list. It's not great to access hidden attributes bu here it creates a large memory overhead not to do that.

src/cobra/core/model.py
```diff
@@ -828,21 +862,17 @@ def remove_reactions(
...
-                        met._reaction.remove(reaction)
-                        if context:
-                            context(partial(met._reaction.add, reaction))
-                        if remove_orphans and len(met._reaction) == 0:
+                        met.reaction_remove(reaction, context)
+                        if remove_orphans and len(met.reactions) == 0:
                             self.remove_metabolites(met)
 
-                for gene in reaction._genes:
+                for gene in reaction.genes:
```
cdiener: same as above

src/cobra/core/reaction.py
```diff
@@ -579,6 +579,23 @@ def metabolites(self) -> Dict[Metabolite, float]:
...
+        """Set metabolites to a dictionary of metabolites and coefficients.
+
+        Parameters
+        ----------
+        value: Dict[Metabolite, float]
+            A dictionary of cobra.Metabolite for keys and floats for coeffecieints.
+            Positive coefficient means the reaction produces this metabolite, while
+            negative coefficient means the reaction consumes this metabolite.
+        """
+        self._metabolites = value
```
cdiener, line 595: See below. Context should be acquired automatically in `add_reaction`.

src/cobra/core/species.py
```diff
@@ -55,8 +57,52 @@ def reactions(self) -> FrozenSet:
         FrozenSet
             A frozenset that includes the reactions of the species.
         """
+        if self.model and self.__class__.__name__ == "Gene":
```
cdiener: This should be implemented in the seriving classes like `Gene` and `Metabolite`. This is what inheritance is for.

src/cobra/core/species.py
```diff
@@ -55,8 +57,52 @@ def reactions(self) -> FrozenSet:
...
+            return frozenset(self.model.reactions.query(lambda x: self in x, "genes"))
+        elif self.model and self.__class__.__name__ == "Metabolite":
+            return frozenset(
+                self.model.reactions.query(lambda x: self in x, "metabolites")
+            )
         return frozenset(self._reaction)
 
+    def reaction_add(
```
cdiener: Rename to `add_reaction` to be consistent with other cobrapy functions.

src/cobra/core/species.py
```diff
@@ -55,8 +57,52 @@ def reactions(self) -> FrozenSet:
...
+        ---------
+        reaction: cobra.Reaction
+        context: HistoryManager, optional
+            context this action is in, optional (defualt None).
+        """
+        self._reaction.add(reaction)
+        if context:
+            context(partial(self._reaction.remove, reaction))
+
+    def reaction_remove(
```
cdiener: Rename to `remove_reaction`.

src/cobra/core/species.py
```diff
@@ -55,8 +57,52 @@ def reactions(self) -> FrozenSet:
...
+        ---------
+        reaction: cobra.Reaction
+        context: HistoryManager, optional
+            context this action is in, optional (defualt None).
+        """
+        self._reaction.remove(reaction)
+        if context:
+            context(partial(self._reaction.add, reaction))
+
+    def reaction_clear(self) -> None:
```
cdiener: Rename to `clear_reactions` or `empty_reactions`.

src/cobra/core/species.py
```diff
@@ -55,8 +57,52 @@ def reactions(self) -> FrozenSet:
...
+        If this is called with a context, will be reversed when exiting the context.
+
+        Parmeters
+        ---------
+        reaction: cobra.Reaction
+        context: HistoryManager, optional
+            context this action is in, optional (defualt None).
+        """
+        self._reaction.remove(reaction)
+        if context:
```
cdiener: I would not provide the context as an argument. You can just get the context with `get_context(reaction._model)`.

src/cobra/core/species.py
```diff
@@ -55,8 +57,52 @@ def reactions(self) -> FrozenSet:
...
+        If this is called with a context, will be reversed when exiting the context.
+
+        Parmeters
+        ---------
+        reaction: cobra.Reaction
+        context: HistoryManager, optional
+            context this action is in, optional (defualt None).
+        """
+        self._reaction.add(reaction)
+        if context:
```
cdiener: I would not provide the context as an argument. You can just get the context with `get_context(reaction._model)`.

cdiener (changes requested):
Looks mostly good to me. 

The necessity to add so many sync functions just to ensure `Species._reactions` mirrors what is defined in `Model` makes it really clear that it would be much easier with a "single source of truth" design where `Species.reactions` only looks through reactions in the models for whether they contain the species and returns an error if the Species in not part of a Model (where that function does not make much sense since that is a 1:1 mapping anyway for reactions that are not part of a model).

src/cobra/core/model.py
akaviaLab: Should I change all other places where I access reaction.metabolites? I don't think so, but I'm not sure

akaviaLab:
Fixed almost everything. If I remove context from species/remove_reaction, species/add_reaction almost everything works, which is simpler.

However, test_remove_genes_with_context() breaks. Can you help me fix that? I'm not good with contexts. The previous version would not try to update contexts when calling add_reaction or remove_reaction from _associate_gene/__dissociate_gene and that would work.

akaviaLab:
@cdiener - can you help me fix test_remove_genes_with_context() please

PR #1272, opened 2022-09-04
refactor: enable flake8

synchon:
* [ ] fix #(issue number)
* [x] tests added/passed

This PR fixes remaining `flake8` checks and enables the check in `gh-actions` workflow.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -129,7 +129,7 @@ exclude =
     __init__.py
 max-line-length = 88
 # The following conflict with `black` which is the more pedantic.
-ignore =
+extend-ignore =
     B023
     E203
     W503
--- a/tox.ini
+++ b/tox.ini
@@ -130,6 +130,7 @@ exclude =
 max-line-length = 88
 # The following conflict with `black` which is the more pedantic.
 ignore =
+    B023
     E203
     W503
     D202
--- a/tox.ini
+++ b/tox.ini
@@ -130,7 +130,6 @@ exclude =
 max-line-length = 88
 # The following conflict with `black` which is the more pedantic.
 extend-ignore =
-    B023
     E203
     W503
     D202
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -39,7 +39,6 @@ jobs:
     - name: black
       run:
         tox -e black
-# We ignore flake8 for the time being because there are tons of things to fix.
-#    - name: flake8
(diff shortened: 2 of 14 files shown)


cdiener (approved):
This looks good to me. IMO, I think we can exclude the tests from flake8, though.

tox.ini
```diff
@@ -129,7 +129,8 @@ exclude =
     __init__.py
 max-line-length = 88
 # The following conflict with `black` which is the more pedantic.
-ignore =
+extend-ignore =
+    B023
```
Midnighter: Why ignore this bugbear error? It seems a relevant thing to look into?

Midnighter (changes requested):
Would like to get some clarity on `B023` before merging. Otherwise it looks good and such a great milestone for the project 🎉 Thank you already for all your hard work @synchon 💪🏼

synchon:
> This looks good to me. IMO, I think we can exclude the tests from flake8, though.

I think having the tests would keep the project tidy.

tox.ini
synchon: B023 has only one instance and it's from https://github.com/opencobra/cobrapy/blob/7b1a4ecaa5df90546f760bd100581f7cdf174d89/src/cobra/io/mat.py#L915 due to variable binding outside of scope in the lambda. I couldn't think of a better way to fix it while keeping the functionality. Maybe we'll find a better way soon but since it's not exactly hurting either way, I chose to exclude the check.

Midnighter: I think you can add a comment to the line:

```py
group_members = model.reactions.query(lambda x: x.subsystem == g_name)  # noqa: B023
```

rather than ignore it for the whole project.

Midnighter:
> I think having the tests would keep the project tidy.

I also agree with running flake8 on the tests.

tox.ini
synchon: Okay yes makes sense, will make the change.

cdiener:
Makes sense, thanks 🙏

tox.ini
synchon: @Midnighter The inline comment for ignoring doesn't work for B023, what do you suggest?

Midnighter: I pushed a fix (I think).

Midnighter (approved):

Midnighter:
🎉 party time 😃

PR #1273, opened 2022-09-07
updated installation instructions 

stellaprins:
Installation via conda-forge channel (instead of the bioconda channel) for latest version (0.25.0) of COBRApy


cdiener (approved):
Looks good. Thanks!

PR #1278, opened 2022-09-19
feat: add repr to dictlist

### LINKED ISSUE #1277: Neater notebook representation of Model Objects
### Checklist

It would be nice to have a neater representation of the model objects - metabolites, reactions, genes, etc. For example,
Screen Shot 2022-09-18 at 11 15 21 PM

This introduces a `_repr_html_` within the `DictList` object. Each `Object` then declares a `_DF_ATTRS` internally that contains the needed column list for the data frame representation to be created. Adding a `to_dataframe()` function to `DictList` would solve this.

```
>>> model.reactions.to_dataframe()
```

I've currently already had this implemented at `https://github.com/achillesrasquinha/cobrapy/tree/feats`

achillesrasquinha:
* [x] fix #1277

It would be nice to have a neater representation of the model objects - metabolites, reactions, genes, etc. For example,
Screen Shot 2022-09-18 at 11 15 21 PM

* [ ] tests added/passed


cdiener:
An alternative approach would be to have DictList inspect its own element type and map the printed columns based on that. Not great for encapsulation, but that does not break the pickling of models. Not sure what would be the best strategy.

achillesrasquinha:
I thought on similar lines as well is what attributes would go in best. But I think sticking to this style made me feel that reliance on instance specific df wouldn't be so concerning.

Midnighter:
I must admit, I'm not fully convinced by this approach. What if we added a new method to all the `DictList` classes here called `_repr_tr` which returns a representation of itself wrapped in `<tr></tr>` so that it can be included easily? Guess we'd also need a class method `_repr_th_` then for the column headers.

Of course, your approach has the benefit of easily transforming to a data frame which may have more general benefits.

src/cobra/core/dictlist.py
```diff
@@ -542,3 +543,28 @@ def __dir__(self) -> list:
...
+        if self:
+            item = self[0]
+            columns = [col for col in item._DF_ATTRS if col != "id"]
+
+        data = []
+        ids = []
+
+        for item in self:
+            ids.append(item.id)
+            data.append([getattr(item, attr) for attr in columns])
```
Midnighter, line 561:
```suggestion
from operator import attrgetter

        column_getter = attrgetter(*columns)
        for item in self:
            ids.append(item.id)
            data.append(column_getter(item))
```

src/cobra/core/dictlist.py
```diff
@@ -542,3 +543,28 @@ def __dir__(self) -> list:
         attributes.append("_dict")
         attributes.extend(self._dict.keys())
         return attributes
+
+    def to_df(self):
+        """Convert to a pandas dataframe."""
+        item = None
+        columns = []
+
+        if self:
```
Midnighter, line 552: Maybe it's clearer to say

```suggestion
        if not self.empty():
```

PR #1283, opened 2022-09-20
feat: add progress to fva

### LINKED ISSUE #1279: Showing progress during flux variability analysis.
### Checklist

- [ ] There are [no similar issues or pull requests](https://github.com/opencobra/cobrapy/issues) for this yet.

The currently implementation of FVA within cobrapy is generally slow (definitely improved using the `processes` flag. Would be nice to see the current progress since the program sometimes seems stalled for large networks. What say? I was planning on using `tqdm` within `variability.py` at line 247.

e.g.

Screen Shot 2022-09-19 at 2 57 36 PM

achillesrasquinha:
* [x] fix #1279 
* [x] description of feature/fix: Add a progress bar to FVA (triggered only during logging.INFO)
* [x] tests added/passed

### DIFF
--- a/src/cobra/flux_analysis/variability.py
+++ b/src/cobra/flux_analysis/variability.py
@@ -1,6 +1,5 @@
 """Provide variability based methods such as flux variability or gene essentiality."""
 
-
 import logging
 from typing import TYPE_CHECKING, List, Optional, Set, Tuple, Union
 from warnings import warn
@@ -8,6 +7,7 @@ from warnings import warn
 import numpy as np
 import pandas as pd
 from optlang.symbolics import Zero
+from rich.progress import Progress
 
 from ..core import Configuration, get_solution
 from ..util import ProcessPool
@@ -24,7 +24,28 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 configuration = Configuration()
+    
+class _FakeProgress:
+    def __enter__(self):
+        return self
 
+    def __exit__(self, *args):
+        pass
+
+    def add_task(self, *args, **kwargs):
+        return object()
+
+    def update(*args, **kwargs):
+        pass
+
+def _update_progress_bar(
+    progress_bar,
+    task,
+    advance: int,
+) -> None:
+    """Update progress bar.
+    """
+    progress_bar.update(task, advance=advance)
 
 def _init_worker(model: "Model", loopless: bool, sense: str) -> None:
     """Initialize a global model object for multiprocessing.
@@ -87,7 +108,6 @@ def _fva_step(reaction_id: str) -> Tuple[str, float]:
         {rxn.forward_variable: 0, rxn.reverse_variable: 0}
     )
     return reaction_id, value
-
 
 def flux_variability_analysis(
     model: "Model",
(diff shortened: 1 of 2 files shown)


src/cobra/flux_analysis/variability.py
```diff
@@ -26,6 +26,29 @@
...
 
+class _FakeProgress:
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        pass
+
+    def add_task(self, *args, **kwargs):
+        return object()
```
Midnighter, line 37: In rich, `add_task` returns an integer ID. So maybe just `return 0` here.

src/cobra/flux_analysis/variability.py
```diff
@@ -26,6 +26,29 @@
...
+
+    def update(*args, **kwargs):
+        pass
+
+
+def _update_progress_bar(
```
Midnighter, line 43: What is the point of this function? It just seems like an unnecessary level of indirection.

Midnighter (changes requested):
Hi @achillesrasquinha,

Thank you for the contribution. I think, having a dummy class is a reasonable approach but I do wonder if having two versions of the function altogether would be better? One with progress and the other without.

In any case, I would want to have a function parameter to enable or disable the progress bar altogether. Something like:

```py
def flux_variability_analysis(..., progress: Optional[bool] = None) -> pd.DataFrame:
    if progress is None and logger.getEffectiveLevel() <= logging.INFO:
        # Do progress bar
    elif progress:
        # Do progress bar
    else:
        # No progress bar
```

This would, of course, also be nice to have for the gene and reaction knock-outs. So maybe we can solve it a little more generally.

src/cobra/flux_analysis/variability.py
cdiener: Do you need this at all? In rich Progress.add_task already has a visible flag that hides the progress bar. Does that have too much overhead?

src/cobra/flux_analysis/variability.py
```diff
@@ -230,25 +253,43 @@ def flux_variability_analysis(
...
+                        initargs=(model, loopless, what[:3]),
+                    ) as pool:
+                        for i, (rxn_id, value) in enumerate(
+                            pool.imap_unordered(
+                                _fva_step, reaction_ids, chunksize=chunk_size
+                            )
+                        ):
+                            fva_result.at[rxn_id, what] = value
+                            _update_progress_bar(
+                                progress, progress_task, (i + 1) / num_reactions
```
cdiener, line 284: I haven't tested this but does this actually show the right numbers? Like the bar is set to a maximum of nr steps but then you do 2*nr updates. Once for the min and once for the max...

cdiener (commented):
Some questions. An example run would be nice to see if the numbers add up. Should also be added to the docs.

cdiener:
@Midnighter what about having the progress bar flag in the Configuration. Or having a verbosity flag there?

src/cobra/flux_analysis/variability.py
Midnighter: Yep, agreed. Didn't mention it now because it seems like we want a different design anyway. Also, use `enumerate(..., start=1)` instead of adding + 1.

Midnighter:
> @Midnighter what about having the progress bar flag in the Configuration. Or having a verbosity flag there?

I think, I would still create a specific `progress` parameter for those functions that show a progress bar but just like with `Configuration.tolerance` if the argument is unset that could be the default.

Overall, good idea to have it on the configuration. I would not use verbosity, though, because basically people should just create a logging handler with the level that they need.

src/cobra/flux_analysis/variability.py
Midnighter: Didn't realize there was that option. Sounds good 🙂

cdiener:
Okay then maybe continue using the logging level as a default and have the arg in that case.

PR #1290, opened 2022-10-18
Implemented a faster algorithm of flux variability analysis

earmingol:
Implemented a new algorithm for flux variability analysis that is faster than the regular algorithm. It is faster because is designed to solve less numbers of LPs.

A full description of this algorithm and benchmarking is available in [its pre-print paper](https://www.researchsquare.com/article/rs-2109496/v1)

* [x] tests added/passed

### DIFF
--- a/src/cobra/flux_analysis/variability.py
+++ b/src/cobra/flux_analysis/variability.py
@@ -402,3 +402,241 @@ def find_essential_reactions(
         deletions["growth"].isna() | (deletions["growth"] < threshold), :
     ].ids
     return {model.reactions.get_by_id(r) for ids in essential for r in ids}
+
+
+def faster_fva(
+    model: "Model",
+    reaction_list: Optional[List[Union["Reaction", str]]] = None,
+    loopless: bool = False,
+    fraction_of_optimum: float = 1.0,
+    pfba_factor: Optional[float] = None,
+    tol: Optional[float] = 1e-13,
+) -> pd.DataFrame:
+    """Determine the minimum and maximum flux value for each reaction.
+
+    This function uses a faster algorithm than flux_variability_analysis.
+
+    Parameters
+    ----------
+    model : cobra.Model
+        The model for which to run the analysis. It will *not* be modified.
+    reaction_list : list of cobra.Reaction or str, optional
+        The reactions for which to obtain min/max fluxes. If None will use
+        all reactions in the model (default None).
+    loopless : bool, optional
+        Whether to return only loopless solutions. This is significantly
+        slower. Please also refer to the notes (default False).
+    fraction_of_optimum : float, optional
+        Must be <= 1.0. Requires that the objective value is at least the
+        fraction times maximum objective value. A value of 0.85 for instance
+        means that the objective has to be at least at 85% percent of its
+        maximum (default 1.0).
+    pfba_factor : float, optional
+        Add an additional constraint to the model that requires the total sum
+        of absolute fluxes must not be larger than this value times the
+        smallest possible sum of absolute fluxes, i.e., by setting the value
+        to 1.1 the total sum of absolute fluxes must not be more than
+        10% larger than the pFBA solution. Since the pFBA solution is the
+        one that optimally minimizes the total flux sum, the `pfba_factor`
+        should, if set, be larger than one. Setting this value may lead to
+        more realistic predictions of the effective flux bounds
+        (default None).
+    tol : float, optional.
+        Tolerance to consider that a flux value is equal to the lower or
+        upper bounds (default 1e-13).
+
+    Returns
+    -------
+    pandas.DataFrame
+        A data frame with reaction identifiers as the index and two columns:
+        - maximum: indicating the highest possible flux
+        - minimum: indicating the lowest possible flux
+
+    Notes
+    -----
+    This implements the a faster algorithm described in [1]_. Please note that
+    the flux distribution containing all minimal/maximal fluxes does not have
+    to be a feasible solution for the model. Fluxes are minimized/maximized
+    individually and a single minimal flux might require all others to be
+    sub-optimal.
+
+    Using the loopless option will lead to a significant increase in
+    computation time (about a factor of 100 for large models). However, the
+    algorithm used here (see [2]_) is still more than 1000x faster than the
+    "naive" version using `add_loopless(model)`. Also note that if you have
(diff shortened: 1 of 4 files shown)


src/cobra/flux_analysis/variability.py
```diff
@@ -402,3 +402,241 @@ def find_essential_reactions(
...
+            lb_problem = lower_bound_problems.pop()
+            rxn_id = reaction_ids[lb_problem]
+
+            # Run FVA step
+            _init_worker(model, loopless, what[:3])
+            value = _fva_step(rxn_id)
+            fva_result.at[rxn_id, what] = value[1]
+            number_lps += 1
+
+            # Find reactions that hit bounds when optimizing objective
```
cdiener, line 601: Have you benchmarked this? I would expect this to be much slower than the previous implementation since it has to request all fluxes. The slowest step in those iterations is not solving the problem, but getting the primal values from the underlying solver interface,

src/cobra/flux_analysis/variability.py
```diff
@@ -402,3 +402,241 @@ def find_essential_reactions(
...
     return {model.reactions.get_by_id(r) for ids in essential for r in ids}
+
+
+def faster_fva(
+    model: "Model",
+    reaction_list: Optional[List[Union["Reaction", str]]] = None,
+    loopless: bool = False,
+    fraction_of_optimum: float = 1.0,
+    pfba_factor: Optional[float] = None,
+    tol: Optional[float] = 1e-13,
```
cdiener, line 413: This should be obtained from `model.tolerance`.

cdiener (commented):
Hmm, this runs much slower than the default implementation for me:

```
Python 3.10.7 (main, Oct  1 2022, 04:31:04) [GCC 12.2.0]
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from cobra.io import load_model

In [2]: from cobra.flux_analysis import flux_variability_analysis, faster_fva

In [3]: mod = load_model("textbook")

In [4]: %time flux_variability_analysis(mod, processes=1)
CPU times: user 59.9 ms, sys: 0 ns, total: 59.9 ms
Wall time: 59.6 ms
Out[4]: 
             minimum       maximum
ACALD  -1.078262e-13  0.000000e+00
ACALDt -7.361427e-14  0.000000e+00
ACKr   -9.290383e-14  0.000000e+00
ACONTa  6.007250e+00  6.007250e+00
ACONTb  6.007250e+00  6.007250e+00
...              ...           ...
TALA    1.496984e+00  1.496984e+00
THD2    0.000000e+00  5.134527e-12
TKT1    1.496984e+00  1.496984e+00
TKT2    1.181498e+00  1.181498e+00
TPI     7.477382e+00  7.477382e+00

[95 rows x 2 columns]

In [5]: %time faster_fva(mod)
CPU times: user 240 ms, sys: 1.56 ms, total: 242 ms
Wall time: 239 ms
Out[5]: 
             minimum       maximum
ACALD   4.315997e-14  0.000000e+00
ACALDt  9.786367e-14  0.000000e+00
ACKr    1.341944e-13  0.000000e+00
ACONTa  6.007250e+00  6.007250e+00
ACONTb  6.007250e+00  6.007250e+00
...              ...           ...
TALA    1.496984e+00  1.496984e+00
THD2    0.000000e+00 -3.095947e-12
TKT1    1.496984e+00  1.496984e+00
TKT2    1.181498e+00  1.181498e+00
TPI     7.477382e+00  7.477382e+00

[95 rows x 2 columns]


```

One thing at play here is that, in practice, the slowest part of solving a model with solution recycling is getting the primal values back from the solver. The default implementation only requires getting a single value from the solver object (the objective value) but the new implementation needs to get n_r primal values which makes each LP run up to a factor of n_r slower. So you run less LPs, but each optimization takes much longer.

earmingol:
Thanks for the comments. I wonder if there is a faster way to get the flux values from the solution? All of our benchmarkings were done by using guroby directly to solve the LPs so the access to the results was quicker. In this repo we have the original code for that https://github.com/DKenefake/fasterfva

I did try benchmarking with a couple of larger models, and the algorithm was faster for those.

Screen Shot 2022-10-18 at 16 41 44

cdiener:
Hmm I can't reproduce that even with larger models. I tried a bunch from your preprint but they all ran much faster on the old implementation. For instance iJO1366 took 5min with the code from this PR for me and 20secs on the old implementation. Getting the primal solution is already done with pure C code so I doubt you could speed it up faster. I recently added some optimizations for Gurobi to optlang but there is some intrinsic tradeoff when getting many primal values, especially by name. Can you give some more info how you run your benchmarks like OS and software versions?

earmingol:
Sorry, I meant if there is a faster way to gather the fluxes from the solver through COBRApy (for example accessing directly to a vector with all values, so I avoid the iterations). If you see the other repository, the benchmarking of the preprint was performed directly through Gurobi instead of COBRApy, allowing the access to the solutions as vectors. This better reflects the theoretical performance of the algorithm.

For the screenshot I put I am using a Macbook, with the Catalina MacOS. However, I just realized if I don't specify the `processes` parameters in the old implementation it automatically initializes all available workers –which looks like it takes a while– explaining why I get the opposite to what you get.

I think the slowest part of this COBRApy implementation of our algorithm is gathering fluxes from the Model class through iterations, so I'll try to improve that either by using more efficient ways to do that or implementing new functions to gather info directly from the solver. I would appreciate any suggestion on how doing that more efficiently.

Thanks again!

cdiener:
Oh yeah process spawning on Mac and Win is very slow.

Like I said I recently pushed some optimizations to optlang that makes this faster for Gurobi. If you install optlang from the GitHub master branch you will get those optimizations. There is about a 10x speed up just in there. If you run the benchmarks in the linked repo against that one the cobrapy version should perform better. For instance iJO1366 with Gurobi takes about 8 seconds on my laptop with that one (single core). But to answer your question the fastest way to get the primal values in cobrapy (in standard form) is to use `model.solver.primal_values` which will always use the fastest available strategy for each solver. However, this will still be slower than solving from a recycled version and only getting the objective value.

cdiener:
Sorry, my bad. The Gurobi optimizations mostly improved getting primal values etc. This shouldn't affect the previous FVA implementation, so your previous benchmarks should not change much unless you used a very outdated optlang or something.

cdiener:
Hi, closing this for now since it has diverged too much. Feel free to send a new PR.

Issue #1305, opened 2022-12-19
Numpy 1.24 breaks builtin type aliases


marvinvanaalst:
#### Problem description

The numpy people marked aliases of builtin types like `np.bool` or `np.float` as deprecated in version [1.20.0](https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated). They are effectively deprecated with version [1.24.0](https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations) which was just released.

This leads to a bunch of `AttributeError`s, like 

```python
AttributeError: module 'numpy' has no attribute 'float'
```

Your `setup.cfg` states compatability for `numpy ~=1.13` which includes all minor versions >= 13 and thus causes CI to crash on this (but that's on the numpy folks).

I've started fixing this in [my fork](https://github.com/opencobra/cobrapy/compare/devel...marvinvanaalst:cobrapy:fix-numpy-builtin-aliases) but noticed during the tests that some of the dependencies like osqp have the same issue.

The best short-term fix is probably to exclude numpy versions `>=1.24` until all the dependencies have caught up on this.

cdiener:
Yes that makes sense. Feel free to send in a PR that pins numpy to a lower version in the setup.cfg.

woutervh:
> Feel free to send in a PR that pins numpy to a lower version in the setup.cfg.
This is the wrong solution. It hides the problem of using deprecated numpy-calls, 
The title wrongly suggests that numpy needs to fix an issue,

Pinning numpy to a lower-version only makes sense when the problem arises in the thirdparty package you cannot control.

in this case  the problem is only a single np.float in 
https://github.com/opencobra/cobrapy/blob/devel/src/cobra/io/dict.py#L71

Midnighter:
PR welcome 😉

marvinvanaalst:
> > Feel free to send in a PR that pins numpy to a lower version in the setup.cfg.
> > This is the wrong solution. It hides the problem of using deprecated numpy-calls,
> > The title wrongly suggests that numpy needs to fix an issue,
> 
> Pinning numpy to a lower-version only makes sense when the problem arises in the thirdparty package you cannot control.
> 
> in this case the problem is only a single np.float in https://github.com/opencobra/cobrapy/blob/devel/src/cobra/io/dict.py#L71

[...]

PR #1306, opened 2022-12-21
fix: set temporary upper bound for numpy version 

marvinvanaalst:
* [x] Sets upper numpy version bound as numpy 1.24 breaks code
* [x] temporary fix for #1305 until dependencies have caught up
* [x] tests passed except for a security warning for mpmath, which is unrelated to this PR and apparently not yet fixed


Midnighter (approved):
Thanks for the fix.

PR #1312, opened 2023-02-07
Fix issue #1300

### LINKED ISSUE #1300: Lower and upper bounds overwriting on save
https://github.com/opencobra/cobrapy/blob/4ef4d797d234e6b7153a1eabca0d9bc1d75a7d9e/src/cobra/io/sbml.py#L1281-L1286

Hi! I'm not sure if this is the intended behavior, but these lines of code are overwriting and replacing all absolute bounds that were changed prior to saving the model. So for example, if I have a model with the following reactions-
```
-1000 < reactionA  < 1000
-1000 < reactionB  < 1000
-1000 < reactionC  < 1000
```
If I were to try changing the bounds of just reaction C with `reactionC.bounds = (-2000,4000)`, the bounds of all reactions upon saving the model (and subsequently reading it from file) would be-
```
-2000 < reactionA  < 4000
-2000 < reactionB  < 4000
-2000 < reactionC  < 4000
```
_**instead of**_
```
-1000 < reactionA  < 1000
-1000 < reactionB  < 1000
-2000 < reactionC  < 4000
```

If this _is_ indeed the intended behavior, I'd argue it doesn't make much sense and entirely defeats the purpose of having a global `configuration` object altogether. It's especially strange considering these changes only kick in when trying to save a model a file. With the way the code is written now, the `lower_bound` and `upper_bound` configurations are entirely ignored once 1 reaction is added to the model.

Suggested fix is removing the first `if` clause like so-
```
min_value = config.lower_bound 
max_value = config.upper_bound 
```

ym2877:
* [x] [fix #1300](https://github.com/opencobra/cobrapy/issues/1300)
* [x] Removed lines of code that are overwriting all min/max bounds with global min/max (and thus disregarding config lower and upper bounds)
* [x] Added 'test_bounds_on_write' to tests/test_io/test_sbml.py
Fixed issue where reaction bounds were being overwritten by global model min/max values as opposed to abiding by config lower_bound and upper_bound

### DIFF
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -3,6 +3,7 @@
 ## New features
 
 ## Fixes
+[#1300](https://github.com/opencobra/cobrapy/pull/1312)- Fixed issue where reaction bounds were being overwritten by global model min/max values when writing sbml model to file
 
 ## Other
 
--- a/src/cobra/core/formula.py
+++ b/src/cobra/core/formula.py
@@ -63,7 +63,7 @@ class Formula(Object):
             return
         composition = {}
         parsed = element_re.findall(tmp_formula)
-        for (element, count) in parsed:
+        for element, count in parsed:
             if count == "":
                 count = 1
             else:
--- a/src/cobra/core/formula.py
+++ b/src/cobra/core/formula.py
@@ -63,7 +63,7 @@ class Formula(Object):
             return
         composition = {}
         parsed = element_re.findall(tmp_formula)
-        for element, count in parsed:
+        for (element, count) in parsed:
             if count == "":
                 count = 1
             else:
--- a/src/cobra/core/metabolite.py
+++ b/src/cobra/core/metabolite.py
@@ -133,7 +133,7 @@ class Metabolite(Species):
             return None
         composition = {}
         parsed = element_re.findall(tmp_formula)
-        for element, count in parsed:
+        for (element, count) in parsed:
             if count == "":
                 count = 1
             else:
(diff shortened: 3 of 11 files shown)


ym2877:
@cdiener  can we kick off workflows?

cdiener (changes requested):
~~Please remove the formatting changes in files irrelevant to the PR goal here. You can submit formatting changes in a separate PR if you'd like though all formatting is managed by black in our CI.~~

EDIT: ignore that, looks like the black version update changed some formatting, you are all good :)

ym2877:
@cdiener no worries! so should I include those formatting changes in this PR, or wait until they're fixed on main?

cdiener:
Better here in that case otherwise the CI will fail.

ym2877:
Ok sounds good! Apologies for the back and forth, just reverted the reversion :)

cdiener:
No apologies necessary. That was all my bad! :)

cdiener:
Awesome, thanks. CI is looking good. I'll review it in the afternoon.

cdiener (approved):
This looks good to me. I think the original behavior was there to avoid the creation of large amounts of parameters if the default bounds were larger but did not compare against the correct values in `_create_bound` (which should have compared against min_value and max_value and not the config values).

In the future we might want some code that checks for the most common bounds and creates parameters for those to avoid creating huge amounts of bound parameters if the bounds deviate from the config values.

cdiener:
@ym2877 Can you add an entry to the release notes in `release-notes/next-release.md` in your branch? after that we can merge.

ym2877:
Yep, my apologies for the delay! Just added the release notes.

PR #1318, opened 2023-02-24
Link to docs fix

mdziurzynski:
This is just a simple link fix for cobrapy gh pages.

### DIFF
--- a/index.html
+++ b/index.html
@@ -100,7 +100,7 @@ the GPL.</p>
     <div class="intro">
         <h3>Getting Help</h3>
         <p>The documentation is browsable online at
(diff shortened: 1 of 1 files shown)


Midnighter:
@cdiener I can't recall, is this HTML doc really maintained or do we also generate it somehow?

I think we should go further and also remove the `/en/`. Not that we offer any other languages right now but since it's auto-detected, why not?

cdiener:
Yeah, this has to be fixed in the Hugo page otherwise it will just be overwritten again. I fixed it for the top bar but forgot the help text. @mdziurzynski Feel free to send a PR to https://github.com/opencobra/cobrapy-website that changes the link in `content/intro/help.md`. Best

PR #1322, opened 2023-03-26
remove obsolete numpy aliases `np.bool` and `np.float`

exaexa:
This basically fixes the following error that many people encountered:

```
AttributeError: module 'numpy' has no attribute 'float'.
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. Did you mean: 'float_'?
```

Migth not work with veeery old versions of numpy but I guess that requiring
numpy>=1.20 is quite safe nowadays.

* [x] fixes an issue in fbc_curation: https://github.com/matthiaskoenig/fbc_curation/issues/98
* [ ] tests added/passed

### DIFF
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -5,6 +5,8 @@
 ## Fixes
 [#1300](https://github.com/opencobra/cobrapy/pull/1312)- Fixed issue where reaction bounds were being overwritten by global model min/max values when writing sbml model to file
 
+[fbc_curation #98](`runfrog` does not work via github actions or local installation)- removed use of obsolete numpy aliases for `float` and `bool`
+
 ## Other
 
 ## Deprecated features
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -3,9 +3,13 @@
 ## New features
 
 ## Fixes
-[#1300](https://github.com/opencobra/cobrapy/pull/1312)- Fixed issue where reaction bounds were being overwritten by global model min/max values when writing sbml model to file
 
-[fbc_curation #98](`runfrog` does not work via github actions or local installation)- removed use of obsolete numpy aliases for `float` and `bool`
+Fixed an issue where reaction bounds were being overwritten by global model min/max values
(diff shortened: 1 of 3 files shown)


cdiener:
Some flake8 complaints. Note that this still pins bumpy to <1.24 due to osqp (optional dependency) having the same issue (and the fix not getting merged).

exaexa:
> @cdiener  Some flake8 complaints. 

OK, hopefully fixed now

> Note that this still pins bumpy to <1.24 due to osqp (optional dependency) having the same issue (and the fix not getting merged).

...how come I got 1.24 installed from pip then? :thinking: 

Anyway this is mainly for the purpose of having fbc_curation runnable (see the linked issue) which doesn't require QP solvers nor OSQP, so hopefully still useful.

cdiener:
That [shouldn't happen](https://github.com/opencobra/cobrapy/blob/93fb0e9a9c34d5ed7e71c35009b12c3784bac60c/setup.cfg#LL63C4-L63C4) unless another dependency forces a higher numpy version in which case you should get a warning from pip.

The fix is fine though. It just means you need to explicitly enforce a newer numpy if you need it.

cdiener (approved):
Looking good, Thanks!

exaexa:
> That [shouldn't happen](https://github.com/opencobra/cobrapy/blob/93fb0e9a9c34d5ed7e71c35009b12c3784bac60c/setup.cfg#LL63C4-L63C4) unless another dependency forces a higher numpy version in which case you should get a warning from pip.

Ah interesting, thanks for the pointers. I don't recall any warning though, will check it better next time. (Btw, so far I naively assumed that pip is sufficiently responsible to refuse any configuration that has versions messed up :sweat_smile: )

cdiener:
Yep it mostly does. But if two packages have incompatible reqs (like numpy <1.24 and numpy>1.24) then the last installed package takes precedence.

PR #1331, opened 2023-04-25
gapfilling.ipynb bug fix

pablodig:
Hi,
I noticed that the gapfilling.ipynb notebook in the documentation has an error in the second cell: First the reaction is not called correctly. I changed add_reaction() for add_reactions() which the correct way of calling it. Second, the argument in the add_reactions() function needs to be an iterable so I enclosed the reaction.copy() in []. Third, the cell needs to import cobra to work.

Finally, some of the output that I got in the following cells differ a little bit from what you have in the original notebook.

I hope you find these changes useful.

* [x] tests added/passed

### DIFF
--- a/documentation_builder/gapfilling.ipynb
+++ b/documentation_builder/gapfilling.ipynb
@@ -42,19 +42,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 3,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Scaling...\n",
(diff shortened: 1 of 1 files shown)


cdiener (changes requested):
Hi, sorry for the delay. That is looking good. Could you rerun it in an environment with only GLPK or set this as the solver with a Configuration singleton? Since the default package only comes with GLPK that is probably a better reference for the docs. Thanks!

cdiener:
Perfect, thanks!

cdiener (approved):
Thanks!

pablodig:
Sure. Im happy to help. 
Great package by the way.

PR #1340, opened 2023-07-05
Fix broken test for utils.solver with symengine

cdiener:
* [X] fix failing test on newer sympy versions
* [X] tests added/passed

Minor change that repairs a broken test. Also, the logic of the test was a bit off and has now been corrected.

### DIFF
--- a/tests/test_util/test_solver.py
+++ b/tests/test_util/test_solver.py
@@ -118,14 +118,19 @@ def test_absolute_expression(model: "Model") -> None:
 def test_fix_objective_as_constraint(solver: str, model: "Model") -> None:
     """Test fixing present objective as a constraint."""
     model.solver = solver
+    opt = model.slim_optimize()
     with model as m:
         su.fix_objective_as_constraint(model, 1.0)
         constraint_name = m.constraints[-1]
-        assert abs(m.constraints[-1].expression - m.objective.expression) < 1e-6
+        assert (m.constraints[-1].expression - m.objective.expression).simplify() == 0
+        assert m.constraints[-1].lb == pytest.approx(opt)
     assert constraint_name not in m.constraints
     su.fix_objective_as_constraint(model)
     constraint_name = model.constraints[-1]
-    assert abs(model.constraints[-1].expression - model.objective.expression) < 1e-6
+    assert (
+        model.constraints[-1].expression - model.objective.expression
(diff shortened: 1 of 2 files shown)


cdiener:
Only affects symengine, apparently.

tests/test_util/test_solver.py
```diff
@@ -118,14 +118,19 @@ def test_absolute_expression(model: "Model") -> None:
 def test_fix_objective_as_constraint(solver: str, model: "Model") -> None:
     """Test fixing present objective as a constraint."""
     model.solver = solver
+    opt = model.slim_optimize()
     with model as m:
         su.fix_objective_as_constraint(model, 1.0)
         constraint_name = m.constraints[-1]
-        assert abs(m.constraints[-1].expression - m.objective.expression) < 1e-6
+        assert (m.constraints[-1].expression - m.objective.expression).simplify() == 0
+        assert m.constraints[-1].lb == pytest.approx(opt)
```
Midnighter: It seems better to me, to pass in a name for the constraint, and then use that name to retrieve the constraint object.

Midnighter: Also, the variable name `constraint_name` is misleading here, as it is in fact the object and not the name.

tests/test_util/test_solver.py
```diff
@@ -118,14 +118,19 @@ def test_absolute_expression(model: "Model") -> None:
...
     assert constraint_name not in m.constraints
     su.fix_objective_as_constraint(model)
     constraint_name = model.constraints[-1]
-    assert abs(model.constraints[-1].expression - model.objective.expression) < 1e-6
+    assert (
+        model.constraints[-1].expression - model.objective.expression
+    ).simplify() == 0
+    assert m.constraints[-1].lb == pytest.approx(opt)
```
Midnighter: As above.

Midnighter (changes requested):
Thank you for the fix. I would at least change it to

```py
fix_constraint = m.constraints[-1]
```

and then perform assertions on that.

tests/test_util/test_solver.py
cdiener: Yeah, I did not write the original test and did not want to modify more than necessary. I changed it now to be as suggested.

Midnighter (approved):
Thanks for the changes.

PR #1341, opened 2023-07-05
fix the objective in CycleFreeFlux

### LINKED ISSUE #935: Loopless FBA objective value is zero
When I run the example from your loopless FBA documentation, I get the objective value 0 (which is also displayed in the rich output), while the flux of the biomass reaction is non-zero.

```python
import cobra.test
from cobra import Reaction, Metabolite, Model
from cobra.flux_analysis.loopless import loopless_solution

salmonella = cobra.test.create_test_model("salmonella")
nominal = salmonella.optimize()
loopless = loopless_solution(salmonella)

print(loopless.objective_value)
>>> 0.0

print(loopless.fluxes.biomass_iRR1083_metals)
>>> 0.380007972274052
```

Do I misunderstand anything about the loopless concept, or is this just bug?

<details>
System Information
==================
OS                     Linux
OS-release 4.15.0-76-generic
Python                 3.6.7

Package Versions
================
cobra                       0.17.1
depinfo                      1.5.1
future                      0.18.2
numpy                       1.18.1
optlang                      1.4.4
pandas                       1.0.1
pip                         20.0.2
ruamel.yaml                 0.16.5
setuptools                  41.2.0
six                         1.14.0
wheel                       0.33.6
</details>

### LINKED ISSUE #1262: `loopless_solution` does not respect objective
#### Problem description

Hi all, I am running loopless FBA and I noticed that when I run several times the optimisation I get different results. This does not occur systematically though: some model objective functions seem to give constant results. 

I observed this behaviour using glpk and cplex solvers. The differences between the results seem worse in cplex - at least in the tests I performed. 

I would have expected the results of loopless FBA to be stable - or more stable at least - since for some tests I can get a positive value right after a 0 value. 

Am I doing something wrong? 

#### Code Sample

```python

from cobra.flux_analysis.loopless import add_loopless, loopless_solution

salmonella = load_model('salmonella')

salmonella.objective = "GLCptspp"

for i in range(8):
    print(f"Loopless FBA {i}: {loopless_solution(salmonella).fluxes['GLCptspp']}")

```

Gives the following output with cplex 22.1.0.0:

```
Loopless FBA 0: 5.0
Loopless FBA 1: 3.611666666666687
Loopless FBA 2: 0.0
Loopless FBA 3: 3.333333333333346
Loopless FBA 4: 5.0
Loopless FBA 5: 5.0
Loopless FBA 6: 1.714285714285893
Loopless FBA 7: 0.0
```

And this output with the default glpk solver:

```
Loopless FBA 0: 5.000000000000094
Loopless FBA 1: 3.10171428571428
Loopless FBA 2: 2.7662337662337544
Loopless FBA 3: 3.0495930047098097
Loopless FBA 4: 3.043293240513111
Loopless FBA 5: 3.4537261236503722
Loopless FBA 6: 4.999999999999877
Loopless FBA 7: 3.0432932405131314
```

### Context

<details>

```
System Information
==================
OS         Darwin
OS-release 20.6.0
Python      3.7.7

Package Versions
================
appdirs                                      1.4.4
black ; extra == 'development'       not installed
bumpversion ; extra == 'development' not installed
cobra                                       0.25.0
depinfo                                      1.7.0
diskcache                                    5.2.1
future                                      0.18.2
httpx                                       0.18.2
importlib-resources                          5.2.0
isort ; extra == 'development'       not installed
numpy                                       1.21.1
optlang                                      1.5.2
pandas                                       1.3.1
pip                                         22.2.2
pydantic                                     1.8.2
python-libsbml                              5.19.0
rich                                        10.6.0
ruamel.yaml                                0.17.10
scipy                                        1.7.1
setuptools                                  57.4.0
swiglpk                                      5.0.3
tox                                         3.24.3
wheel                                       0.36.2
```

</details>

cdiener:
* [X] fix #1262 
* [X] fix #935 
* [X] tests added/passed
* [X] updated docs

A bit embarrassing I took so long for such a simple fix, but here it is.

This now fixes the objective value as well when running CycleFreeFlux as in the original publication. 

## Fixed behavior

Compare with #1262:

```
Python 3.11.4 (main, Jun  7 2023, 10:13:09) [GCC 12.2.0]
IPython 8.8.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from cobra.io import load_model
   ...: from cobra.flux_analysis.loopless import add_loopless, loopless_solution
   ...: 
   ...: salmonella = load_model('salmonella')
   ...: 
   ...: salmonella.objective = "GLCptspp"
   ...: 
   ...: for i in range(8):
   ...:     print(f"Loopless FBA {i}: {loopless_solution(salmonella).fluxes['GLCptspp']}")
   ...: 
Loopless FBA 0: 5.0
Loopless FBA 1: 5.0
Loopless FBA 2: 5.0
Loopless FBA 3: 5.0
Loopless FBA 4: 5.0
Loopless FBA 5: 5.0
Loopless FBA 6: 5.0
Loopless FBA 7: 5.0
```


cdiener:
~~Tests will probably fail for now until #1340 is merged and this has been rebased.~~

EDIT: not necessary since CI does not check with symengine yet

documentation_builder/plot_helper.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from matplotlib.pyplot import figure, xlim, ylim, gca, arrow, text, scatter
-from mpl_toolkits.axes_grid.axislines import SubplotZero
+from mpl_toolkits.axisartist.axislines import SubplotZero
```
Midnighter, line 3: Is that module change from a particular version? Then we should add a minimum version to the requirements.

src/cobra/flux_analysis/loopless.py
```diff
@@ -154,14 +154,18 @@ def loopless_solution(
     The returned flux solution has the following properties:
 
     - It contains the minimal number of loops possible and no loops at all
-      if all flux bounds include zero.
-    - It has an objective value close to the original one and the same
-      objective value id the objective expression can not form a cycle
+      if all flux bounds include zero and the objective is not in a cycle.
+    - It has the same objective value as the original flux solution and assumes
+      the that the objective does not participate in a cycle
```
Midnighter:
```suggestion
      that the objective does not participate in a cycle
```

src/cobra/flux_analysis/loopless.py
```diff
@@ -154,14 +154,18 @@ def loopless_solution(
...
       (which is usually true since it consumes metabolites).
     - It has the same exact exchange fluxes as the previous solution.
     - All fluxes have the same sign (flow in the same direction) as the
       previous solution.
 
+    When providing fluxes to the method also note that those have to come the model
+    provided, meaning that no bounds or coefficients have been changed and the
+    optimum has remained the same.
```
Midnighter:
```suggestion
    When providing fluxes to the method, please note that those have to come from the exact same model
    that you provided, meaning that no bounds or coefficients have been changed, and the
    optimum has remained the same.
```

Midnighter (changes requested):
Some minor comments. Are you sure that the original intention wasn't that if the objective was part of a cycle, breaking that cycle could lead to a lower optimum value? It seems to me that this will no longer be possible now.

documentation_builder/plot_helper.py
cdiener: Done.

cdiener:
> Some minor comments. Are you sure that the original intention wasn't that if the objective was part of a cycle, breaking that cycle could lead to a lower optimum value? It seems to me that this will no longer be possible now.

Yes that was my intention when I originally wrote that code. But as others have pointed out now that actually does not work (https://github.com/opencobra/cobrapy/issues/698, https://github.com/opencobra/cobrapy/issues/1262) and can lead to wrong results. So that's why this now returns it to the exact implementation from the CycleFreeFlux paper (which did fix the objective). If you wanted to do this you would have to run a strategy similar to the loopless FVA where that is done with an iterative approach. We could add something like this if we wanted it in another PR.

Midnighter (approved):

Midnighter:
Cheers, thank you for the work and the explanation 😃

Issue #1345, opened 2023-08-01
Incompatibility with pydantic version 2


teddygroves:
#### Problem description

I would like my project to have both the latest versions of cobrapy and pydantic as dependencies. 

However this isn't currently possible as cobrapy pins pydantic version 1.6.

This is a problem because pydantic version 2 introduced some breaking changes, so would take a lot of work to make my project compatible with version 1.6.

#### Code Sample

Here is a minimal `pyproject.toml` that demonstrates the problem

```toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[project]
name = "myproject"
version = "0.0.1"

dependencies = [
    "cobra >= 0.26.3",
    "pydantic >= 2.0.0",
]
```

and here is the error message when I run `pip install -e .` in an empty folder containing this `pyproject.toml`:

```
ERROR: Cannot install cobra==0.26.3 and myproject==0.0.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    myproject 0.0.1 depends on pydantic>=2.0.0
    cobra 0.26.3 depends on pydantic~=1.6

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

```

From a quick look I don't think cobrapy uses pydantic except for [in this file](https://github.com/opencobra/cobrapy/blob/c6f0bb4a3fb9bc8509428edb6b5ff329dfe285d8/src/cobra/io/web/biomodels_repository.py#L9), which I think is compatible with pydantic version 2 as is. It therefore doesn't seem like a very hard bug to fix: just remove the pin or change it to a version greater than 2.

### Context

Not relevant as the bug happens before cobra is installed

<details>

```
```

</details>

cdiener:
Seems reasonable to me. Feel free to send a PR with a newer requirement. That way we will see if the CI passes.

Midnighter:
Maybe just >=1.6 would allow both users of <2 and >2 to happily use cobrapy.

teddygroves:
The PR above just changes the `~` [here](https://github.com/opencobra/cobrapy/blob/c6f0bb4a3fb9bc8509428edb6b5ff329dfe285d8/setup.cfg#L66) to a `>` as @Midnighter suggests. 

[...]

PR #1346, opened 2023-08-02
Make COBRApy compatible with pydantic version 2

teddygroves:
This change addresses issue #1345, allowing projects that use any pydantic version  greater than 1.6 to also use COBRApy.

I thought it would be better not to pin a specific version number as the parts of pydantic that COBRApy currently uses aren't specific to a particular version and should be supported by future versions.

* [x] fix #(issue number)
* [ ] tests added/passed


Midnighter (approved):
Cheers, thank you for the fix 😃

teddygroves:
I think the flake8 failure relates to [this line](https://github.com/opencobra/cobrapy/blob/c6f0bb4a3fb9bc8509428edb6b5ff329dfe285d8/tests/test_core/test_core_reaction.py#L1039) which is not part of this PR.

the error
```
flake8: commands[0]> flake8 /home/runner/work/cobrapy/cobrapy/src/cobra /home/runner/work/cobrapy/cobrapy/setup.py /home/runner/work/cobrapy/cobrapy/tests
  /home/runner/work/cobrapy/cobrapy/tests/test_core/test_core_reaction.py:1039:12: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```

The [line](https://github.com/opencobra/cobrapy/blob/c6f0bb4a3fb9bc8509428edb6b5ff329dfe285d8/tests/test_core/test_core_reaction.py#L1039):
```python
    assert type(state["_gpr"]) == str
```

PR #1352, opened 2023-08-26
[Suggestion] drop suppport for Python 3.7

Midnighter:
Python 3.7 has reached end-of-life and several vulnerabilities have been discovered in numpy and scipy that themselves no longer support 3.7.


cdiener (approved):
Absolutely, makes perfect sense.

PR #1361, opened 2023-10-19
Update depinfo

Midnighter:
* [ ] fix #(issue number)
* [x] tests added/passed

Update the version of depinfo to >2 and change issue templates to use the form input.


.github/ISSUE_TEMPLATE/02-question.yaml
```diff
@@ -0,0 +1,26 @@
+name: Question
```
cdiener: Shouldn't we rather encourage that questions all go to discussions? Those are much more visible there, especially after they have been answered.

.github/ISSUE_TEMPLATE/01-bug-report.yaml
```diff
@@ -0,0 +1,64 @@
+name: 🐞 Bug report
```
cdiener, line 1: I will assume you have tested all of those and they show up well.

cdiener (commented):
Looks good to me, but I would prefer questions in the discussion section I think to have them all in one place.

.github/ISSUE_TEMPLATE/01-bug-report.yaml
Midnighter: Yes, you can try them out here if you like https://github.com/taxprofiler/taxpasta/issues/new/choose

.github/ISSUE_TEMPLATE/02-question.yaml
Midnighter: You're perfectly right, no need to duplicate.

Midnighter:
Is this more like you were thinking?

cdiener (approved):
Yep, thanks!

PR #1363, opened 2023-10-19
chore: update actions

Midnighter:
* Update actions to the latest.
* Only test oldest and newest supported Python version.
* Change GH release action.


cdiener (approved):
Looks good. Might also want to set up dependabot for the action versions.

PR #1364, opened 2023-10-19
chore: allow broader range of pandas versions

### LINKED ISSUE #1362: upgrading dependency to allow pandas>=2.0
Now the requirement is `pandas~=1.0` which is already a problem on python 3.12 (there are no wheels for `pandas==1.5.3`).
Changing to `pandas>=1.0`, or `pandas~=2.0` would solve it.

* pandas 2.0 was released on April 3, 2023.

Midnighter:
* [x] fix #1362 
* [x] tests added/passed

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,7 +61,7 @@ install_requires =
 	importlib_resources
 	numpy >=1.13
 	optlang ~=1.5
-	pandas ~=1.0
+	pandas >=1.0,<3.0
 	pydantic >=1.6
 	python-libsbml ~=5.19
 	rich >=8.0


setup.cfg
```diff
@@ -61,7 +61,7 @@ install_requires =
 	importlib_resources
 	numpy >=1.13
 	optlang ~=1.5
-	pandas ~=1.0
+	pandas >=1.0,<3.0
```
cdiener, line 64: I do see the point of doing it this way, but I feel that for packages with as few maintainers as cobrapy maybe we should only limit max versions if we know for a fact that the new major version is incompatible with cobrapy code. Especially for pandas where the max version has broken things before. This is because a user can always downgrade pandas by hand if required but going above the max version is much harder for the user.

cdiener (changes requested):

setup.cfg
Midnighter: Initially, I agreed with you, but then I thought back to past experiences and I'm less sure now. I would be okay with doing so for pandas who are very committed to maintaining a stable API, however, I don't think we should make it a general practice. For other packages like libSBML, I was even tempted to pin the version exactly, because upgrades caused us quite a lot of pain. What I'm trying to say, I think it's a quick and easy change to increase the version range at a time of _our_ choosing, when we had time to test a new major version. It's a lot more work to handle issue reports and make fixes just when a dependency comes out with an update.

cdiener: Hmm makes sense as well and in your defense there were cases where the lack of constraints broke things (like numpy deprecating the internal types). In that case we should probably get better in fixing those things (I mean being quicker about it).

cdiener (approved):
You convinced me, go ahead.

PR #1369, opened 2023-10-31
Use correct accessor for SBML group name

### LINKED ISSUE #1302: trouble referencing the name variable
#### Problem description

I've tried to update the Debian packaged version of cobrapy to the latest upstream version 0.26.0.  Unfortunately this resulted in build time test errors which you can review in the [Debian CI](https://salsa.debian.org/med-team/python-cobra/-/jobs/3564099).
Nilesh Patra has found a promising [patch](https://salsa.debian.org/med-team/python-cobra/-/blob/master/debian/patches/do-not-seek-for-name.patch) where he simply replaced an instance of `group.name` by `group.getName()` which makes the test pass.  I would like you to confirm that this patch is OK.

Kind regards, Andreas.

cdiener:
* [X] fix #1302
* [X] tests added/passed

This just adds the Debian patch to the source directly. I am still baffled this does not fail in our CI but will fail in the Debian tests :man_shrugging:

### DIFF
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -4,6 +4,8 @@
 
 ## Fixes
 
+Fixeed a bug with SBML group parsing that affects the Debian package.
+
 ## Other
 
 ## Deprecated features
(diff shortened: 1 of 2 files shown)


Midnighter (approved):
Do our test models have reactions that are part of groups?

cdiener:
> Do our test models have reactions that are part of groups?

They were running our test suite in #1302, so I guess so.

PR #1375, opened 2024-02-08
fix: reaction bounds in _add_cycle_free function

alexpan00:
fix: reaction bounds in _add_cycle_free function are assigned according to the description in CycleFreeFlux paper

* [x] fix #1374
* [x] description of feature/fix

In _add_cycle_free function, when the original flux is > 0, the ub is forced to be the maximum value between the flux and the reaction ub. It should be the minimum between these 2 values according to the CycleFreeFlux paper. The same happens with the lb of reactions with flux < 0. It has been modified to be the maximum between the original flux and the reaction lb.

* [x] tests added/passed
* [x] add an entry to the [next release](../release-notes/next-release.md)

PR #1378, opened 2024-02-18
Fix style

Midnighter:
Correct the style issues discovered in https://github.com/opencobra/cobrapy/pull/1375. So this should be merged first, then the other rebased and merged.


cdiener (approved):
🙏

PR #1397, opened 2024-08-05
Upgrade GitHub workflows

Midnighter:
## Main changes

Upgrade the GitHub workflows in the following ways:

*    Create a reusable test workflow
*    Run tests on push and PR
*    Run release on pushed tags
*    Run a weekly test job, so we can notice problems sooner
*    Rely on OIDC tokens for publishing to PyPI

## Other

Fixes the use of a deprecated numpy type check. This will work with numpy 1.


cdiener (approved):
Should be fine now.

PR #1401, opened 2024-08-23
Do not error on duplicate boundary addition

### LINKED ISSUE #1398: [Feature] way to add_boundary if not already present
### Checklist

### Problem

I want to add a boundary reaction if it is not present.
Currently the easiest way to do this is to use `model.add_boundary` which will throw a `ValueError` if the reaction is already present [as shown here](https://github.com/opencobra/cobrapy/blob/5bc58c9ab35820c213a6a7fd596c71fcbf2f33b7/src/cobra/core/model.py#L679) then catch that `ValueError` and continue.
This has a problem as `add_boundary` also throws a `ValueError` for a number of other error circumstances.
So we are forced to catch the error and check it's message if we want to be robust against that, and this is awkward.

### Solution

Three options:
- Make duplicate boundary addition throw a different exception type
- Make other error conditions not throw `ValueError`
- Add a bool argument that on duplicate reaction being added doesn't throw. `throw_on_existing=false` or some such

### Alternatives

One can't just check if reaction is already in `model.reactions` before adding, as the automatic generation of reaction id is also in that function, [here](https://github.com/opencobra/cobrapy/blob/5bc58c9ab35820c213a6a7fd596c71fcbf2f33b7/src/cobra/core/model.py#L670).

### Anything else?

_No response_

oxinabox:
Fixed #1398 , it does so following the pattern of `add_metabolites` where it just ignores duplicate additions.
And gives neither error nor warning.
Which works ideally for my use case, so I think might be reasonable for others.

* [x] fix #1398 
* [x] tests added/passed

### DIFF
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -676,7 +676,8 @@ class Model(Object):
                 "identifier. Please set the `reaction_id`."
             )
         if reaction_id in self.reactions:
-            raise ValueError(f"Boundary reaction '{reaction_id}' already exists.")
+            # It already exists so just retrieve it.
+            return self.reactions.get_by_id(reaction_id)
         name = f"{metabolite.name} {type}"
         rxn = Reaction(id=reaction_id, name=name, lower_bound=lb, upper_bound=ub)
         rxn.add_metabolites({metabolite: -1})
--- a/src/cobra/core/model.py
+++ b/src/cobra/core/model.py
@@ -677,6 +677,7 @@ class Model(Object):
             )
         if reaction_id in self.reactions:
             # It already exists so just retrieve it.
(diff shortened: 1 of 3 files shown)


src/cobra/core/model.py
```diff
@@ -676,7 +676,8 @@ def add_boundary(
                 "identifier. Please set the `reaction_id`."
             )
         if reaction_id in self.reactions:
-            raise ValueError(f"Boundary reaction '{reaction_id}' already exists.")
+            # It already exists so just retrieve it.
```
cdiener, line 679: Maybe add a log message that this happened? Does not have to be warning, but an info might be good.

cdiener (commented):
Thank you! Looking good. Just a minor comment.

cdiener (approved):

PR #1424, opened 2025-01-13
Direct gene.id modification warning message.

mdziurzynski:
Unlike with metabolites or reactions, direct modification of Gene class id property may lead implicit gene duplications in the model. The user should be notified that there is a special, safe function designed just for that.

Im just not quite sure about the `super(Gene, self.__class__).id.fset(self, new_id)` syntax and will welcome any comments.

### DIFF
--- a/src/cobra/core/gene.py
+++ b/src/cobra/core/gene.py
@@ -222,6 +222,19 @@ class Gene(Species):
         self._functional = functional
 
     @property
+    def id(self) -> str:
+        return super().id
+
+    @id.setter
+    def id(self, new_id) -> None:
+        warnings.warn(
+            "Direct modification of gene.id is discouraged and can lead to unexpected gene duplications. "
+            "Use cobra.manipulation.modify.rename_genes to safely rename genes.",
+            UserWarning
+        )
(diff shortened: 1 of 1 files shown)


Midnighter:
I think that modifying the ID can cause so many issues that we should consider disallowing that.

src/cobra/core/gene.py
```diff
@@ -221,6 +221,19 @@ def __init__(self, id: str = None, name: str = "", functional: bool = True) -> N
...
+        return super().id
+
+    @id.setter
+    def id(self, new_id) -> None:
+        warnings.warn(
+            "Direct modification of gene.id is discouraged and can lead to unexpected gene duplications. "
+            "Use cobra.manipulation.modify.rename_genes to safely rename genes.",
+            UserWarning
+        )
+        super(Gene, self.__class__).id.fset(self, new_id)
```
cdiener, line 235: As Moritz said, better not even set it but throw an error with the same message instead of a warning.

cdiener (commented):
Sorry for the late review. Might need a rebase to the latest version and an entry in the release notes. Thx

PR #1427, opened 2025-02-19
Added warning to inform users of known bug and issues fastcc.py

### LINKED ISSUE #1154: FastCC results differ from respective MATLAB function
Hello, I would like to raise a question regarding the FastCC implementation of Cobrapy. After multiple runs, FastCC outputs a different number of blocked reactions using the Recon1 model. It might be reasonable because optimization never delivers unique solutions. But, the respective MATLAB function from the COBRAToolbox always returns the same results after several runs. Since both implementations are based on the same algorithm, I guess both should give the same results. So, I think this might be something worths checking. 

I noticed that the number of irreversible reactions detected in the FastCC implementation deviates from the number of irreversible reactions computed by the MATLAB function. Reactions that are irreversible in the backward direction are not reported as irreversible in the MATLAB command window. 

For Recon1:
MATLAB: 
- 2186 irreversible reactions
- 1274 blocked reactions

Python:
- 2192 irreversible reactions
- blocked reactions differ after each run
Used solver: glpk, python 3.8.5

dagl1:
* [x] fix #1154 
* [x] description: added warning message to show that known errors are present in the fastcc code, no bugfixes
* [ ] tests added/passed

### DIFF
--- a/src/cobra/flux_analysis/fastcc.py
+++ b/src/cobra/flux_analysis/fastcc.py
@@ -34,13 +34,15 @@ def _find_sparse_mode(
         The list of reactions to consider as consistent.
 
     """
-    warn("Use find_blocked_reactions instead as "
-         "this function contains a known bug "
-         "which allows reactions with zero net flux to be considered active,"
-         " as well as known discrepancies between this function "
-         "and its Matlab implementation "
-         "present in the COBRA toolbox. "
-         "Please, see https://github.com/opencobra/cobrapy/issues/1154 for more information.")
+    warn(
+        "Use find_blocked_reactions instead as "
+        "this function contains a known bug "
+        "which allows reactions with zero net flux to be considered active,"
+        " as well as known discrepancies between this function "
+        "and its Matlab implementation "
+        "present in the COBRA toolbox. "
+        "Please, see https://github.com/opencobra/cobrapy/issues/1154 for more information."
+    )
     if rxns:
         obj_vars = []
         vars_and_cons = []
--- a/src/cobra/flux_analysis/fastcc.py
+++ b/src/cobra/flux_analysis/fastcc.py
@@ -2,6 +2,8 @@
 
 from typing import TYPE_CHECKING, List, Optional
 
+from warnings import warn
+
 from optlang.symbolics import Zero
 
 from .helpers import normalize_cutoff
@@ -33,6 +35,12 @@ def _find_sparse_mode(
         The list of reactions to consider as consistent.
 
     """
+    warn("As of 2021, release V... this function contains a known bug "
+         "which allows reactions with zero net flux to be considered active,"
+         " as well as known discrepancies between this function "
+         "(the cobrapy implementation) and its matlab implementation "
+         "present in the cobra toolbox, "
+         "see: https://github.com/opencobra/cobrapy/issues/1154")
     if rxns:
         obj_vars = []
         vars_and_cons = []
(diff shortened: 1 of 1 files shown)


src/cobra/flux_analysis/fastcc.py
```diff
@@ -33,6 +35,12 @@ def _find_sparse_mode(
         The list of reactions to consider as consistent.
 
     """
+    warn("As of 2021, release V... this function contains a known bug "
+         "which allows reactions with zero net flux to be considered active,"
+         " as well as known discrepancies between this function "
+         "(the cobrapy implementation) and its matlab implementation "
+         "present in the cobra toolbox, "
+         "see: https://github.com/opencobra/cobrapy/issues/1154")
```
Midnighter: Can't properly line wrap in this comment.

```suggestion
    warn("This function contains a known bug "
         "which allows reactions with zero net flux to be considered active,"
         " as well as known discrepancies between this function "
         "and its Matlab implementation "
         "present in the COBRA toolbox. "
         "Please, see https://github.com/opencobra/cobrapy/issues/1154 for more information.")
```

src/cobra/flux_analysis/fastcc.py
```diff
@@ -2,6 +2,8 @@
 
 from typing import TYPE_CHECKING, List, Optional
 
+from warnings import warn
```
Midnighter, line 4: Linting will complain about the empty line above.

Midnighter (changes requested):
Thank you for the contribution 🙂. I suggest a slightly different wording of the warning.

src/cobra/flux_analysis/fastcc.py
dagl1: Sorry I think I might not understand, the original is:

""Provide an implementation of FASTCC."""

from typing import TYPE_CHECKING, List, Optional

from optlang.symbolics import Zero

from .helpers import normalize_cutoff
__________

Should there not be an empty line between the built-in functions, but empty lines between: built-ins, other packages, other modules? If so I understand, otherwise I am a bit lost

cdiener (changes requested):
The warning should probably also point towards find_blocked_reactions which is slower but works.

dagl1:
> The warning should probably also point towards find_blocked_reactions which is slower but works.

Done, this is FVA based right, however doesn't use loopless FVA if I am not mistaken (I am aware that some definitions of blocked reactions don't concern themselves with loops, but in reality any isolated offshoot that forms thermodynamically infeasible loop, is still blocked, but can carry flux at the same time).?

Regarding the error, could you point me in the right direction, it is a bit unclear to me why that 1 ubuntu gives errors while the rest of the workflow was fine?

cdiener:
Yes that does not use loopless FVA. FastCC should also allow loops I think, because all those parsimony-based methods will only avoid loops when the objective itself is not part of a loop. 

You could use `add_loopless` first and then run find_blocked_reactions but it would be very slow.

The CI is complaining because of black. Running black on the changed file before pushing should fix this.

dagl1:
> Yes that does not use loopless FVA. FastCC should also allow loops I think, because all those parsimony-based methods will only avoid loops when the objective itself is not part of a loop.
> 
> You could use `add_loopless` first and then run find_blocked_reactions but it would be very slow.
> 
> The CI is complaining because of black. Running black on the changed file before pushing should fix this.

Okay I don't understand what I am doing wrong (sorry....). I ran black on the script, it reformatted a single line, and it is sucessfully committed. I feel I am wasting a lot of your time with this ;// , but also am not entirely sure where the issue comes from

cdiener:
You are all good. Has nothing to do with you. Black got updated yet again and now formats files differently. The affected files have nothing to do with your PR so just ignore this for now.

@Midnighter We should pin the black version in tox because that is happening in almost every PR now.

cdiener:
I think this might be closed in favor of #1428 which fixes FastCC.

PR #1428, opened 2025-02-26
Fix FastCC

cdiener:
* [X] fix #1427 
* [x] tests added/passed

This brings the method in line with the paper. Still not super fast but it works. Still missing some more tests.

### DIFF
--- a/src/cobra/flux_analysis/fastcc.py
+++ b/src/cobra/flux_analysis/fastcc.py
@@ -14,6 +14,7 @@ if TYPE_CHECKING:
 logger = getLogger(__name__)
 LARGE_VALUE = 1.0e6
 
+
 def _add_lp7_vars(
     model: "Model", rxns: List["Reaction"], flux_threshold: float
 ) -> None:
@@ -33,9 +34,7 @@ def _add_lp7_vars(
     vars_and_cons = []
 
     for rxn in rxns:
-        var = prob.Variable(
-            "auxiliary_{}".format(rxn.id), lb=0.0, ub=flux_threshold
-        )
+        var = prob.Variable(f"auxiliary_{rxn.id}", lb=0.0, ub=flux_threshold)
         const = prob.Constraint(
             rxn.flux_expression - var,
             name="aux_constraint_{}".format(rxn.id),
@@ -72,9 +71,9 @@ def _find_sparse_mode(
 
     # Enable constraints for the reactions
     for rid in rxn_ids:
-        model.constraints.get("aux_constraint_{}".format(rid)).lb = 0.0
+        model.constraints.get(f"aux_constraint_{rid}").lb = 0.0
 
-    obj_vars = [model.variables.get("auxiliary_{}".format(rid)) for rid in rxn_ids]
+    obj_vars = [model.variables.get(f"auxiliary_{rid}") for rid in rxn_ids]
     model.objective = Zero
     model.objective.set_linear_coefficients({v: 1.0 for v in obj_vars})
 
@@ -102,11 +101,12 @@ def _flip_coefficients(model: "Model", rxn_ids: Set[str]) -> None:
         return
     # flip reactions
     for rxn in rxn_ids:
-        const = model.constraints.get("aux_constraint_{}".format(rxn))
-        var = model.variables.get("auxiliary_{}".format(rxn))
+        const = model.constraints.get(f"aux_constraint_{rxn}")
+        var = model.variables.get(f"auxiliary_{rxn}")
         coefs = const.get_linear_coefficients(const.variables)
         const.set_linear_coefficients({k: -v for k, v in coefs.items() if k is not var})
         model.solver.update()
+
 
 def _any_set(s):
     for x in s:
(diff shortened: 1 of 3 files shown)


src/cobra/flux_analysis/fastcc.py
```diff
@@ -10,9 +11,44 @@
...
+    flux_threshold: float
+        The upper threshold an auxiliary variable can have.
+
+    """
+    prob = model.problem
+    vars_and_cons = []
+
+    for rxn in rxns:
+        var = prob.Variable(
+            "auxiliary_{}".format(rxn.id), lb=0.0, ub=flux_threshold
```
Midnighter:
```suggestion
            f"auxiliary_{rxn.id}", lb=0.0, ub=flux_threshold
```

src/cobra/flux_analysis/fastcc.py
```diff
@@ -33,36 +67,27 @@ def _find_sparse_mode(
...
-                lb=0.0,
-            )
-            vars_and_cons.extend([var, const])
-            obj_vars.append(var)
+    if not rxn_ids:
+        return set()
+
+    # Enable constraints for the reactions
+    for rid in rxn_ids:
+        model.constraints.get("aux_constraint_{}".format(rid)).lb = 0.0
```
Midnighter:
```suggestion
        model.constraints.get(f"aux_constraint_{rid}").lb = 0.0
```

src/cobra/flux_analysis/fastcc.py
```diff
@@ -33,36 +67,27 @@ def _find_sparse_mode(
...
+
+    obj_vars = [model.variables.get("auxiliary_{}".format(rid)) for rid in rxn_ids]
```
Midnighter:
```suggestion
    obj_vars = [model.variables.get(f"auxiliary_{rid}") for rid in rxn_ids]
```

src/cobra/flux_analysis/fastcc.py
```diff
@@ -73,17 +98,19 @@ def _flip_coefficients(model: "Model", rxns: List["Reaction"]) -> None:
...
 
     """
+    if not rxn_ids:
+        return
     # flip reactions
-    for rxn in rxns:
-        const = model.constraints.get("constraint_{}".format(rxn.id))
-        var = model.variables.get("auxiliary_{}".format(rxn.id))
+    for rxn in rxn_ids:
+        const = model.constraints.get("aux_constraint_{}".format(rxn))
```
Midnighter:
```suggestion
        const = model.constraints.get(f"aux_constraint_{rxn}")
```

src/cobra/flux_analysis/fastcc.py
```diff
@@ -73,17 +98,19 @@ def _flip_coefficients(model: "Model", rxns: List["Reaction"]) -> None:
...
+        var = model.variables.get("auxiliary_{}".format(rxn))
```
Midnighter:
```suggestion
        var = model.variables.get(f"auxiliary_{rxn}")
```

Midnighter (changes requested):
I mostly checked for style etc. Perhaps @dagl1 could look at the functional side, I'm not into the matter at the moment.

dagl1:
@Midnighter Ah sorry I didn't realise things required checking.
So output-wise this fastCC version corresponds to the cobra toolbox (matlab) implementation, and both are congruent with checking for blocked reactions (as in, after removing any things that could not carry flux, no more blocked reactions are detected with other methods (one is MACAW from https://www.biorxiv.org/lookup/doi/10.1101/2024.06.24.600481 )
Code wise it **looks** correct, however I did not do a line by line comparison. Additionally this is all going by the idea that the Matlab implementation is indeed correct. I will run FVA to make sure that the results corroborate (although I am expecting some reactions with very "bad" stoichiometries to show some deviance). 

Some small things in regards to the code (but both of you are my seniors so take them as questions)
Edit: I will see if I have time for a full FVA run tomorrow or Tuesday, if that corroborates I think we are good

src/cobra/flux_analysis/fastcc.py
```diff
@@ -33,36 +67,27 @@ def _find_sparse_mode(
...
-        model.objective.set_linear_coefficients({v: 1.0 for v in obj_vars})
+    sol = model.optimize(objective_sense="max")
 
-        model.optimize(objective_sense="max")
-        result = [rxn for rxn in model.reactions if abs(rxn.flux) > zero_cutoff]
-    else:
-        result = []
+    # Disable constraints for the reactions
+    for rid in rxn_ids:
+        model.constraints.get("aux_constraint_{}".format(rid)).lb = -LARGE_VALUE
```
dagl1, line 84: Optlang includes removal of constraints, as far as I can see, keeping these for later use isn't necessary?

src/cobra/flux_analysis/fastcc.py
```diff
@@ -10,9 +11,44 @@
 if TYPE_CHECKING:
     from cobra.core import Model, Reaction
 
+logger = getLogger(__name__)
+LARGE_VALUE = 1.0e6
```
dagl1, line 15: Floating constant; in the matlab implementation this is derived from the solver tolerance (which I don't like that much either), would there be anything against directly adding this as optional argument for the functions?

Midnighter:
Thank you for your evaluation @dagl1.

> Ah sorry I didn't realise things required checking.

It's not so much that it's strictly required but since most of us work alone rather than pair program, it's always better to have a second pair of eyes.

src/cobra/flux_analysis/fastcc.py
cdiener: You could add and remove them in every iteration but this is much slower. Adding constraints means extending the constraint matrix and removing contracting it. Changing a bound however is a constant time operation and very quick.

src/cobra/flux_analysis/fastcc.py
cdiener: Note that this is not the tolerance (it's 1e6 not 1e-6). Do deactivate the constraints I just want to open its bound, but optlang does not allow unbounded constraints so I just set a very large (in absolute value) lower bound here to turn off the constraint.

PR #1429, opened 2025-03-05
Pin lint tools versions

Midnighter:
This PR pins our lint tools versions as proposed [here](https://github.com/opencobra/cobrapy/pull/1427#issuecomment-2674230850). Additionally, I suggest to offer pre-commit hooks with the same versions to catch these during local development.


cdiener (approved):
Makes sense to me. The pre commit hook should probably be mentioned in the developer docs too.

Midnighter:
I added a paragraph.

Issue #1449, opened 2025-06-27
[Feature] Fast Gapfill


djinnome:
### Checklist

- [x] There are [no similar issues or pull requests](https://github.com/opencobra/cobrapy/issues) for this yet.

### Problem

I want to be able to run a fast (but approximate) gapfill over a large universal matrix and get a quick and dirty answer that I can use to reduce the size of the universal matrix for slow (but accurate) gap filling.

### Solution

The proposed solution for this feature is to add an optional boolean parameter `fast_gapfilling`  to `gapfill()` that relaxes the gapfilling problem so that instead of binary indicator values, you have continuous indicator values constrained to be between zero and one as described in the supplementary material of:

> Dreyfuss, J. M., Zucker, J. D., Hood, H. M., Ocasio, L. R., Sachs, M. S., & Galagan, J. E. (2013). Reconstruction and validation of a genome-scale metabolic model for the filamentous fungus Neurospora crassa using FARM. PLoS Computational Biology, 9(7), e1003126. https://doi.org/10.1371/journal.pcbi.1003126

Then you run `gapfill()` by minimizing the sum of the indicator variables (L1 metric), and voila: an approximate solution.  It may contain more reactions than are strictly necessary, but decreasing the integer threshold can help resolve this.

### Anything else?

There are many enzyme prediction algorithms that provide the likelihood that a protein sequence and/or structure catalyzes a particular reaction, such as:

> Yang, Y., Jerger, A., Feng, S., Wang, Z., Brasfield, C., Cheung, M. S., Zucker, J., & Guan, Q. (2024). Improved enzyme functional annotation prediction using contrastive learning with structural inference. Communications Biology, 7(1), 1690. https://doi.org/10.1038/s42003-024-07359-z

And they could be used to provide (negative) penalties for the indicator variables.  Furthermore, many reactions contain thermodynamic information, and that could be combined with the enzyme evidence to produce an overall probability that a reaction should be included in the model.

PR #1450, opened 2025-06-27
Added fast gapfilling flag. All tests pass with fast_gapfill=True

djinnome:
* [x] adds feature request from #1449
* [x] Added a boolean parameter `fast_gapfill`  to `cobra.flux_analysis.gapfilling.gapfill()` that, when `True`, changes the indicator variable types to be `continuous` instead of `binary`, enabling fast gapfilling
* [x] Added `test_fast_gapfilling` test in `tests/test_flux_analysis/test_gapfilling.py`  and it passes
* [x] added an entry to the [next release](../release-notes/next-release.md)
* [x] passes flake8 tests.

### DIFF
--- a/src/cobra/flux_analysis/gapfilling.py
+++ b/src/cobra/flux_analysis/gapfilling.py
@@ -228,13 +228,14 @@ class GapFiller:
         constraints = []
         big_m = max(max(abs(b) for b in r.bounds) for r in self.model.reactions)
         prob = self.model.problem
+        if self.fast_gapfill:
+            indicator_type = "continuous"
+        else:
+            indicator_type = "binary"
         for rxn in self.model.reactions:
             if not hasattr(rxn, "gapfilling_type"):
                 continue
-            if self.fast_gapfill:
-                indicator_type = "continuous"
-            else:
-                indicator_type = "binary"
+            
             indicator = prob.Variable(
                 name=f"indicator_{rxn.id}", lb=0, ub=1, type=indicator_type
             )
--- a/src/cobra/flux_analysis/gapfilling.py
+++ b/src/cobra/flux_analysis/gapfilling.py
@@ -67,6 +67,8 @@ class GapFiller:
         The threshold at which a value is considered non-zero (aka
         integrality threshold). If gapfilled models fail to validate,
         you may want to lower this value (default 1E-6).
+    fast_gapfill : bool, optional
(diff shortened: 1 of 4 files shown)


djinnome:
Hey folks, apologies for the slow turnaround. I finally got around to fixing the linting errors. Please rerun the workflow and it should hopefully pass all the tests, now.

djinnome:
Hmm... the unit tests failed because the Biomodels.net website was down.  Reported this issue https://github.com/opencobra/cobrapy/issues/1455 and suggested a fix.

src/cobra/flux_analysis/gapfilling.py
```diff
@@ -67,6 +67,8 @@ class GapFiller:
         The threshold at which a value is considered non-zero (aka
         integrality threshold). If gapfilled models fail to validate,
         you may want to lower this value (default 1E-6).
+    fast_gapfill : bool, optional
```
cdiener, line 70: Because this is already a gapfill function I think having it in the argument name is redundant. Just calling it "fast" should be enough. Or call it "method" and make it a string to allow for other algorithms in the future.

src/cobra/flux_analysis/gapfilling.py
```diff
@@ -375,7 +386,11 @@ def gapfill(
         which may include previously used reactions i.e., with enough
         iterations pathways including 10 steps will eventually be reported
         even if the shortest pathway is a single reaction (default 1).
-
+    fast_gapfill : bool, optional
```
cdiener, line 389: Same as before.

tests/test_io/test_web/test_load.py
```diff
@@ -89,6 +88,9 @@ def test_remote_load(model_id: str, num_metabolites: int, num_reactions: int) ->
         The total number of reactions in the model having ID `model_id`.
 
     """
+    if os.getenv("CI") == "true":
```
cdiener, line 91: Better remove this here and let's solve this in a separate PR which we can then rebase on. Right now this is also missing the `os` import which is making the CI fail. Thx

cdiener (changes requested):

cdiener:
Looking great. I will start a separate PR for the Biomodels tests issue we might need to merge beforehand.

cdiener:
Hi, the Biomodels tests should be fixed now, but you will need to rebase on the latest devel branch. Thanks!

PR #1453, opened 2025-07-04
C++-based CHRR implementation

### LINKED ISSUE #669: Biased sampling in ACHR and availability of CHRR method?
@HAherrm a new issue seemed more fitting.

> I have been using the ACHR sampler using cobra both in Matlab and Python. In Python when plotting the flux distributions it is obvious that the sampler gets stuck in a certain value which appears much more frequently than the others. This is not the case in Matlab. Do we know why this is?

> Also, I really like the CHRR algorithm in Matlab and was hoping to use it using cobrapy. Will this algorithm ever be available in cobrapy also?

> Thank you!

ripaul:
* [x] fix #669 
* [x] tests added/passed

# TL;DR

Upon popular demand, this PR implements the CHRR algorithm for asymptotically unbiased flux sampling. The provided implementation is blazingly fast due to using the [hopsy](https://github.com/modsim/hopsy) module, which provides a Python interface to the C++-implemented sampling algorithms from [HOPS](https://github.com/modsim/hops).

# Changes

A `HopsySampler` class which inherits from `HRSampler` is implemented, which interfaces to the hopsy library. The `HopsySampler` applies a rounding transformation and samples the polytope defined by the COBRApy `model` using the uniform Coordinate Hit-and-Run algorithm.

Based on the improved performance at high numbers of samples, the default method in `cobra.sampling.sample` is changed to `chrr`, though dependent on the availability of the hopsy package. In the case that it is not available, we switch back to the OptGP sampler and raise a warning. 

As hopsy is a pre-compiled package, guaranteeing its availability on all platforms is not an easy task. Therefore, it is only added as an optional dependency. A corresponding installation option `cobra[chrr]` is added.

The tests were extended to cover CHRR.

# Performance

A small test example shows the performance benefits from using hopsy's CHRR implementation:
```python
from cobra.sampling import sample

import hopsy
import numpy as np
import matplotlib.pyplot as plt

import time

model = load_model("e_coli_core")

seed = 42
methods = ['chrr', 'optgp', ]
thinning, n_chains = 100, 4
n_samples = [1_000, 5_000, 10_000, 50_000, 100_000]

ess = dict()
samples = dict()
stats = []

ess, rhat, elapsed = {}, {}, {}
for m in methods:
    ess[m], rhat[m], elapsed[m] = [], [], []
    for n in n_samples:
        _elapsed = time.time()
        samples = sample(model, n=n, processes=n_chains, method=m, thinning=thinning, seed=seed).values
        samples = samples.reshape(n_chains, n // n_chains, -1)
        _elapsed = time.time() - _elapsed
        
        ess[m].append(hopsy.ess(samples).flatten())
        rhat[m].append(hopsy.rhat(samples).flatten())
        elapsed[m].append(_elapsed)
        
        print(m, n, ess[m][-1].mean(), ess[m][-1].min(), elapsed[m][-1])
    ess[m], rhat[m], elapsed[m] = np.array(ess[m]), np.array(rhat[m]), np.array(elapsed[m])    

for i, m in enumerate(methods):
    mean, std = np.mean(ess[m], axis=1), np.std(ess[m], axis=1)
    color = plt.cm.tab10(i)
    plt.plot(n_samples, mean / elapsed[m], color=color, label=m+' (mean)')
    plt.fill_between(n_samples, (mean+std) / elapsed[m], (mean-std) / elapsed[m], color=color, alpha=.2)
    plt.plot(n_samples, np.min(ess[m], axis=1) / elapsed[m], color=color, linestyle='--', label=m+' (min)')
plt.xlabel('number of samples')
plt.ylabel(r'effective samples per second $\rightarrow$')
plt.legend()
plt.show()

for i, m in enumerate(methods):
    plt.plot(n_samples, elapsed[m], color=plt.cm.tab10(i), label=m)
plt.xlabel('number of samples')
plt.ylabel(r'elapsed time [s] $\leftarrow$')
plt.legend()
plt.show()

dims = samples['chrr'].shape[-1]
n, m = int(np.ceil(np.sqrt(dims))), int(np.ceil(np.sqrt(dims)))
fig, axs = plt.subplots(n, m, figsize=(3*m, 3*n), dpi=100)
d = 0
for i in range(n):
    for j in range(m):
        if d < dims:
            axs[i,j].set_title(names[d])
            for k, method in enumerate(samples):
                s = samples[method]
                label = f'ess={round(ess[method][0,d])}'
                axs[i,j].hist(s[:,::,d].T, histtype='step', linewidth=2, color=[f'C{k}' for _ in range(s.shape[0])], density=True, label=label)
                axs[i,j].legend()
            d += 1
        else:
            axs[i,j].axis('off')
        axs[i,j].set_yscale('log')
fig.subplots_adjust(wspace=.25, hspace=.4)
handles, _ = axs[0,0].get_legend_handles_labels()
fig.legend(handles, methods, loc='lower right', bbox_to_anchor=(.8, .15), ncols=len(methods))
plt.show()
```
The results shown were obtained on a 16-core AMD Ryzen Threadripper PRO 3955WX. 
image
image
image

The main costs inflicted by the CHRR arise from the rounding transformation, which in this case is computed using [PolyRound](https://gitlab.com/csb.ethz/PolyRound), optlang and Gurobi. Porting the rounding transformation to C++ may further improve performance. 

# Discussion

Regarding the final plot on the marginal flux distributions, it is not exactly clear to me what happens to the `GLNabc` flux, where OptGP seems to achieve very bad mixing (as measured by the ESS), which explains the bad performance of OptGP when considering the minimum ESS across all dimensions. In particular, it would be important to make sure which one of the samplers samples the correct distribution here. While I'm very confident that hopsy samples correctly in principle, it might be that I overlooked something in the problem setup. I would appreciate some double checking there. Also testing other problems than just the `e_coli_core` might be meaningful.


cdiener:
Sorry for the delay, currently on paternity leave. I will try to review in the next weeks.

I think it's a great addition and indeed something people have asked for a lot. Regarding the Glnac flux this is because in the other implementations fluxes with a bound difference smaller than the solver tolerance are considered fixed. So a flux with [-1e-12, 1e-12] would be fixed to zero.

ripaul:
Sure, no rush :) Thanks for having a look at it!

Regarding the GLNabc flux, I will check on that again. Afaik, PolyRound, the tool we use for the rounding transformation should have a similar option for just fixing fluxes dimensions with very tight bounds.

src/cobra/sampling/sampling.py
```diff
@@ -15,22 +23,30 @@
 def sample(
     model: "Model",
     n: int,
-    method: str = "optgp",
+    method: str = "chrr",
```
cdiener: So the default should work on any installation. Maybe add an "auto" option that used hopsy if installed and optgp otherwise.

cdiener (changes requested):
Looks pretty good except for the default method thing.

src/cobra/sampling/sampling.py
ripaul: I added an "auto" option and set it to be the default. Is that how you had in mind? Or would you still suggest to put "optgp" as default?

cdiener: Yes, that is exactly what I meant, thanks!

tests/test_sampling/test_sampling.py
```diff
@@ -7,7 +7,7 @@
 
 from cobra.core import Metabolite, Model, Reaction
 from cobra.flux_analysis.parsimonious import pfba
-from cobra.sampling import ACHRSampler, OptGPSampler, sample
+from cobra.sampling import ACHRSampler, OptGPSampler, hopsy_is_available, sample
```
cdiener, line 10: Because of the new auto option most of the tests will fall back to ootgp here now. The sample calls have to be adjusted to force chrr I think and it might need some fences to check whether hopsy is installed. Also need to adjust the tox.ini so hopsy is installed during the CI. Thx

cdiener (changes requested):

tests/test_sampling/test_sampling.py
ripaul: Sorry, got distracted with other work, but wanted to come back to this. Makes sense to add tests, that force CHRR, however, I'm new to `tox`, so I would need a bit of a hint here on where exactly in the `tox.ini` to install `hopsy`. Especially, that test should be conditional on a `pip install cobra[chrr]`. If that pip install already fails with a meaningful error, like, `hopsy` not being available for the user's OS/Python version, then I would not consider that a failed test, would you? That's why it's optional in the first place.

cdiener: In the tox.ini it would go into the `testenv` block and maybe the coverage one. For instance after the osqp requirement. For the tests you can just check for the hopy import similar to what you already do in the test and then skip the tests if it is not found like in https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-test-functions I think.

ripaul: Alright, I implemented that bit, however, I guess the CI will now fail whenever its running in environments that do not support hopsy. While I could change the tox.ini such that hopsy becomes an optional dependency, this would again introduce the problem of possibly running through all tests without actually testing hopsy. So I am unsure what the optimal approach would be here.

ripaul: Guess I was wrong, CI works and no tests seem to have been skipped :) Ready when you are!

src/cobra/sampling/hopsy.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+__all__ = (
+    "HopsySampler",
+    "hopsy_is_available",
+)
+
+configuration = Configuration()
+
+try:
```
cdiener, line 20: I think instead of having the entire implementation in a try block it would be nicer to have the import check in the init and sampling files and then just import HopsySampler there if hopsy is installed.

src/cobra/sampling/sampling.py
```diff
@@ -69,28 +88,66 @@ def sample(
...
        Operations Research 199846:1 , 84-95
        https://doi.org/10.1287/opre.46.1.84
 
     """
+    if method == "auto":
+        if hopsy_is_available:
+            method = "chrr"
+        else:
+            method = "optgp"
+            warnings.warn(
```
cdiener, line 133: Use the logger with the WARN level not a normal warning. We are switching away from normal warnings to logging. Thanks.

tests/test_sampling/test_sampling.py
```diff
@@ -124,6 +185,40 @@ def test_complicated_model() -> None:
     assert sum(achr.validate(achr_samples) == "v") > 95
 
 
+@pytest.mark.skipif(not hopsy_is_available, reason="hopsy not installed")
```
cdiener, line 188: For tests that test the interface itself, it's probably better to put them in their own test file similar to test_optgp.py.

cdiener (changes requested):
Some small comments. It would be great to also have a test_chrr.py test suite like for the other sampling interfaces. This also makes it more obvious in the pytest output whether it ran the chrr tests or not. 

The import thing and try/except block is a suggestion only. I leave it up to you whether you want to refactor that way.

ripaul:
I refactored as requested, happy about feedback if that looks mergeable now :)

cdiener (approved):
Perfect, thanks 🚀

PR #1465, opened 2025-11-26
Update type hinting for DictList

### LINKED ISSUE #1464: [Feature] Improved type hinting for DictList
### Checklist

### Problem

I do a lot of my work in an IDE like Visual Studio Code and really appreciate good type hinting that allows for autocomplete/suggestion features. Currently, the DictList class in COBRApy doesn't have very good type hinting. For example, in the following code, I would like VSCode to automaticaly know that `reaction` is a `cobra.Reaction` object:

```python
model = cobra.io.read_sbml_model("path/to/model.sbml")
for reaction in model.reactions:
    ... # hinted type for reaction is Any but should be cobra.Reaction

# Or using .get_by_id
reaction = model.reactions.get_by_id("rxn_id") # hinted type is cobra.Object but should be cobra.Reaction
```

Currently, I have to explicitly state that `reaction` is a `cobra.Reaction` object in order to get the expected autocomplete/suggestion features like this:

```python
model = cobra.io.read_sbml_model("path/to/model.sbml")
for reaction in model.reactions:
    reaction: cobra.Reaction
    ... # hinted type for reaction is cobra.Reaction

# Or using .get_by_id
reaction: cobra.Reaction = model.reactions.get_by_id("rxn_id")
```

### Solution

I believe the appropriate type hinting can be accomplished using a `TypeVar` from the typing module. I believe something along these lines should do the trick (and should support python versions 3.5+):

```python
from typing import TypeVar, List, Iterable
from .object import Object

_TObject = TypeVar("_TObject", Object)

class DictList(List[_TObject]):

    def __init__(self, *args: Iterable[_TObject]):
        ...

    # example type hinting for get_by_id method
    def get_by_id(self, id: Union[_TObject, str]) -> _TObject:
        ...
```

You would then have to add some type hinting in the `cobra.Model` class to specify the type for each DictList instance:

```python
# within __init__ method for cobra.core.model.Model
self.reactions: DictList[Reaction] = DictList()
self.metabolites: DictList[Metabolite] = DictList()
# etc.
```

### Alternatives

_No response_

### Anything else?

Not sure if this is something that other people would want added or if there is a reason to not add it that I am missing. I would be happy to put together a PR for it.

abates20:
* [x] Resolves #1464 
* [x] Improves type hinting for the `DictList` class by adding a TypeVar. This allows the specific type of object contained within an instantiated `DictList` to be defined. For example, the type for `model.reactions` is now `DictList[Reaction]` instead of just `DictList`. This means that now the hinted return type for `model.reactions.get_by_id` is `Reaction` instead of `Object`. 
* [x] No tests added since only type hinting was changed.

### DIFF
--- a/src/cobra/core/dictlist.py
+++ b/src/cobra/core/dictlist.py
@@ -525,7 +525,7 @@ class DictList(List[_TObject]):
         """Remove slice."""
         self.__delitem__(slice(i, j))
 
-    def __getattr__(self, attr: Any) -> Any:
+    def __getattr__(self, attr: Any) -> _TObject:
         """Get an attribute by id."""
         try:
             return DictList.get_by_id(self, attr)
--- a/src/cobra/core/dictlist.py
+++ b/src/cobra/core/dictlist.py
@@ -16,8 +16,10 @@ from typing import (
     Union,
 )
 
+from .object import Object
 
-_TObject = TypeVar("_TObject")
+
+_TObject = TypeVar("_TObject", bound=Object)
 
 
 class DictList(List[_TObject]):
--- a/src/cobra/core/dictlist.py
+++ b/src/cobra/core/dictlist.py
@@ -19,10 +19,10 @@ from typing import (
 from .object import Object
 
 
-_TObject = TypeVar("_TObject", bound=Object)
+CobraObject = TypeVar("CobraObject", bound=Object)
 
 
-class DictList(List[_TObject]):
+class DictList(List[CobraObject]):
     """
     Define a combined dict and list.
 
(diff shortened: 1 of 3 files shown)


src/cobra/core/dictlist.py
```diff
@@ -12,13 +12,17 @@
...
     Type,
+    TypeVar,
     Union,
 )
 
 from .object import Object
 
 
-class DictList(list):
+_TObject = TypeVar("_TObject", bound=Object)
```
cdiener: I would prefer a more descriptive name. Also does it need to be a hidden object? 

For instance

```python
CobraObject = TypeVar("CobraObject", bound=Object)
```

looks clearer to me.

cdiener (changes requested):

src/cobra/core/dictlist.py
abates20: Changing it from `_TObject` to `CobraObject` sounds good to me. I went ahead and updated it.

cdiener (approved):

cdiener:
Thank you!

PR #1470, opened 2026-03-23
Temporarily remove web caching due to diskcache CVE 

cdiener:
* [X] fix failing safety checks
* [X] tests added/passed

The `diskcache` package is currently flagged with a [CVE](https://github.com/grantjenks/python-diskcache/issues/357). Unfortunately the project seems to be abandoned and no fix has been merged for 1+ months. This temporarily removes the dependency of diskcache by disabling the caching. I am happy to reactivate *if* the CVE gets fixed or an alternative pops up. I could not find a package that does the same.

### DIFF
--- a/.github/workflows/safety.yml
+++ b/.github/workflows/safety.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.10, 3.13]
+        python-version: ["3.10", "3.13"]
     timeout-minutes: 360
 
     steps:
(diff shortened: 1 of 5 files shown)


Midnighter:
There is a caching layer directly for HTTPX that uses sqlite underneath. Perhaps that could be used and would remove some of the logic here.

https://hishel.com/

I've been wondering whether code for interacting with repositories should be moved into a separate package...

cdiener:
Might work but it does not provide a max cache size and it would have to be implemented at another place in the codebase (the actual download). Though I would prefer the cache to be a directory of SBML files, not some binary format hidden away in a DB.

cdiener:
I agree with everything else. Separate package for web loading and caching would be great. Also because the testing strategy probably should be different. Outages in Biomodels currently break our CI for instance.

PR #1472, opened 2026-04-28
feat: add loopless blocked reactions search + new "potentials" loopless constraints

isaf27:
* [x] adds a new loopless constraint formulation and loopless blocked reaction detection
* [x] tests added
* [ ] tests passed

# TL;DR

This PR extends loopless analysis with a metabolite-potential formulation. It also adds support for efficient clipped flux variability analysis and faster, direction-aware blocked reaction detection with loopless constraints.

The main user-facing changes are:

- `add_loopless` now supports `method="potentials"` in addition to `method="original"` and `method="fastSNP"`.
- `add_loopless` can build directional loopless constraints when `flux_threshold` is provided.
- `flux_variability_analysis` now supports `loopless="potentials"`, `fraction_of_optimum=None`, and `abs_flux_clip` for clipped flux variability analysis.
- `find_blocked_reactions` has moved into the new `cobra.flux_analysis.blocked` module and now returns a list-compatible result with direction-specific `forward_blocked` and `reverse_blocked` attributes.
- `find_blocked_reactions` supports loopless blocked reaction detection with `loopless="potentials"` and `loopless="fastSNP"`.
- `cobra.manipulation.clip_reaction_bounds` provides a helper for clipping all reaction bounds to a finite absolute value before running numerically sensitive MILP analyses.

# Code modifications

## Function `add_loopless`

`add_loopless` was modified to support:

- nullspace-based loopless constraints: `method="original"` or `method="fastSNP"`.
- directional nullspace-based loopless constraints: `method="original"` or `method="fastSNP"` together with `flux_threshold`.
- metabolite-potential loopless constraints: `method="potentials"`.
- directional metabolite-potential loopless constraints: `method="potentials"` together with `flux_threshold`.

The new `method="potentials"` formulation adds metabolite potential variables directly instead of constructing nullspace constraints. This type of thermodynamic feasibility constraint was considered, for example, by [Beard et al. (2004)](https://doi.org/10.1016/j.jtbi.2004.01.008). It provides an alternative exact loopless MILP formulation that can be used by both `add_loopless` and `flux_variability_analysis`. We found that loopless constraints added with `method="potentials"` can make optimization more efficient than `method="fastSNP"` for some tasks.

The new `flux_threshold` parameter switches `add_loopless` to a directional formulation. In this mode, each constrained reaction gets separate forward and reverse binary indicators, allowing downstream routines to determine whether positive or negative loopless flux is feasible.

## Function `flux_variability_analysis`

- `loopless="potentials"` is accepted alongside `loopless="fastSNP"` and `loopless="cycleFreeFlux"`.
- `fraction_of_optimum=None` disables the constraint for the original objective.
- `abs_flux_clip` can be used to find FVA bounds clipped to a given absolute value. Flux variability analysis is more efficient when this parameter is specified, because it does not need to find the optimal FVA bounds if they exceed the clipping value.

## New module `cobra.flux_analysis.blocked`

Blocked reaction detection was moved from `variability.py` into a new `blocked.py` module.

`find_blocked_reactions` preserves backward-compatible list behavior for fully blocked reactions, but now returns `BlockedReactionsResult`. This result is still a list of reactions blocked in both directions, and also exposes:

- `forward_blocked`: reactions that cannot carry positive flux.
- `reverse_blocked`: reactions that cannot carry negative flux.

The function now accepts `loopless="potentials"` or `loopless="fastSNP"`. It provides a fast specialized path for loopless blocked reaction detection.

Direct blocked reaction detection calls clipped flux variability analysis with a small default clipping value, `abs_flux_clip=0.1`, and compares the resulting fluxes with `zero_cutoff`.

## Helper `clip_reaction_bounds`

The new `cobra.manipulation.clip_reaction_bounds(model, max_bound)` helper clips every reaction lower and upper bound to the interval `[-max_bound, max_bound]`.

This is useful before loopless MILP analyses on models with very large bounds. Large reaction bounds can make the optimization problem numerically unstable and lead to incorrect analysis results. We recommend using a maximum bound of at most `10000`; `1000` is a good default.

# Clipped flux variability analysis

This PR introduces a new problem: clipped flux variability analysis. In this problem, optimal FVA bounds clipped to the interval `[-abs_flux_clip, abs_flux_clip]` are computed.

## Brief algorithm idea

For clipped flux variability analysis without loopless constraints, we use an auxiliary variable to maximize `min(abs_flux_clip, |flux|)`.

Clipped flux variability analysis with loopless constraints adds one additional step: a clipped reaction search that skips reactions already found during FVA over all reactions. To do this, we add directional loopless constraints with `flux_threshold=abs_flux_clip` and maximize the sum of indicators until the sum is zero. During this process, each MILP iteration can find multiple clipped reactions, which significantly decreases the total number of MILP solver calls.

## Blocked reaction detection

Although a separate `find_blocked_reactions` function exists, the best approach for blocked reaction analysis is to run clipped flux variability analysis with a small `abs_flux_clip=0.1`, and then choose `zero_cutoff` afterwards. This is better because blocked reaction detection depends strongly on the small `zero_cutoff` value and can be unstable. Calling `find_blocked_reactions` multiple times with different `zero_cutoff` values is less efficient than a single clipped FVA call.

## Code snippets

As mentioned above, the recommended way to perform blocked reaction analysis is to run clipped flux variability analysis first and choose `zero_cutoff` afterwards:

```python


model = load_model("textbook")

clipped_flux_span = flux_variability_analysis(
    model,
    # loopless="potentials",  # uncomment to add loopless constraints
    fraction_of_optimum=None,
    abs_flux_clip=0.1,
)

zero_cutoff = model.tolerance
blocked = clipped_flux_span.index[
    (clipped_flux_span["minimum"] >= -zero_cutoff)
    & (clipped_flux_span["maximum"] <= zero_cutoff)
].to_list()
```

Alternatively, blocked reactions can still be found directly:

```python
from cobra.flux_analysis import find_blocked_reactions


model = load_model("textbook")
blocked = find_blocked_reactions(
    model,
    # loopless="potentials",  # uncomment to add loopless constraints
)
```


isaf27:
Hi @cdiener, I’ve finalized this PR. Could you take a look and give a feedback when you have a chance?

cdiener:
Like mentioned in the mail a bit more encapsulation would be good. So have most of the potentials code in its own module and then call it in the appropriate places like the FVA runner add_loopless etc. Also, have you seen #1483 ? Maybe you could sync for a shared or stacked PRs?

assaron:
@cdiener yes, we did a bit more benchmarking and probably will adjust the algorithm a little bit, and then will try to repack it according to your suggestion. 

PR #1483 doesn't seem to add anything extra. We might incorporate time limits along the lines of #1484 though — we already use them internally anyway.

marouenbg:
Hey @assaron , for the time limits, try switching solvers if a solver is stuck on a hard milps, I find that they have a wide range of performance on the same reaction.

PR #1479, opened 2026-07-12
fix: allow Formula.__add__ to accept a str operand

steps-re:
* [x] fix #(no linked issue)
* [x] `Formula.__add__` is typed `Union[Formula, str]` and its docstring lists `other_formula : Formula, str`, but the code always read `other_formula.formula`. `Formula("H2O") + "C"` raised `AttributeError: 'str' object has no attribute 'formula'` instead of returning a combined formula. Now the operand is only unwrapped to `.formula` when it's a `Formula`, otherwise the string is used directly.
* [x] tests added/passed
* [x] add an entry to the next release notes


cdiener (approved):
Thanks!

PR #1481, opened 2026-08-01
Fix build_reaction_from_string clobbering user-set bounds

### LINKED ISSUE #1463: [BUG] build_reaction_from_string wipes reaction's bounds
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem description

I thought i would set up a reaction
1.  First by creating the reaction and setting its bounds and ID in the constructor
2. then adding it to a model
3. Then calling `build_from_string` on it to set its stoichiometry

However, on doing this i found that the bounds i set in step 1 were not being respected.

### Code sample

Code run:

```python

import cobra

rxn = cobra.Reaction("rxn", lower_bound=0, upper_bound=10)
rxn.bounds  # as expected outputs (0,10)

m1 = cobra.Model()
m1.add_reactions([rxn])
rxn.bounds

rxn.build_reaction_from_string("A->B")
rxn.bounds  # unexpected outputs (0, 1000)
```

### Environment

<details>
 ❯ pixi run depinfo --markdown cobra

### Package Information

| Package | Version |
|:--------|--------:|
| cobra   |  0.29.1 |

### Dependency Information

|       Package       |   Version   |
|:--------------------|------------:|
| appdirs~            | **missing** |
| black               | **missing** |
| bumpversion         | **missing** |
| depinfo~            | **missing** |
| diskcache~          | **missing** |
| future              |       1.0.0 |
| httpx~              | **missing** |
| importlib-resources |       6.5.2 |
| isort               | **missing** |
| numpy               |       2.3.4 |
| optlang~            | **missing** |
| pandas              |       2.3.3 |
| pydantic            |     2.11.10 |
| python-libsbml~     | **missing** |
| rich                |      14.2.0 |
| ruamel.yaml~        | **missing** |
| scipy               |      1.16.3 |
| swiglpk             |      5.0.12 |
| tox                 | **missing** |

### Build Tools Information

|  Package   | Version |
|:-----------|--------:|
| pip        |  24.3.1 |
| setuptools |  80.9.0 |
| wheel      |  0.45.1 |

### Platform Information

|         |                         |
|:--------|------------------------:|
| Linux   | 6.8.0-62-generic-x86_64 |
| CPython |                 3.12.12 |

</details>

### Anything else?

_No response_

Mohit-Ak:
## What the problem was

`build_reaction_from_string` unconditionally reset a reaction's bounds to the
configuration defaults every time it detected an arrow in the equation string.
A reaction created with custom, arrow-consistent bounds (e.g. `lower_bound=0,
upper_bound=10` with a forward `-->` arrow) always came out with `upper_bound=1000`
after the call — silently discarding the user's value.

All three arrow branches were affected:
- `-->` (forward irreversible) → always wrote `(0, config.upper_bound)`
- `<--` (reverse irreversible) → always wrote `(config.lower_bound, 0)`
- `<=>` (reversible)           → always wrote `(config.lower_bound, config.upper_bound)`

Minimal reproduction (reported in #1463):

```python
import cobra
rxn = cobra.Reaction("rxn", lower_bound=0, upper_bound=10)
cobra.Model().add_reactions([rxn])
rxn.build_reaction_from_string("A->B")
print(rxn.bounds)  # Was: (0, 1000.0)  — Expected: (0, 10)
```

## What the fix does

Reset bounds **only when the pre-existing bounds contradict the arrow's direction**:

| Arrow | Reset condition |
|-------|----------------|
| `-->` | `lower_bound < 0` (lb that forbids forward flux) |
| `<--` | `upper_bound > 0` (ub that forbids reverse flux) |
| `<=>` | NOT `(lb < 0 < ub)` (bounds that prevent at least one direction) |

Compatible bounds (like `(0, 10)` with `-->`) are left untouched.
Inconsistent bounds (like `(-10, 10)` with `-->`) are still reset to defaults,
preserving the original intent that arrow type controls directionality.

The change is in `src/cobra/core/reaction.py`, 10 lines modified in the three
arrow branches.  No other code is touched.

## How it was tested

**RED → GREEN (local, verified):**

1. Ran the exact repro from the issue — confirmed `(0, 1000.0)` output on
   `devel` HEAD before the fix, `(0, 10)` output after.

2. Enumerated all three arrow branches × consistent / inconsistent bounds
   (8 cases) — all behaved correctly after the fix.

3. New regression test `test_build_from_string_preserves_compatible_bounds`
   added in `tests/test_core/test_core_reaction.py`. It covers:
   - forward arrow preserving `(0, 10)` and `(5, 10)` (consistent)
   - forward arrow resetting `(-10, 10)` (inconsistent lb)
   - reverse arrow preserving `(-10, 0)` (consistent)
   - reversible arrow preserving `(-10, 10)` (consistent)
   - reversible arrow resetting `(0, 10)` (inconsistent — no negative lb)

   The test **FAILS** on `devel` HEAD (`(0, 1000.0) != (0, 10)`) and
   **PASSES** with the fix applied.

4. Pre-existing tests `test_build_from_string` and
   `test_build_from_string_creating_metabolites` both pass without changes.

5. Full `tests/test_core/test_core_reaction.py` run: **54 passed, 6 errors**
   (the 6 errors are baseline — missing `pytest-benchmark` fixture, identical
   on pristine `devel`).

6. Linters: `flake8`, `black --check`, `isort --check-only` — all clean on
   both changed files.

Fixes #1463

### DIFF
--- a/src/cobra/core/reaction.py
+++ b/src/cobra/core/reaction.py
@@ -1555,19 +1555,29 @@ class Reaction(Object):
         # reversible case
         arrow_match = reversible_arrow_finder.search(reaction_str)
         if arrow_match is not None:
-            self.bounds = config.lower_bound, config.upper_bound
+            # Only reset bounds when they are inconsistent with a reversible
+            # arrow (i.e. the current bounds do not allow flux in both
+            # directions).  Preserves user-set bounds that are already valid.
+            if not (self._lower_bound < 0 < self._upper_bound):
+                self.bounds = config.lower_bound, config.upper_bound
         else:  # irreversible
             # try forward
             arrow_match = forward_arrow_finder.search(reaction_str)
             if arrow_match is not None:
-                self.bounds = 0, config.upper_bound
+                # Only reset when the lower bound contradicts forward-only
+                # directionality (lb < 0).  Preserve a tighter upper bound.
+                if self._lower_bound < 0:
+                    self.bounds = 0, config.upper_bound
             else:
                 # must be reverse
                 arrow_match = reverse_arrow_finder.search(reaction_str)
                 if arrow_match is None:
                     raise ValueError(f"no suitable arrow found in '{reaction_str}'")
                 else:
-                    self.bounds = config.lower_bound, 0
+                    # Only reset when the upper bound contradicts reverse-only
+                    # directionality (ub > 0).  Preserve a tighter lower bound.
+                    if self._upper_bound > 0:
+                        self.bounds = config.lower_bound, 0
         reactant_str = reaction_str[: arrow_match.start()].strip()
         product_str = reaction_str[arrow_match.end() :].strip()
 
--- a/tests/test_core/test_core_reaction.py
+++ b/tests/test_core/test_core_reaction.py
@@ -396,6 +396,64 @@ def test_build_from_string_creating_metabolites() -> None:
     assert model.reactions.R1.compartments == set(["c"])
 
 
+def test_build_from_string_preserves_compatible_bounds() -> None:
+    """Test that build_reaction_from_string preserves bounds consistent with the arrow.
+
+    Regression test for https://github.com/opencobra/cobrapy/issues/1463.
+    Before the fix, ALL three arrow branches unconditionally reset bounds to
+    the config defaults, discarding any user-set bounds.
+    """
+    from cobra.core.configuration import Configuration
+
+    config = Configuration()
+
+    # --- forward arrow: lb >= 0 is consistent, should be preserved ---
+    model = Model()
+    rxn = Reaction("fwd", lower_bound=0, upper_bound=10)
+    model.add_reactions([rxn])
+    rxn.build_reaction_from_string("A --> B", verbose=False)
+    assert rxn.bounds == (
+        0,
+        10,
+    ), "forward-compatible bounds must be preserved (issue #1463)"
+
+    # Tighter non-zero lb that is still >= 0 should also be preserved
+    model2 = Model()
+    rxn2 = Reaction("fwd2", lower_bound=5, upper_bound=10)
+    model2.add_reactions([rxn2])
+    rxn2.build_reaction_from_string("A --> B", verbose=False)
+    assert rxn2.bounds == (5, 10)
+
+    # Inconsistent lb < 0 for a forward arrow → reset to config default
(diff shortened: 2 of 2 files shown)


Mohit-Ak:
Marked ready for review.

One note on the red badge so it doesn't cost a reviewer time: the six failing `safety` jobs are not caused by this change. They all die at the same place, before any dependency is actually scanned:

```
File ".tox/safety/lib/python3.10/site-packages/nltk/inisec.py", line 128, in find_spec
    raise ImportError(
ImportError: Blocked import of regex from current working directory for security reasons.
Use '-P' or set PYTHONSAFEPATH to prevent Python from searching the current working directory.
```

That's `safety`'s vendored `nltk` refusing to import `regex` from the checkout directory — an environment/tooling issue in the `safety` tox env, unrelated to the diff here. This PR touches only `src/cobra/core/reaction.py` (+13/-3) and `tests/test_core/test_core_reaction.py` (+58/-0); it adds no dependencies and changes nothing about the packaging or the tox config.

The `prerequisites` jobs that do exercise the code are all green: lint plus the full test matrix on Linux, macOS and Windows across 3.10 and 3.14.

src/cobra/core/reaction.py
```diff
@@ -1555,19 +1555,29 @@ def build_reaction_from_string(
         # reversible case
         arrow_match = reversible_arrow_finder.search(reaction_str)
         if arrow_match is not None:
-            self.bounds = config.lower_bound, config.upper_bound
+            # Only reset bounds when they are inconsistent with a reversible
+            # arrow (i.e. the current bounds do not allow flux in both
+            # directions).  Preserves user-set bounds that are already valid.
+            if not (self._lower_bound < 0 < self._upper_bound):
+                self.bounds = config.lower_bound, config.upper_bound
```
cdiener, line 1562: Makes sense. Should we also throw a warning through the logger? I think I have a slight prefernce to respect the specific bound in the valid direction. So changing a reaction with `<=> [-10, 10]` to `-> [0, 10]` and `=> [0, 10]` to `[-10, 10]`. This might be preferable for models that were created with different defaults. For instance an SBML model might have [-100, 100] set as default bounds.

cdiener: I think this should also be `<=`. Zero itself is a valid bound and [0, 0] bounds are permitted.

PR #1482, opened 2026-08-07
feat: add return_fluxes to flux_variability_analysis

marouenbg:
### Motivation

`flux_variability_analysis` solves one optimization per reaction per direction and keeps only the objective value — every flux distribution it finds is discarded.

Callers who want those distributions (as a starting pool for sampling, as warm starts, or for ensemble methods) currently have to either re-run an FVA-sized batch of solves to recover them, or reimplement the parallel driver. Both duplicate work that has already been done.

This is particularly awkward on the loopless path: `loopless_fva_iter` already accepts `solution=True` and can return the whole solution, but `flux_variability_analysis` always calls it with the default and drops the result.

### Change

Adds an opt-in `return_fluxes` argument. When `True`:

```python
bounds, fluxes = flux_variability_analysis(model, loopless="fastSNP", return_fluxes=True)

fluxes["maximum"].loc["PGK"]     # the flux distribution found when maximising PGK
```

The second element maps `"minimum"`/`"maximum"` to data frames indexed by the optimized reaction, with reaction identifiers as columns.

**The default return value and all existing behaviour are unchanged** — `return_fluxes` defaults to `False` and the function still returns a single data frame.

### Returned distributions are loopless when a loopless method is requested

This needed a second commit and is the part most worth reviewing.

Reactions that cannot carry a loop are optimized *without* the loop constraints, because their optimum is identical either way and the LP is cheaper. That shortcut is invisible while only the objective value is returned. But those solves happen in the **first** group, before `add_loopless` is ever called, so their solution *vectors* come from a completely unconstrained model and can contain loops elsewhere in the network. On `e_coli_core` that is 78 of 87 reactions.

Silently returning mostly-loopy vectors from a function the user asked to be loopless seemed like the worst possible default, so the loop law is now applied up front whenever `return_fluxes=True` and `loopless="fastSNP"`.

The bounds are unaffected: only cyclic reactions can carry loops, so constraining them is the entire loop law and the feasible set is identical either way — only the vectors change. It is slower than the default, which is the trade the flag implies, and the docstring says so.

Verified on `e_coli_core`: **190/190 returned distributions are loopless** under an independent MILP check, and the bounds are unchanged by the flag.

### Coverage

Works on the plain, `"fastSNP"` and `"cycleFreeFlux"` paths, serial and parallel. For `"cycleFreeFlux"` the loop-removed solution is returned, i.e. the one whose objective value FVA reports.

Also verified on `e_coli_core`:

- bounds identical with and without the flag
- each returned distribution attains the reported optimum for its own reaction (max error 0.0)
- returned distributions satisfy the steady-state constraint (max `|Sv|` = 1.6e-12)
- `processes=1` and `processes=4` agree

### Tests

`test_fva_return_fluxes`, parametrized over `loopless=None` and `loopless="fastSNP"`, asserting that the bounds are unaffected, that each kept distribution attains the reported optimum, that it satisfies `Sv = 0`, and — for the loopless case — that the distributions are feasible for the loopless MILP.

### One thing reviewers may want to weigh in on

The returned bounds are incumbent values, so with a MILP-based loopless FVA they can be inner by up to `MIPGap`. A vector produced by one reaction's solve can then legitimately fall slightly outside the reported range of another reaction — we saw 42 such vectors on iND750, all on a single reaction whose entire range is 4.5e-4, exceeded by 3.7e-6 (consistent with big-M × `IntFeasTol`).

That is pre-existing behaviour and I have not touched it, since reporting an *achievable* value is a defensible reading of "highest possible flux". But it becomes visible for the first time with this feature, because callers now hold both the bounds and the vectors. Happy to add a documentation note, or leave it.

### DIFF
--- a/src/cobra/flux_analysis/variability.py
+++ b/src/cobra/flux_analysis/variability.py
@@ -1,7 +1,7 @@
 """Provide variability based methods such as flux variability or gene essentiality."""
 
 import logging
-from typing import TYPE_CHECKING, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple, Union
 from warnings import warn
 
 import numpy as np
@@ -26,7 +26,9 @@ logger = logging.getLogger(__name__)
 configuration = Configuration()
 
 
-def _init_worker(model: "Model", loopless: bool, sense: str) -> None:
+def _init_worker(
+    model: "Model", loopless: bool, sense: str, return_fluxes: bool = False
+) -> None:
     """Initialize a global model object for multiprocessing.
 
     Parameters
@@ -41,12 +43,14 @@ def _init_worker(model: "Model", loopless: bool, sense: str) -> None:
     """
     global _model
     global _loopless
+    global _return_fluxes
     _model = model
     _model.solver.objective.direction = sense
     _loopless = loopless
+    _return_fluxes = return_fluxes
 
 
-def _fva_step(reaction_id: str) -> Tuple[str, float]:
+def _fva_step(reaction_id: str) -> Tuple[str, float, Optional[Dict[str, float]]]:
     """Take a step for calculating FVA.
 
     Parameters
@@ -62,6 +66,7 @@ def _fva_step(reaction_id: str) -> Tuple[str, float]:
     """
     global _model
     global _loopless
+    global _return_fluxes
     rxn = _model.reactions.get_by_id(reaction_id)
     # The previous objective assignment already triggers a reset
     # so directly update coefs here to not trigger redundant resets
@@ -72,10 +77,18 @@ def _fva_step(reaction_id: str) -> Tuple[str, float]:
     )
     _model.slim_optimize()
     sutil.check_solver_status(_model.solver.status)
+    fluxes = None
     if _loopless:
-        value = loopless_fva_iter(_model, rxn)
+        if _return_fluxes:
+            solution = loopless_fva_iter(_model, rxn, solution=True)
+            value = None if solution is None else solution.fluxes[reaction_id]
+            fluxes = None if solution is None else solution.fluxes
+        else:
+            value = loopless_fva_iter(_model, rxn)
     else:
         value = _model.solver.objective.value
+        if _return_fluxes:
+            fluxes = get_solution(_model).fluxes
     # handle infeasible case
     if value is None:
         value = float("nan")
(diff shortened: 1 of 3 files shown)


src/cobra/flux_analysis/variability.py
```diff
@@ -129,13 +143,26 @@ def flux_variability_analysis(
     processes : int, optional
         The number of parallel processes to run. If not explicitly passed,
         will be set from the global configuration singleton (default None).
+    return_fluxes : bool, optional
```
cdiener: Not super fond of the name of the arg. FVA already returns fluxes. Something like `all_fluxes` or even just `all` might be clearer.

src/cobra/flux_analysis/variability.py
```diff
@@ -129,13 +143,26 @@ def flux_variability_analysis(
...
+        reactions that can carry a loop.
 
     Returns
     -------
-    pandas.DataFrame
+    pandas.DataFrame or tuple of (pandas.DataFrame, dict of {str: pandas.DataFrame})
         A data frame with reaction identifiers as the index and two columns:
         - maximum: indicating the highest possible flux
         - minimum: indicating the lowest possible flux
+        If `return_fluxes` is True, a tuple whose second element maps
```
cdiener: This is a bit hard to grasp I feel. Maybe rather say a
```
A tuple with slots "maximum" and "minimum", each containing a DataFrame with the optimized reactions as rows/index and the obtained flux distributions for all reactions as columns.
```

or something similar.

cdiener (changes requested):
Looks pretty good to me. ~~Ideally this would also change the flux sampling code right away to use this function but this can also happen later.~~ (Nevermind, for this it would have to report primal values which would make the API more complicated). Some small requests to improve the wording. The lint also wants you to fix an import statement.

src/cobra/flux_analysis/variability.py
marouenbg: Hey @cdiener, thanks for the review. FVA returns bounds not fluxes, these bounds are not actually solutions. What do you think?

src/cobra/flux_analysis/variability.py
marouenbg: sounds good.

src/cobra/flux_analysis/variability.py
cdiener: I mean it does return the minimum and maximum flux of a reaction given a particular lower bound on the objective. Those are still fluxes for an individual reaction. Bounds for me are the human-imposed bounds on constraints and variables (right side of the inequality equation). Other alternatives would be `complete=True`, `flux_distribution=True`.

marouenbg:
Sounds good @cdiener, just changed it to all_fluxes

cdiener (approved):
Thanks 🚀

PR #1483, opened 2026-08-08
feat: add method="potentials" to add_loopless (41x faster construction at genome scale)

marouenbg:
### Motivation

`add_loopless` becomes construction-bound at genome scale. Timing its sub-phases on Recon 2.05 (7,324 reactions, 3,842 cyclic):

| phase | time | share |
|---|---|---|
| `find_cyclic_reactions` | 9.9 s | 1% |
| **`nullspace_fast_snp`** | **1000.6 s** | **98%** |
| optlang constraint construction | 15.4 s | 1% |
| total | 1025.9 s | |

A CPU sample during the dominant phase shows 100% of time in Python property-getter chains constructing the basis LPs — the solver is idle. The null-space computation, not the solver and not the constraint loop, is where the minutes go.

### Change

The basis is avoidable entirely. The loop law requires ΔG orthogonal to every internal cycle, and

```
G ⊥ null(S_int)  ⟺  G ∈ range(S_intᵀ)  ⟺  G = S_intᵀ μ
```

`method="potentials"` introduces one free variable per participating metabolite and defines each `delta_g` through that identity — orthogonality holds identically (a cycle `n` satisfies `S_int·n = 0`, hence `n·G = (S_int·n)·μ = 0`), so **no null space is ever computed**. The indicator/big-M machinery is reused unchanged. This is also the formulation with a physical reading: μ are chemical potentials, ΔG their differences — the same parameterization TFA-style methods use.

`flux_variability_analysis` accepts `loopless="potentials"` and routes it through the same exact path as `"fastSNP"`. **All defaults unchanged.**

### Equivalence

The encoded flux space is identical to the existing methods:

- loopless optima agree with `method="fastSNP"` to **8e-13** (e_coli_core) and **7e-10** (iJR904) across 50 objectives each
- the existing `test_add_loopless` loop-elimination test passes parametrized over `"potentials"` (feasible without the forced loop reaction, infeasible with it)

### Performance

| model | fastSNP construction | potentials construction |
|---|---|---|
| Recon 2.05 | 1025.9 s | **24.8 s** (3,842 indicators, 2,109 potentials) |

Small models are unaffected (construction was already sub-second).

### Tests

Existing `test_add_loopless` and `test_loopless_benchmark_before` parametrizations extended with `"potentials"`; 9/9 relevant tests pass.

### DIFF
--- a/release-notes/next-release.md
+++ b/release-notes/next-release.md
@@ -2,6 +2,8 @@
 
 ## New features
 
+- `add_loopless` gained `method="potentials"`, which encodes the loop law through free metabolite potentials (`G = S_intᵀμ`) instead of a computed null-space basis. The feasible flux space is identical to the existing methods; only construction cost differs. On Recon2 the null-space step dominates construction (98% of a 17-minute build), which "potentials" skips entirely, building the same constraints in seconds. `flux_variability_analysis` accepts `loopless="potentials"` accordingly.
+
 ## Fixes
 
 ## Other
--- a/src/cobra/flux_analysis/loopless.py
+++ b/src/cobra/flux_analysis/loopless.py
@@ -45,11 +45,18 @@ def add_loopless(
         Cutoff used for null space. Coefficients with an absolute value
         smaller than `zero_cutoff` are considered to be zero. The default
         uses the `model.tolerance` (default None).
-    method : str, "original" or "fastSNP", optional
-        The method to use for finding the null space. The "original" method
-        uses the original method from [1]_, while "fastSNP" uses a faster
-        implementation based on the FastSNP algorithm. The "fastSNP" method
-        is much faster and should be used in most cases.
+    method : str, "original", "fastSNP" or "potentials", optional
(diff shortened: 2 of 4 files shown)


cdiener:
Related to #1472 . Would be good if those could be combined or stacked.

marouenbg:
Closing in favor of #1472, which independently implements the same `method="potentials"` formulation as part of a broader loopless feature set (directional constraints, loopless blocked-reaction detection). I've moved the genome-scale benchmarks and validation evidence over there: https://github.com/opencobra/cobrapy/pull/1472#issuecomment-5254446943. #1482 (`return_fluxes`) is independent and stays open.

cdiener:
Sounds good. Would still be great to get your review on #1472 as well once it is done.

marouenbg:
Thanks @cdiener for flagging the duplicate, I had a look at #1472 before closing #1483 and it replicates internal results. Thank you!

PR #1484, opened 2026-08-08
feat: time_limit and accept_incumbent for flux_variability_analysis (genome-scale loopless FVA)

marouenbg:
## Motivation

Exact loopless FVA (MILP variants) does not terminate at genome scale: on Recon2 (7,324 reactions, 3,842 cyclic), individual proven-optimality solves routinely exceed 120 s with an unbounded MIP gap, and three uncapped attempts ran 1.5-8.5 h without returning. Yet the *feasible* side of those same solves is cheap: interrupted branch-and-bound almost always holds a feasible incumbent. Today `_fva_step` either silently records such unproven values (plain path reads `objective.value` after a mere status warning) or crashes on attribute extraction when no incumbent exists.

## What this adds

- `time_limit`: per-solve solver time limit in seconds, applied via `solver.configuration.timeout` in each worker (solver-agnostic through optlang; restored on the caller's model in serial mode).
- `accept_incumbent`: explicit semantics for interrupted solves. `False` (default): record NaN — stricter and safer than the current implicit behavior. `True`: record the incumbent, with new boolean result columns `minimum_proven`/`maximum_proven` flagging per direction whether the value carries an optimality proof. Unproven values are one-sided estimates (an unproven maximum under-estimates the true maximum). With `return_fluxes`, incumbent distributions are returned like optimal ones — feasible (loopless under the MILP variants, up to integrality tolerance) but not necessarily extreme.
- Non-optimal solves without an incumbent now record NaN with a warning instead of raising from attribute extraction.

Measured on Recon2 with a 4 s limit (potentials loop-law encoding, 8 processes): ~56% of the 14,648 solves prove optimality within the cap and the remainder return feasible loopless flux distributions — turning a computation that previously never finished into ~1 h, with provenance flags making the approximation explicit.

## Notes

Stacked on #1482 (`return_fluxes`) — only the last commit is new. Complements #1483 (`method="potentials"`). Default behavior without the new arguments is unchanged (two-column frame, same values).


marouenbg:
Closing this one: it was motivated by genome-scale capped loopless FVA, but for that use case we have settled on a native solver driver (per-solve warm-start chains make it ~2.6x faster than the portable path measured on Recon2), so I don't want to ask for review of an API we are not using ourselves. One piece may return as a minimal standalone fix: `_fva_step` can currently raise from attribute extraction when a solve ends non-optimal without a feasible solution (e.g. an externally set solver timeout), where a NaN-with-warning would be safer. #1482 and #1483 are unaffected.

Issue #1455, opened 2025-08-03
[BUG] test_load fails if Biomodels website happens to be down.


djinnome:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem description

When I submitted pull request: https://github.com/opencobra/cobrapy/pull/1450, github actions ran all the unit tests, and the job failed, not because my contribution had a bug, but because Biomodels happened to be down.  

When I asked github copilot to explain the error, this is what it said:

The job failed because the test test_remote_load for the model_id 'BIOMD0000000633' encountered a ReadTimeout when trying to download the model from the BioModels repository. The test is defined in `tests/test_io/test_web/test_load.py` and tries to load the model from a remote source:

```python
@pytest.mark.parametrize(
    "model_id, num_metabolites, num_reactions",
    [("e_coli_core", 72, 95), ("BIOMD0000000633", 50, 35)],
)
def test_remote_load(model_id: str, num_metabolites: int, num_reactions: int) -> None:
    model = load_model(model_id, cache=False)
    assert len(model.metabolites) == num_metabolites
    assert len(model.reactions) == num_reactions
```

The failure was caused by a network timeout:
```
httpx.ReadTimeout: The read operation timed out
...
raise RuntimeError(
    f"The connection to the BioModels repository failed."
) from error
```

**Solution:**
1. Mark this test with @pytest.mark.flaky or @pytest.mark.network to indicate it is network-dependent and may occasionally fail due to external issues.
2. Increase the timeout parameter (if supported) in the code that fetches remote models.
3. Skip remote tests in CI by adding a conditional skip, e.g., using an environment variable:
   ```python
   import os
   import pytest

   @pytest.mark.parametrize(...)
   def test_remote_load(...):
       if os.getenv("CI") == "true":
           pytest.skip("Skipping remote load tests on CI")
       ...
   ```
4. Alternatively, mock the remote response in CI, as is already done for other tests in this file.

Implementing any of the above will make CI more robust to transient network issues and avoid false negatives. For reference, see the test file here (ref: ca0ed139aa3ea6775ccf149bb91af1c26e71e6b0):  
https://github.com/opencobra/cobrapy/blob/ca0ed139aa3ea6775ccf149bb91af1c26e71e6b0/tests/test_io/test_web/test_load.py

### Code sample

Code run:

```bash
pytest tests/test_io/test_web/test_load.py
```

Traceback:

```text
=================================== FAILURES ===================================
[...]

PR #1487, opened 2026-08-18
fix: raise when BioModels lists no SBML document (OMEX-only models)

### LINKED ISSUE #1466: [BUG] Error loading model from BioModels repository
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem description

I am trying to use `cobra.io.load_model` to get models from the BioModels repository, but I keep getting a UnicodeDecodeError. My expectation is that using `cobra.io.load_model` should load the model from BioModels without raising an error.

### Code sample

Code run:

```python
import cobra
cobra.io.load_model("BIOMD0000001089")
```

Traceback:

```text
  File "/Users/fbg3yp/repos/reconstructor/dev/test.py", line 2, in <module>
    cobra.io.load_model("BIOMD0000001089")
  File "/Users/fbg3yp/Library/Caches/pypoetry/virtualenvs/reconstructor-rVg7mL4q-py3.9/lib/python3.9/site-packages/cobra/io/web/load.py", line 94, in load_model
    return get_model_from_gzip_sbml(data)
  File "/Users/fbg3yp/Library/Caches/pypoetry/virtualenvs/reconstructor-rVg7mL4q-py3.9/lib/python3.9/site-packages/cobra/io/web/load.py", line 199, in get_model_from_gzip_sbml
    libsbml.readSBMLFromString(gzip.decompress(stream).decode("utf-8"))
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 10: invalid start byte
```

### Environment

<details>

### Package Information

| Package | Version |
|:--------|--------:|
| cobra   |  0.30.0 |

### Dependency Information

|     Package     |   Version   |
|:----------------|------------:|
| appdirs~        | **missing** |
| black           |     25.11.0 |
| bumpversion     | **missing** |
| depinfo~        | **missing** |
| diskcache~      | **missing** |
| future          |       1.0.0 |
| httpx~          | **missing** |
| isort           | **missing** |
| numpy           |      1.24.4 |
| optlang~        | **missing** |
| pandas          |       2.3.3 |
| pydantic        |     2.11.10 |
| python-libsbml~ | **missing** |
| rich            |      14.1.0 |
| ruamel.yaml~    | **missing** |
| scipy           | **missing** |
| swiglpk         |      5.0.12 |
| tox             | **missing** |

### Build Tools Information

| Package | Version |
|:--------|--------:|
| pip     |    25.3 |

### Platform Information

|         |               |
|:--------|--------------:|
| Darwin  | 25.1.0-x86_64 |
| CPython |        3.9.24 |
</details>

### Anything else?

It appears that the file downloaded from BioModels using `BioModels.get_sbml` is actually a gzipped zip archive. For example, if you decompress the file contents with `gzip.decompress`, you can then open the archive with `zipfile.ZipFile`:

```python
import cobra
import gzip
import zipfile

biomodels = cobra.io.BioModels()
raw_contents = biomodels.get_sbml("BIOMD0000001089")
with open("model.sbml.zip", "wb") as file:
    file.write(gzip.decompress(raw_contents))

with zipfile.ZipFile("model.sbml.zip") as file:
    print(file.namelist())
```

Output from running the above code:

```
['manifest.xml', '253_2019_9630_MOESM3_ESM.xml', 'FROG/01_objective.tsv', 'FROG/02_fva.tsv', 'FROG/03_gene_deletion.tsv', 'FROG/04_reaction_deletion.tsv', 'FROG/metadata.json']
```

I haven't checked thoroughly but I would suspect that '253_2019_9630_MOESM3_ESM.xml' is the actual SBML file for the model.

Mohit-Ak:
### Problem

`cobra.io.load_model("BIOMD0000001089")` fails with

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 10: invalid start byte
```

which says nothing about what actually went wrong.

The cause is in `BioModels.get_sbml`. It walks the repository's `main` file
listing looking for an SBML document and has a `for`/`else` to handle the case
where there isn't one:

```python
for model in files.main:
    if model.name.endswith("xml"):
        break
else:
    RuntimeError(f"Could not find an SBML document for '{model_id}'.")
```

The `else` branch constructs the exception but never raises it. So the guard
does nothing, `model` is left pointing at the last (non-SBML) entry, and the
method goes on to download it anyway. For this model that file is
`253_2019_9630_MOESM3_ESM.omex` — a COMBINE archive, i.e. a zip — and the zip
bytes only blow up later in `get_model_from_gzip_sbml`, where the decode
happens. As @Midnighter noted on the issue, more models are being deposited as
OMEX archives, so this path is getting more common rather than less.

There is a second failure mode behind the same missing `raise`: if `main` is
empty the loop body never runs, `model` is never bound, and the next line
raises `UnboundLocalError: cannot access local variable 'model'` instead of
anything meaningful.

I checked the live API to confirm the shape rather than guessing —
`https://biomodels.org/model/files/BIOMD0000001089` currently returns a `main`
section whose only entry is the `.omex` file, with no `.xml` alongside it.

### The fix

Raise the error that was already meant to be raised, and include the file names
that were actually found so the message points at the cause:

```
Could not find an SBML document for 'BIOMD0000001089'. The model's main files
are ['253_2019_9630_MOESM3_ESM.omex']. Models deposited as COMBINE (OMEX)
archives are not supported yet.
```

This also fixes the empty-listing case, since the `else` branch now exits
before anything reads `model`. `load_model` already wraps repository lookups,
so callers see a `RuntimeError` on a path that is documented to raise one.

I deliberately kept this to the reporting bug. Actually reading OMEX archives
is a feature, not a fix, and it belongs in its own PR — this one just makes the
failure honest instead of surfacing as a decode error 100 lines away.

I also added the `RuntimeError` to the method's `Raises` docstring section,
which only listed `httpx.HTTPError`.

### Testing

Added `test_biomodels_without_sbml_document`, parametrised over the two cases —
an OMEX-only listing (mirroring the real response above) and an empty listing.
It asserts the `RuntimeError` and that `httpx.stream` is never called, since
not downloading a file we know is unusable is the actual behavioural change.

Confirmed the test is not vacuous by stashing only the source file and keeping
the test, against `main`:

```
$ git stash push -- src/cobra/io/web/biomodels_repository.py
$ python -m pytest tests/test_io/test_web/test_load.py::test_biomodels_without_sbml_document -q
E       UnboundLocalError: cannot access local variable 'model' where it is not associated with a value
2 failed
```

With the fix applied:

```
$ python -m pytest tests/test_io/test_web/test_load.py -q
7 passed, 1 skipped, 1 xfailed

$ python -m pytest tests/test_io -q
74 passed, 24 skipped, 7 xfailed in 11.41s
```

Linters, at the versions pinned in `tox.ini`:

```
$ black --check src/cobra/io/web/biomodels_repository.py tests/test_io/test_web/test_load.py
2 files would be left unchanged.
$ isort --check-only ...    # clean
$ flake8 ...                # clean
```

The tests are fully mocked, so they don't add another network-dependent case to
a suite that already has trouble with BioModels being flaky (#1455).

Fixes #1466

### DIFF
--- a/src/cobra/io/web/biomodels_repository.py
+++ b/src/cobra/io/web/biomodels_repository.py
@@ -70,6 +70,9 @@ class BioModels(AbstractModelRepository):
         ------
         httpx.HTTPError
             In case there are any connection problems.
+        RuntimeError
+            If the model's file listing contains no SBML document, for example
+            when the model is only deposited as a COMBINE (OMEX) archive.
 
         """
         data = BytesIO()
@@ -83,7 +86,12 @@ class BioModels(AbstractModelRepository):
             if model.name.endswith("xml"):
                 break
         else:
-            RuntimeError(f"Could not find an SBML document for '{model_id}'.")
+            raise RuntimeError(
+                f"Could not find an SBML document for '{model_id}'. The model's main "
+                f"files are "
+                f"{sorted(f.name for f in files.main) if files.main else 'empty'}. "
+                f"Models deposited as COMBINE (OMEX) archives are not supported yet."
+            )
         with self._progress, httpx.stream(
             method="GET",
             url=self._url.join(f"download/{model_id}"),
--- a/tests/test_io/test_web/test_load.py
+++ b/tests/test_io/test_web/test_load.py
@@ -10,6 +10,7 @@ import pytest
 
 from cobra import Configuration
 from cobra.io import BiGGModels, BioModels, load_model
+from cobra.io.web import biomodels_repository
 
 
 if TYPE_CHECKING:
@@ -75,6 +76,45 @@ def test_unknown_model() -> None:
     """Expect that a not found error is raised (e2e)."""
     with pytest.raises(RuntimeError):
         load_model("MODELWHO?", cache=False)
+
+
+@pytest.mark.parametrize(
+    "main_files",
+    [
+        pytest.param(
+            [{"name": "253_2019_9630_MOESM3_ESM.omex", "fileSize": "106359"}],
+            id="omex-only",
+        ),
+        pytest.param([], id="empty-listing"),
+    ],
+)
+def test_biomodels_without_sbml_document(
+    mocker: "MockerFixture", main_files: list
+) -> None:
+    """Expect a clear error when BioModels lists no SBML document.
+
+    Some models are deposited only as COMBINE (OMEX) archives, so the file
+    listing contains no ``.xml`` entry. The download must not be attempted,
+    because the archive bytes are not an SBML document and would surface much
+    later as a ``UnicodeDecodeError``.
+
+    Parameters
+    ----------
+    mocker : pytest_mock.MockerFixture
+        The mocking fixture.
+    main_files : list
+        The ``main`` section of the mocked BioModels files response.
+
+    """
+    response = mocker.Mock()
+    response.json.return_value = {"additional": [], "main": main_files}
+    mocker.patch.object(biomodels_repository.httpx, "get", return_value=response)
+    stream = mocker.patch.object(biomodels_repository.httpx, "stream")
+
+    with pytest.raises(RuntimeError, match="Could not find an SBML document"):
+        BioModels().get_sbml(model_id="BIOMD0000001089")
+
+    stream.assert_not_called()
 
 
 @pytest.mark.parametrize(


cdiener (approved):

cdiener:
Awesome, thanks!

PR #1488, opened 2026-09-04
Use a typevar for resettable

oxinabox:
the resettable decorator had incorrect type annotation.
`Callable[[Any], Any]]` is only correct for methods with no 1 argument.
But it is used on property setters, which always have 2 arguments: `self` and `value`.
We could change it to `Callable[[Any, Any], Any]]` which would say just that.
But i think just making it a type-var is better. Then it just falls though so the wrapped function can be checked

Mypy does not pickup this error at site-of-use.
But pyrefly does.

* [ ] fix #(issue number)
* [ ] tests added/passed

### DIFF
--- a/src/cobra/util/context.py
+++ b/src/cobra/util/context.py
@@ -1,7 +1,7 @@
 """Context manager for the package."""
 
 from functools import partial
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Optional, TypeVar
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/cobra/util/context.py
```diff
@@ -79,7 +79,9 @@ def get_context(obj: "Object") -> Optional[HistoryManager]:
         pass
 
 
-def resettable(func: Callable[[Any], Any]) -> Callable[[Any], Any]:
+F = TypeVar('F', bound=Callable)
```
oxinabox:
```suggestion
F = TypeVar("F", bound=Callable)
```

cdiener:
Looks good. Seems like there is an unrelated issue with an outdated URL for BIGG. We will probably need to fix this first.

cdiener (approved):

cdiener:
Failures are due to timeouts in BIGG. Merging for now.

Issue #811, opened 2019-03-05
How to handle `boundaryCondition=true` species on SBML import?


matthiaskoenig:
At the moment species with `boundaryCondition=true` are read incorrectly from SBML, resulting in many issues. Currently, such species are just dropped from the model. 
This results in:
* missing species (not all species are read, information is lost)
* incorrect charge and chemical composition balance (if species are just dropped from reactions, the mass and charge balance of the reaction will be incorrect). This affects tools working with the cobrapy Model like memote, which reports incorrect charge and mass balances on such reactions.

The correct behavior for parsing boundarySpecies is to:
* read the species in the model
* add an exchange reaction for the boundary species (to fullfill the steady state condition for the boundary species)
* set the flux bounds of the exchange reaction according to the stoichiometry (to avoid additional constraints, i.e. limiting the reaction flux via the exchange reaction, the flux bounds of the exchange reaction must be set to [ stoichiometric_coefficient * lower_bound, stoichiometric_coefficient * upper_bound ].)

I would implement the correct behavior for the SBML parser. If there are any objections please comment.

cdiener:
I think the old behavior is based on the SBML FAQ:

>  What is this "boundary condition" business?
> The boundaryCondition attribute on <species> codifies the following notion: In some systems of reactions, certain chemical species are unchanged by the reactions in the system. For example, this might happen if there is a vast overabundance of the species compared to the other ones, or the particular species is maintained (buffered) by some external means. In models of such systems, it is important to be able to indicate that the model interpreter should not generate an ODE for that species. In SBML, this is indicated by setting the boundaryCondition attribute of that species to "true".
>A critical point in the previous paragraph is the part that says the model interpreter should not generate an ODE from the system of reactions for species that are labeled as boundary conditions. It does not mean that there cannot be an ODE or other constraint generated as a result of a rule in the model. You can have boundaryCondition="true" for a species and that species can still appear on the left-hand side of an ODE if there is an SBML Rule for it.
>Summary: **boundaryCondition="true" means do not generate an ODE for this species from the stoichiometry matrix of the system of reactions.** 

So the interpretation was to not include the reaction (even though it does not allow the use of that species on the left-hand side as specified as well). 

[...]

PR #1493, opened 2026-09-16
Keep SBML-named exchanges when filling boundary species

### LINKED ISSUE #1446: [BUG] Adding unnecessary exchange reactions
### Problem description

Hi,
I have an issue with the models built with [CarveMe](https://github.com/cdanielmachado/carveme). When I load the model with cobra, it says:
```
Adding exchange reaction EX_12ppd__R_e with default bounds for boundary metabolite: 12ppd__R_e.
Adding exchange reaction EX_12ppd__S_e with default bounds for boundary metabolite: 12ppd__S_e.
[...]
```
for all the exchange reactions and then:
```
Ignoring reaction 'EX_12ppd__R_e' since it already exists.
Ignoring reaction 'EX_12ppd__S_e' since it already exists.
[...]
```
The exchange reactions are indeed in the file. When doing so cobra remove all the annotations and the bounds of the reactions. It was not the case with older version of Cobra (I could not tell which one exactly but probably 0.20 or 0.19).
I use CarveMe 1.6 (I think it does that with 1.5 too), and Cobrapy 0.29 and 0.29.1, on Linux.

[Here](https://github.com/user-attachments/files/20825979/example.xml.gz) is a model that will "do the trick". You just have to load it with the io.read_sbml_model().

Thank you for your help.
B.

### Environment
Linux
CobraPy: 0.29,0.29.1
CarveMe: 1.6,1.5

shaneraphel:
* [x] fix #1446
* [x] keep a document-named exchange instead of a later synthetic `EX_{id}`
* [x] tests added/passed

## Description

`read_sbml_model` walks two SBML channels in one load: species with
`boundaryCondition=true`, and `ListOfReactions`. The current fill invents
`EX_{met.id}` for every boundary species *before* the reaction list is
turned into cobra objects. `Model.add_reactions` then sees a duplicate id
and keeps the first object, so a CarveMe (or any other) exchange that the
document already named loses its fbc bounds and annotations to the
default interval (−1000, 1000).

That is an ordering join, not a missing exchange. The SBML-maintainer
suggestion discussed in #811 still stands for a true boundary metabolite
that has no reaction of its own: those extras are still created. Only a
synthetic whose id is already in `ListOfReactions` is skipped, after the
same `F_REACTION` replacement the later reaction loader uses.

`Model.add_boundary` already refuses to overwrite an existing id. This
change applies that same rule at the SBML fill, instead of
rewriting `add_reactions` or dropping the extra-exchange path.

Measured on a two-species document: `EX_A_e` named in
`ListOfReactions` with fbc bounds (−10, 10) is occluded to (−1000, 1000)
before the change and keeps (−10, 10) after; `EX_B_e` for an unnamed
boundary species is still synthesized.

## Changelog

- `read_sbml_model` no longer invents a default-bound `EX_{id}` when
  `ListOfReactions` already names that exchange (`#1446`).

## Status

Ready for review. Local: `test_named_sbml_exchange_keeps_document_bounds`
and `test_boundary_conditions` (fbc_ex2 still receives `EX_X` / `EX_Ac`).


release-notes/next-release.md
```diff
@@ -10,6 +10,10 @@
 
 - type-annotations on resetable properties fixed.
 - fixed the URL for the BIGG repository
+- `read_sbml_model` no longer invents a default-bound `EX_{id}` when
```
matthiaskoenig, line 13: minor language change: no longer invents -> no longer creates

matthiaskoenig (commented):
Looks good, minor language change

shaneraphel:
tanks for review!

